PR #10, opened 2017-12-05
media_libva: fix iHD decoder dump NV12 format UV error issue.

mypopydev:
iHD decoder dump NV12 U/V date is wrong, root casue is when calc data
size, need tp celing the width with 32 or 64.

Signed-off-by: Jun Zhao <jun.zhao@intel.com>


yakuizhao:
LGTM.

PR #18, opened 2017-12-07
DdiMedia_QueryConfigAttributes: do not check input value of num_attribs

### LINKED ISSUE #15: vaQueryConfigAttributes should not check input value of num_attribs
`num_attribs` is an "out" only parameter for vaQueryConfigAttributes and therefore implementations should not make any assumptions of its input value when passed (other than it's not NULL).

DdiMedia_QueryConfigAttributes() checks the input value of `num_attribs` but shouldn't since user might have not explicitly initialized its value (therefore, it could be any arbitrary value to start with).  The only contract with users is that its value is set appropriately by the implementation upon returning.

https://github.com/intel/media-driver/blob/310104a4693c476a215de13e7e9fabdf2afbad0a/media_driver/linux/common/ddi/media_libva.cpp#L2004

uartie:
num_attribs parameter is an "out" parameter.  Therefore, there is no
contract with caller about its input value.

Fixes #15

Signed-off-by: U. Artie Eoff <ullysses.a.eoff@intel.com>


XinfengZhang:
mistake when merge the PR, reverted.  the PR should be reopened.

XinfengZhang:
manually cherry-picked.

Issue #8, opened 2017-12-04
HEVC ENC need to setting VA_RC_MB


mypopydev:
User has to set VA_RC_MB when using VA_RC_CBR or VA_RC_VBR in HEVC ENC, but AVC encoding does not need to setting  VA_RC_MB for CBR/VBR mode.

xinfengz:
MBBRC should come from mb_rate_control, and driver should only set VA_RC_CBR or VBR ---- in the structure VAEncMiscParameterTypeRateControl.  so driver only need to check VA_RC_CBR | m_encodeCtx->uiRCMethod

PR #19, opened 2017-12-07
fix for HEVC MBBRC issue

XinfengZhang:
description:driver should use mb_rate_control to control the MBBRC
user just need to set the typical BRC control methold
it's a fix for #8 
Signed-off-by: xinfengz <carl.zhang@intel.com>


mypopydev:
LGTM

PR #23, opened 2017-12-12
Fix the build warning about ‘if’ clause does not guard.

mypopydev:
fix the build warning: this ‘if’ clause does not guard.

Signed-off-by: Jun Zhao <jun.zhao@intel.com>


HeFan2017:
Hi Jun. You're write, that is a bug in our code. Thanks for fixing. I will approve this change.

Issue #26, opened 2017-12-14
Incorrect EU count is used on some SKU with one EU disabled.


yakuizhao:
The driver will try to query the EU_count/sub_slicecount by using drm_ioctl. But as no slice_count is queried, it uses the below code to update the eu_count/sub_slicecount based on device_id.
    >sysInfo->SliceCount    = devInfo->SliceCount;
        sysInfo->SubSliceCount = devInfo->SubSliceCount;
        sysInfo->EUCount       = devInfo->EUCount;

See table and notes in https://01.org/sites/default/files/documentation/intel-gfx-prm-osrc-skl-vol04-configurations.pdf - "[a] Particular SKUs produced by Intel may have one EU disabled."
So the EUCount in driver is incorrect on the particular SKU.

dahuaxie:
Yes the logic seems correct to me. With the patch EU/Slice/SubSlice is individually initialized and not affect each other

PR #27, opened 2017-12-14
Only fill the uninitialized EUCount/SliceCount/SubSliceCount to fix t…

yakuizhao:
…he incorrect EU count

It is mentioned that some particular skus produced may have one EU disabled as
mentioned in
The https://01.org/sites/default/files/documentation/intel-gfx-prm-osrc-skl-vol04-configurations.pdf - "

The current code will reintialize EUCount/SliceCount/SubSliceCount even when
only one of them is not initialized.

Fix https://github.com/intel/media-driver/issues/26

Signed-off-by: Zhao Yakui <yakui.zhao@intel.com>

### DIFF
--- a/media_driver/linux/gen10/ddi/media_sysinfo_g10.cpp
+++ b/media_driver/linux/gen10/ddi/media_sysinfo_g10.cpp
@@ -45,12 +45,18 @@ static bool InitGen10MediaSysInfo(struct GfxDeviceInfo *devInfo, MEDIA_GT_SYSTEM
         return false;
     }
 
-    if (!sysInfo->SliceCount ||
-        !sysInfo->SubSliceCount ||
-        !sysInfo->EUCount)
+    if (!sysInfo->SliceCount)
     {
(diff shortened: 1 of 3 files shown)


mypopydev:
So now change the logic is to priority use Driver info (using drmIoctl query Linux kernel), the use per-defined Device table if can't query by drmIoctl?  Can you add more detail comments in the commit message or in the code? Tks.

yakuizhao:
@mypopydev , you are right.
Before this function is executed, it already tries to query the info of Eucount/subsliceount based on drm_ioctl and then initialize it.
      > sysInfo->SliceCount        = drvInfo.sliceCount;
      > sysInfo->SubSliceCount     = drvInfo.subSliceCount;
      > sysInfo->EUCount           = drvInfo.euCount;

But it is possible that some of them are zero. In such case it will cause that all of them are initialized by using the pre-defined value.  
After adding this patch, the pre-defined value is used only when it is not initialized.

Sure. I will update it in the commit log.

mypopydev:
LGTM

Issue #9, opened 2017-12-04
vaQuerySurfaceAttributes doesn't support ENC


mypopydev:
vaQuerySurfaceAttributes support DEC and VPP, but doesn't support ENC

xinfengz:
yes, not implemented for encoder, I wonder whether it is  "MUST SUPPORT"  , thanks

mypopydev:
If user want to query ENC some information (e,g,  input pixel format, max size ... ) ,  vaQuerySurfaceAttributes is the only interface for this type query, and if ENC change some thing (e,g, maybe support YUV422 in feature， change the max size from 4K to 16K...), no other way to query the ENC, so I think it is "MUST SUPPORT".

xinfengz:
input pixel format can be covered by VAConfigAttribRTFormat, max encode resolution can be supported by VAConfigAttribMaxPictureWidth VAConfigAttribMaxPictureHeight. vaQuerySurfaceAttributes can be used for surface attribute. for example: surface maxium width can be larger than encode ability resolution. they are different things.

mypopydev:
Yes, when encoding, we need to use vaQuerySurfaceAttributes to get the surface constraints, and I don't think VAConfigAttribRTFormat can cover precise pixel format, e,g VA_RT_FORMAT_YUV420 means YUV420P or NV12?

XinfengZhang:
please help to review  #29 , thanks

PR #29, opened 2017-12-16
add vaQuerySurfaceAttributes support for encode

XinfengZhang:
add vaQuerySurfaceAttributes support for encode
fix issue #9 

### DIFF
--- a/media_driver/linux/common/ddi/media_libva_caps.cpp
+++ b/media_driver/linux/common/ddi/media_libva_caps.cpp
@@ -90,6 +90,14 @@ const uint32_t MediaLibvaCaps::m_jpegSurfaceAttr[m_numJpegSurfaceAttr] =
     VA_FOURCC_444P
 };
 
+const uint32_t MediaLibvaCaps::m_jpegEncSurfaceAttr[m_numJpegEncSurfaceAttr] =
+{
+    VA_FOURCC_NV12,
+    VA_FOURCC_YUY2,
+    VA_FOURCC_UYVY,
+    VA_FOURCC_Y800
+};
+
 const VAImageFormat MediaLibvaCaps::m_supportedImageformats[] =
 {   {VA_FOURCC_BGRA, VA_LSB_FIRST, 32,  24, 0x00ff0000, 0x0000ff00, 0x000000ff,  0xff000000},
     {VA_FOURCC_ARGB, VA_LSB_FIRST, 32,  24, 0x00ff0000, 0x0000ff00, 0x000000ff,  0xff000000},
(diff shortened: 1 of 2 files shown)


media_driver/linux/common/ddi/media_libva_caps.cpp
```diff
@@ -1988,6 +1988,63 @@ VAStatus MediaLibvaCaps::QuerySurfaceAttributes(
...
+    {
+        attribs[i].type = VASurfaceAttribPixelFormat;
+        attribs[i].value.type = VAGenericValueTypeInteger;
+        attribs[i].flags = VA_SURFACE_ATTRIB_GETTABLE | VA_SURFACE_ATTRIB_SETTABLE;
+        attribs[i].value.value.i = VA_FOURCC('N', 'V', '1', '2');
+        i++;
+        
+        attribs[i].type = VASurfaceAttribMaxWidth;
+        attribs[i].value.type = VAGenericValueTypeInteger;
+        attribs[i].flags = VA_SURFACE_ATTRIB_GETTABLE | VA_SURFACE_ATTRIB_SETTABLE;
```
fhvwy: The Max/Min attributes should only be GETTABLE, not SETTABLE - it doesn't make sense to pass them to vaCreateSurfaces().

media_driver/linux/common/ddi/media_libva_caps.cpp
```diff
@@ -1988,6 +1988,63 @@ VAStatus MediaLibvaCaps::QuerySurfaceAttributes(
             i++;
         }
     }
+    else if(entrypoint == VAEntrypointEncSlice)
+    {
+        attribs[i].type = VASurfaceAttribPixelFormat;
+        attribs[i].value.type = VAGenericValueTypeInteger;
+        attribs[i].flags = VA_SURFACE_ATTRIB_GETTABLE | VA_SURFACE_ATTRIB_SETTABLE;
+        attribs[i].value.value.i = VA_FOURCC('N', 'V', '1', '2');
```
fhvwy: I realise that the only released platforms that the driver supports are old ones without 10-bit support (no KBL/CFL/GLK), but this will need to offer P010 as well on newer platforms for 10-bit encode.  (Probably fine to just ignore this for now, though.)

media_driver/linux/common/ddi/media_libva_caps.cpp
mypopydev: Agree

XinfengZhang: will change it.

Issue #25, opened 2017-12-13
libva-utils ENC test samples avcenc can't work.


mypopydev:
https://github.com/01org/libva-utils/blob/master/encode/avcenc.c can't work in media-driver, the root case is setting 0/0 in    /* Create a context for this decode pipe */
    va_status = vaCreateContext(va_dpy, avcenc_context.config_id,
                                picture_width, picture_height,
                                VA_PROGRESSIVE, 
                                0, 0,
                                &avcenc_context.context_id);

XinfengZhang:
driver store the surfaceID array to be reference or reconstruct. all recon/reference should be in the stored surfaceID array. can change it to dynamic register the surface, but is it very important?  I can evaluate the priority

mypopydev:
@XinfengZhang I observe FEI didn't follow this restrict, any  specialized reason?

XinfengZhang:
I implemented it by requirement from real usagecases,only decode and FEI.  we will add the support for legacy encode.

XinfengZhang:
fixed by  the commit 039e892

PR #31, opened 2017-12-18
dynamic register recon/ref surfaces for encode

XinfengZhang:
with this fix, application can set the context tied surface with vaCreateSurface
also it can register the recon & ref into context tied surface by vaRenderPicture
fix issue #25 

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -1770,7 +1770,7 @@ static VAStatus DdiMedia_DestroySurfaces (
             surface->pReferenceFrameSemaphore = nullptr;
         }
 
-        DdiDecode_UnRegisterRTSurfaces(ctx, surface);
+        DdiMediaUtil_UnRegisterRTSurfaces(ctx, surface);
 
         DdiMediaUtil_FreeSurface(surface);
         MOS_FreeMemory(surface);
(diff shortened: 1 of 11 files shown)


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -2208,6 +2208,7 @@ static VAStatus DdiMedia_DestroySurfaces (
         }
 
         DdiDecode_UnRegisterRTSurfaces(ctx, pSurface);
+        DdiEncode_UnRegisterRTSurfaces(ctx, pSurface);
```
yakuizhao: Can the function of DdiDecode/Encode_UnRegisterRTSurface be merged into one function? (DdiBase_UnRegiserRTSurface)
Otherwise it needs to call it twice when one surface is removed.

XinfengZhang:
two context heap. move into one function is ok to me, but it will not help the performance.

yakuizhao:
@XinfengZhang,  It only adds one wrapper that the surface is unregistered by calling once instead of twice calling. And performance won't be affected.

XinfengZhang:
thanks，will merge them together

XinfengZhang:
thanks，will merge them together

yakuizhao:
The updated patch set looks good to me. 
And they are split into several small patches. It is convenient to review them.

Issue #32, opened 2017-12-18
Provide support for cmake 2.8 and align cmake requirements across sources


dvrogozh:
Some Linux distributions, especially server distributions, still has older cmake versions than 3.5 by default. Cent OS 7.4 is an example:
```
# cmake --version
cmake version 2.8.12.2
```
Yocto is another example:
```
# cmake --version
cmake version 3.3.1

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```
Please, provide compatibility with older versions of cmake starting from 2.8 to extend Linux distributions coverage.

Also, please, align cmake version requirements across sources. As it stands now:
```
# fgrep -rsn cmake_minimum_required .
./Tools/MediaDriverTools/GenKrnBin/CMakeLists.txt:21:cmake_minimum_required(VERSION 2.6)
./Tools/MediaDriverTools/KernelBinToSource/CMakeLists.txt:21:cmake_minimum_required (VERSION 2.8)
./Tools/MediaDriverTools/KrnToHex_IGA/CMakeLists.txt:21:cmake_minimum_required(VERSION 2.6)
./cmrtlib/CMakeLists.txt:22:cmake_minimum_required(VERSION 2.8)
./cmrtlib/linux/CMakeLists.txt:21:cmake_minimum_required(VERSION 2.8)
./media_driver/CMakeLists.txt:22:cmake_minimum_required(VERSION 3.5)
./CMakeLists.txt:21:cmake_minimum_required(VERSION 3.5)
```

Mind that I have submitted similar request against gmmlib: https://github.com/intel/gmmlib/issues/1

oliver-sang:
though [intel/gmmlib#1](https://github.com/intel/gmmlib/issues/1) has been rejected, media will keep this issue open.
we have plan to decouple the media/gmmlib build in the future. At that time we will re-evaluate this cmake version request. Thanks

XinfengZhang:
#939

XinfengZhang:
the issue still there, previously, @dvrogozh tried 2.8.12 and it works. 
now, there are more different requirements, 
from 2.6 to 3.5, need to unify the requirement.  @MicroYY , could you help to check media-driver minimum requirement to keep maximum convenience to user?

MicroYY:
The minimum required cmake version is 3.4, as cmake function `pkg_get_variable` is introduced in cmake 3.4.
[...]

PR #33, opened 2017-12-18
To provide compatibility with cmake >= 2.8

dvrogozh:
Fixes: #32

### DIFF
--- a/media_driver/linux/common/ddi/media_libva_putsurface_linux.h
+++ b/media_driver/linux/common/ddi/media_libva_putsurface_linux.h
@@ -31,7 +31,7 @@
 #include <va/va_dricommon.h>
 #include "mos_defs.h"
 
-#define LIBVA_X11_NAME "libva-x11.so"
+#define LIBVA_X11_NAME "libva-x11.so.2"
 typedef struct dri_drawable *(*dri_get_drawable_func)(
     VADriverContextP ctx, XID drawable);
 typedef union dri_buffer *(*dri_get_rendering_buffer_func)(


CMakeLists.txt
```diff
@@ -18,7 +18,7 @@
 # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
 
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 2.8)
```
uartie: GNUInstallDirs requires cmake >= 2.8.5

dvrogozh: Ok, I will change to >=2.8.5 then. Any other hits?

yakuizhao:
@dvrogozh , can it be compiled and work well if it is based on 2.8.5 cmake?

dvrogozh:
I don't have 2.8.5, but I checked with 2.8.12 on centos. It works here. Let me see if I can lay my hands on cmake 2.8.5. Hold on, please.

uartie:
Perhaps you could just enforce >= 2.8.12 since you've tested it and it may be hard to find a distro with anything older than that.  I know for a fact that currently supported Ubuntu and Fedora releases all have cmake packages >= 3.5.

uartie:
If users encounter a distro with something older, then they could file an issue at that time.  But at least we won't be erroneously trying to use non-existent features if they have an older untested version.

yakuizhao:
Agree with what Arite mentioned.  Most ubuntu and fedora releases already use the cmake package >=3.5. And as the driver is built/validated on the cmake >=3.5,  it is unknown whether it has some issues when using the older version cmake. 
It is recommended that the cmake can be upgraded.

uartie:
@yakuizhao In other words, I think if @dvrogozh confirms it works on 2.8.12, then making that the minimum required is a low risk

yakuizhao:
If it is confirmed that it can work on 2.8.12,  it is also ok that the 2.8.12 is added as the lowest requirement. 
But as there exists some other versions between 2.8.12 and 3.5, we are not sure whether it can work on these versions. In such case it is much safer to use the 3.5 as the lowest requirement.

dvrogozh:
2.8.12 can work. I use exactly this version on CentOS 7.4. As for whether we can build with the middle versions, I think that's some overthinking. After all, did you verify that you can build with 3.7, 3.8. 3.9 or 3.10? Looks like cmake actually respects backward compatibility, so we should expect that we will be able to build with >=2.8.12.

I will bump requirement to tested 2.8.12.

dvrogozh:
2.8.5 actually won't work. I tried. So, that was a nice catch. There are 2 reasons to that fyi:
1. add_compile_options which appeared in 2.8.12
2. DIRECTORY added as a component to get_filename_component in 2.8.11 (PATH is legacy backward supported name)

2) is easily fixable, 1) is not that trivial. So, 2.8.12 looks really as a best minimum candidate,

oliver-sang:
the main reason for media to use 3.5 is that gmmlib is using 3.5. media needs to build gmmlib then link it for now. So if media supplies a lower version then users will get wrong idea that she/he could use that version but finally she/he will see build failure quite unexpectedly.
cmrtlib does not depend on gmmlib, so no need to do this force.
Several media tools are pretty simple and were built standalone, so no need to do this force, either.

cmake version requirement is only for build machines, I don't believe for projects you mentioned that each target machine need to clone git repo then build from source, right? So I don't think we need to limit the cmake version according to target machine. It's pretty easy to install cmake version on (several) build machine(s) as I think. Thoughts?

dvrogozh:
gmmlib can be built with cmake 2.8.12, they don't need 3.5. And I have submitted a pull request for them as well, here it is: https://github.com/intel/gmmlib/pull/2. I think that gmmlib owners simply had this version when they developed.

Folks, I think everyone understand that any version of cmake can be installed from sources and/or build server can be used. But as a developer I prefer to build on the target when I develop or investigate a bug. Additional step to transfer the built target from one machine to another is error prone and time consuming. Thus, I strongly believe that provide convenient option to develop is precious.

uartie:
LGTM

dvrogozh:
outdated

Issue #36, opened 2017-12-19
AVC enc doesn't initialize the framerate for BRC when VAEncMiscParameterFrameRate uses zero denominator


yakuizhao:
Currently the framerate in  VAEncMiscParameterFrameRate can be intercept as the below:
>     * denominator = framerate >> 16 & 0xffff;
>    * numerator   = framerate & 0xffff;
>    frame_rate = numerator / denomintor

Of course it still allows that the dominator is zero.  In such case AVC enc deesn't initialize the correct framerate for BRC.

XinfengZhang:
fix by commit 50ec6254

PR #37, opened 2017-12-19
Fix the uninitialized framerate for AVC when VAEncMiscParameterFrameR…

yakuizhao:
…ate uses zero denominator

Fix https://github.com/intel/media-driver/issues/36

Signed-off-by: Zhao Yakui <yakui.zhao@intel.com>


yakuizhao:
Now this code is already merged. 
So close this PR.

PR #56, opened 2017-12-26
keep quality level until new VAEncMiscParameterBufferQualityLevel

### LINKED ISSUE #7: VAEncMiscParameterBufferQualityLevel need to reset for AVC/HEVC when sequence parameters reset.
User need to reset  VAEncMiscParameterBufferQualityLevel after reset sequence parameters , otherwise iHD driver won't use the expected quality level for encoding.

BTW:  iHD driver uses a different quality level for AVC/HEVC when reset, AVC use 1, HEVC use 4. 

I don't think this action follow the VA-API Spec, and iHD follow a  counterintuitive way for Quality Level setting.

XinfengZhang:
fix #7 

### DIFF
--- a/media_driver/linux/common/codec/ddi/media_ddi_encode_hevc.cpp
+++ b/media_driver/linux/common/codec/ddi/media_ddi_encode_hevc.cpp
@@ -315,6 +315,8 @@ VAStatus DdiEncodeHevc::EncodeInCodecHal(uint32_t numSlices)
         encodeParams.bMbQpDataEnabled  = true;
     }
 
+    PCODEC_HEVC_ENCODE_SEQUENCE_PARAMS hevcSeqParams = (PCODEC_HEVC_ENCODE_SEQUENCE_PARAMS)((uint8_t *)m_encodeCtx->pSeqParams);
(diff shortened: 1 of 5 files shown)


media_driver/linux/common/codec/ddi/media_ddi_encode_hevc.cpp
```diff
@@ -115,7 +115,7 @@ VAStatus DdiEncodeHevc::ContextInitialize(
     codecHalSettings->pCpParams = m_encodeCtx->pCpDdiInterface->GetParams();
 
     VAStatus eStatus = VA_STATUS_SUCCESS;
-
+    m_encodeCtx->targetUsage = TARGETUSAGE_RT_SPEED;
```
yakuizhao: Can the HEVC target usage also be initialized to BEST_QUALITY? Or the default usage of AVC/MPEG2 is set to RT_SPEED?

XinfengZhang:
@applewang001

media_driver/linux/common/codec/ddi/media_ddi_encode_hevc.cpp
XinfengZhang: yes, all will be unified set default  to RT_SPEED

Issue #62, opened 2017-12-28
[DDI]vaCreateConfig for Enc will failed without BRC setting in the attrib_list


XinfengZhang:
it is a part of #16

PR #66, opened 2017-12-29
set default RC mode to CQP, to avoid CreateConfig failed

XinfengZhang:
fix issue #62


uartie:
LGTM

PR #69, opened 2018-01-08
Remove hard dependency on X11

rlim77:
Signed-off-by: Raynald Lim <raynald.lim@intel.com>


cmrtlib/linux/CMakeLists.txt
```diff
@@ -63,9 +63,6 @@ else()
     if(NOT LIBVA_FOUND)
         pkg_check_modules(LIBVA REQUIRED libva>=1.0.0)
     endif()
-    if(NOT LIBVA_X11_FOUND)
```
uartie: If you remove this, then what defines LIBVA_X11_FOUND at cmake level?

media_driver/linux/common/codec/ddi/media_libva_decoder.cpp
```diff
@@ -42,8 +42,10 @@
 #include "media_ddi_decode_const.h"
 
 #ifndef ANDROID
+#if LIBVA_X11_FOUND
```
uartie: Where does this get defined at preprocessor level?

cmrtlib/linux/CMakeLists.txt
rlim77: Nowhere. Admittedly I was just getting this to compiled on my side without X11. I'm not sure how we should handle this elegantly (using USE_LIBVA_DRM instead seems to be an option or just having a separate compile option like -DENABLE_LIBVA). 

I would prefer removing all presentation implementation from the driver and let the application choose the presentation method it wants to use instead (i.e. do not use vaPutSurface for X11 or Wayland). Upper layers like gst-msdk, Android  and MSDK Wayland/DRI3-Present samples aren't using vaPutSurface anyway.

hefanf: currently in cmrtlib code, it chooses LIBVA_DRM by default no matter LIBVA_X11 is found or not. In cm_include.h file, USE_LIBVA_DRM is always defined. So i think it is ok for cmrtlib to totally remove LIBVA_X11_FOUND, as we just remove the dependency on libva_x11. [patch](https://github.com/intel/media-driver/commit/6b67cf3e8704e831c33aab1b9d63a5b24c7d684d).

uartie: If the code paths are never used, then it is dead code... thus, there is no point in keeping it.

If the goal is to remove all dependency on libva_x11/x11, then the code that uses it should be removed too (not just the cmake stuff).  In other words, if there is no way to enable the code via configuration (cmake), then just remove the code.  Otherwise it just creates confusion and unnecessary maintenance overhead.

The beauty of using source control (i.e. git) is that we can always get the code back if we decide it's useful in the future.

PR #76, opened 2018-01-12
remove the alignment check for encode create context

XinfengZhang:
driver should not check resolution alignment , but need to align the resolution


XinfengZhang:
closed for there are similar change in #122

Issue #57, opened 2017-12-27
[SKL][VP8 DEC] VP8 DEC lead to GPU reset in iHD.


mypopydev:
FFmpeg VA-API Vp8 DEC work with iHD driver will lead to GPU reset, and if skip the first Byte in macroblock_offset, this issue is disappear. I give a workround in FFmpeg, more information you can refer to the link: [https://github.com/mypopydev/FFmpeg/commit/8f78dbea03056d125f3afe2148b5b658ec537eee](https://github.com/mypopydev/FFmpeg/commit/8f78dbea03056d125f3afe2148b5b658ec537eee)

XinfengZhang:
macroblock offset should be calculated base on whether it's align with 8bits

lakulako:
Try instead of #239:
https://github.com/lakulako/media-driver/pull/1
https://github.com/Intel-Media-SDK/MediaSDK/pull/493
macroblock_offset should be exact number of consumed bits after range in bool decoder made >=128.

PR #86, opened 2018-01-19
calculate the real byte offset from macroblock_offset

XinfengZhang:
#test solution for issue #57 


XinfengZhang:
@mypopydev please help to check whether it can fix the VP8 issue of ffmpeg

media_driver/linux/common/codec/ddi/media_ddi_decode_vp8.cpp
```diff
@@ -83,7 +83,7 @@ VAStatus DdiDecodeVP8::ParseSliceParams(
     num_token_partitions                = slcParam->num_of_partitions - 1;
     picParams->CodedCoeffTokenPartition = (num_token_partitions != 8) ? (num_token_partitions >> 1) : 3;
     //macroblock_offset is in unit of bit.
-    picParams->uiFirstMbByteOffset = slcParam->slice_data_offset + slcParam->macroblock_offset >> 3;
+    picParams->uiFirstMbByteOffset = slcParam->slice_data_offset + (slcParam->macroblock_offset + 7) >> 3;
```
fhvwy: This looks wrong - slice_data_offset should be in bytes while macroblock_offset is in bits, while this is treating both as bits (with misleading bracketing).

Also, what is the rounding intended to do?  If the macroblock_offset isn't on a byte boundary then this will point the offset to the byte after the first one, which doesn't sound right.

XinfengZhang: @fhvwy thanks, will add () to ensure the priority of operators, about the rounding, it is because HW requirement. if the macroblock offset is not on the edge of  a byte, HW need to add 1 to shift the position.

XinfengZhang:
@dmitryermilov @onabiull please take a look

dmitryermilov:
Hi @XinfengZhang , can we merge the changes?

Issue #82, opened 2018-01-17
[SKL][MPEG2ENC] MPEG2ENC lead to GPU hang in SKL


mypopydev:
in SKL, mpeg2enc will hang when setting macro block address like this:

  **vslice->macroblock_address = priv->mb_width * slice->index;**
  vslice->num_macroblocks    = priv->mb_width;

but when change the code as:   **vslice->macroblock_address =slice->index**;, it's working, I think media-driver doesn't follow VA-API Spec for  uint32_t macroblock_address in VAEncSliceParameterBufferMPEG2.  (https://github.com/01org/libva/blob/c88000b75602ded8ec24e6ea032f4841a4437049/va/va_enc_mpeg2.h#L278)

XinfengZhang:
will fix it.

PR #87, opened 2018-01-23
change the mb address value to the real MB address

XinfengZhang:
following VAAPI definition. #fix issue #82


onabiull:
Symmetric MediaSDK change: [pr110](https://github.com/Intel-Media-SDK/MediaSDK/pull/110)

Issue #42, opened 2017-12-20
[gstreamer-vaapi-master][iHD_SKL] [HEVC_decode] "h265parse0: Internal data stream error" error message occurs during run h265 decoding cases with iHD driver on SKL


wangzj0601:
Test Env:
============================================
Platform:               SKL
Arch:          		x86_64
Kernel:			4.12.0-rc2
libva:			master branch df544cd5a31e54d4cbd33a391795a8747ddaf789 
                        repo: https://github.com/01org/libva.git
media_driver:		master branch 81796c8a9e44b878e26064c898e0f4730e8220cf
                        repo: https://github.com/intel/media-driver
gmmlib:                 repo: https://github.com/intel/gmmlib
gst-vaapi:		master branch b0d41c5db8c29b3df2a5b23d3da55141d701b992

Reproduce Steps:
============================================
1. build env as above lists or according to wiki https://github.com/intel/media-driver
2. take one case for example, run command
gst-launch-1.0 -q filesrc location=/media/HEVC/MSFT//MSHDRef_Difficult_Edge_06_720x576i25f_lowdelay_P_main.bin ! h265parse ! vaapih265dec ! vaapipostproc format=i420 ! checksumsink2 frame-checksum=FALSE file-checksum=TRUE plane-checksum=FALSE
3. error info. as below
0:00:00.039642921  4763      0x208c230 ERROR       vaapivideomemory gstvaapivideomemory.c:278:map_vaapi_memory: failed to make image current
0:00:00.039683951  4763      0x208c230 ERROR                default video-frame.c:161:gst_video_frame_map_id: failed to map video frame plane 0
0:00:00.039736491  4763      0x208c230 ERROR          checksumsink2 gstchecksumsink.c:381:gst_cksum_image_sink_render:<cksumimagesink0> failed to map frame
ERROR: from element /GstPipeline:pipeline0/GstH265Parse:h265parse0: Internal data stream error.
Additional debug info:
gstbaseparse.c(3597): gst_base_parse_loop (): /GstPipeline:pipeline0/GstH265Parse:h265parse0:
streaming stopped, reason error (-5)
d41d8cd98f00b204e9800998ecf8427e

Tianhaol:
Root Cause: Same as #38

Issue #38, opened 2017-12-20
[gstreamer-vaapi-master][iHD_SKL] [AVC_decode] "h264parse0: Internal data stream error" error message occurs during run h264 decoding cases with iHD driver on SKL


wangzj0601:
Test Env:
============================================
Platform:               SKL
Arch:          		x86_64
Kernel:			4.12.0-rc2
libva:			master branch df544cd5a31e54d4cbd33a391795a8747ddaf789 
                        repo: https://github.com/01org/libva.git
media_driver:		master branch 81796c8a9e44b878e26064c898e0f4730e8220cf
                        repo: https://github.com/intel/media-driver
gmmlib:                 repo: https://github.com/intel/gmmlib
gst-vaapi:		master branch b0d41c5db8c29b3df2a5b23d3da55141d701b992

Reproduce Steps:
============================================
1. build env as above lists or according to wiki https://github.com/intel/media-driver
2. take one case for example, run command
gst-launch-1.0 -q filesrc location=/media//h264/CAPCM1_Sand_E.264 ! h264parse ! vaapih264dec ! vaapipostproc format=i420 ! checksumsink2 frame-checksum=FALSE file-checksum=TRUE plane-checksum=FALSE
3. error info. as below
0:00:00.041966159  2891      0x1520e30 ERROR       vaapivideomemory gstvaapivideomemory.c:278:map_vaapi_memory: failed to make image current
0:00:00.041992987  2891      0x1520e30 ERROR                default video-frame.c:161:gst_video_frame_map_id: failed to map video frame plane 0
0:00:00.042020944  2891      0x1520e30 ERROR          checksumsink2 gstchecksumsink.c:381:gst_cksum_image_sink_render:<cksumimagesink0> failed to map frame
ERROR: from element /GstPipeline:pipeline0/GstH264Parse:h264parse0: Internal data stream error.
Additional debug info:
gstbaseparse.c(3597): gst_base_parse_loop (): /GstPipeline:pipeline0/GstH264Parse:h264parse0:
streaming stopped, reason error (-5)
d41d8cd98f00b204e9800998ecf8427e

XinfengZhang:
are there vaapi level log? which vaapi failed? it can help us to quick check.

Tianhaol:
[...]

PR #113, opened 2018-03-15
add I420 format for vaCreateImage

XinfengZhang:


XinfengZhang:
this is for #38 to #42

PR #130, opened 2018-03-28
[KBL,CFL] Added alpha support KabyLake and CoffeeLake

gkrivor:
Added alpha support of KabyLake and CoffeeLake (using Skylake feature set)

Change-Id: Ib65db1048295f7d1831eeff387817c46275ae33a


oliver-sang:
we will upstream KBL/CFL through official way, so we won't accept this patch. sorry.
you could use it in your own forked branch for test purpose.

PR #138, opened 2018-04-10
Set destructor protected for virtual classes

DavidMarec:
To avoid future mistakes, disallow the deletion of a derivate from its base class.

There might be a need for destructors to be virtual. 

This comes out from an open discussion in Media Driver:
https://github.com/Intel-Media-SDK/MediaSDK/pull/125

Latest compilers disallow the use of non-virtual or non-protected destructor for classes that can be derived. (i.e., with virtual members)

-- _as far as  this is quite hazardous_ --


HeFan2017:
Hi David
Sorry for the delay response.
cm_rt.h is a header in our C-for-Media Development Package, and we need a full validation for the API changes in it. Please expect a long validation duration.
Will update the results here.

HeFan2017:
hi David
We have verified that adding the protected destructors is compatible with old drivers. Can you please also add the destructors in cm_rt_api_os.h in cmrtlib/linux/share in this PR?

DavidMarec:
Done.
But one already patched the first file.
https://github.com/intel/media-driver/commit/eaaf54969947dd8114a10d69a2c7915147486e3b

HeFan2017:
Thanks a lot David. The changes are perfect. Would you resolve the conflict with eaaf549? Or do you prefer me to resolve the conflict from my side?

DavidMarec:
Finally, I set up the destructor virtual for the two last classes as it sounds that there are class factories for these in the media sdk.

HeFan2017:
Change verified. LGTM

PR #139, opened 2018-04-10
fix A/B*B, brackets missing

DavidMarec:
Rewrite A/B\*B to A/(B\*B)
split operation to avoid div0 trap

### DIFF
--- a/media_driver/agnostic/common/cm/cm_kernel_rt.cpp
+++ b/media_driver/agnostic/common/cm/cm_kernel_rt.cpp
@@ -5999,32 +5999,42 @@ int CmKernelRT::UpdateSamplerHeap(CmKernelData *kernelData)
                             if ((iter_next != sampler_heap->end()) && (iter_next->elementType > iter->elementType))
                             {
                                 // Aligns heapOffset to next nearest multiple of sampler size if next sampler is a different element type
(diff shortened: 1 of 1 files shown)


falefa:
Hi David. We use the (A+B-1)/B*B to do the alignment. 

For example:  ( 101 + 10 - 1 )  /  10 * 10 = 110. // This aligns 101 to 110.  
If a bracket is added: ( 101 + 10 – 1 ) / ( 10 * 10 ) = 1.

DavidMarec:
@falefa 
Ok, when I saw
`(A+B-1) / B *B`
which is equal to 
` (A+B -1)`.
I thought is was a mistake and the author wanted to write ` (A+B-1)/B²`.

So far, I do not understand the need of ( `/B*B` ) in this operation.
First, it introduces a possible division by zero.
Second, while B is integer, the compiler may decide to  optimize and then remove it.

falefa:
Hi David. Right, we can add the zero checking in the next check-in for error-proof, although theoretically the "B" part in our implementation should always be a positive number.

Like in your example. A = 101 and B = 6, "A" cannot be divided completely by "B". So what we want to do here, is to make A to the next smallest number that can be divided by B, which is the result: 102. 

So that's the purpose of our calculation in here, and 102 is the exact number that we want.

DavidMarec:
Thanks for the explanation.
Last point, if the zero-checking was not _theoretically_ need, this s not that obvious look at the commit I was responding to.
;)

Issue #150, opened 2018-04-18
X11 on-screen rendering don't work in runtime only environment


dvrogozh:
See:
https://github.com/intel/media-driver/blob/45f063fa2677981725fd5169dd2f042983bcf4ff/media_driver/linux/common/ddi/media_libva_putsurface_linux.h#L34

This is a linker name which is not available if end-user will install only runtime environment (i.e. libva.rpm without libva-devel.rpm). Thus, onscreen rendering will not work. This can be easily tried with:
`sample_decode h264 -i input.264 -r`
where sample_decode is a mediasdk sample (see https://github.com/Intel-Media-SDK/MediaSDK).

PR #151, opened 2018-04-18
Use libva-x11.so.2 soname for dlopen

dvrogozh:
libva-x11.so is a linker name and is not available in the runtime
environment (if libva-devel package is not installed). Thus, on-screen
rendering with X Server will not work. This patch changes to use
soname instead.

Fixes: #150

Change-Id: I82b9943c10a9d8f0e7d12c824ebcddd9c6b6b0bd

### DIFF
--- a/media_driver/linux/common/ddi/media_libva_putsurface_linux.h
+++ b/media_driver/linux/common/ddi/media_libva_putsurface_linux.h
@@ -31,7 +31,7 @@
 #include <va/va_dricommon.h>
 #include "mos_defs.h"
 
-#define LIBVA_X11_NAME "libva-x11.so"
+#define LIBVA_X11_NAME "libva-x11.so.2"
 typedef struct dri_drawable *(*dri_get_drawable_func)(
     VADriverContextP ctx, XID drawable);
 typedef union dri_buffer *(*dri_get_rendering_buffer_func)(


XinfengZhang:
LGTM

media_driver/linux/common/ddi/media_libva_putsurface_linux.h
```diff
@@ -31,7 +31,7 @@
 #include <va/va_dricommon.h>
 #include "mos_defs.h"
 
-#define LIBVA_X11_NAME "libva-x11.so"
+#define LIBVA_X11_NAME "libva-x11.so.2"
```
oliver-sang: what's the libva version scheme? How often it could be updated?

dvrogozh: Read here:
https://github.com/intel/libva/blob/d6fd111e2062bb4732db8a05ed55fc01771087b4/configure.ac#L55

Shortly, libva*.so.x.y, where x=API_MAJOR+1. That's why we now have .so.2 and it will remain the same as long as we will not want to completely refactor LibVA again.

dvrogozh:
@oliver-sang : do we expect anything else to merge the change?

PR #162, opened 2018-05-16
Removed X11 dependency

maxha651:
This solves the same issue as #69, namely being able to build media-driver without x11 support.

I did this by just removing all dependent code, which might be a bit drastic, but the discussion in #69 seems to indicate that this might be an okay solution.

If you still want to keep the X11-dependent stuff, then what would be a good way to make things compile without X11?


maxha651:
Well, not that surprisingly this didn't work at all. libva complains about the missing function...

But what would be a good way to disable X11? Is something like the patch in #69 the right way to go instead?

My use-case is a headless embedded device using Media SDK and it would be nice to not have to include libx11 etc.

XinfengZhang:
vaPutSurface is used only for display, maybe you can remove it even the checks in libva

maxha651:
Thanks, that would probably work. I have postponed this until we know what libva/media-driver versions we'll be using, but I would eventually rather not have to patch this in the final version.

Just removing the dependees is definitely easiest for me, but my main question now is what solution you would consider merging?

XinfengZhang:
from the comments of #254 , we removed the hardcode from Cmakelist

maxha651:
I haven't had time to test this, but I can't see that anything from #254 was actually merged.
It's probably better to move this discussion to an issue anyway so I'm closing this in any case. I will test this when I can and if the problem remains I will open an issue for it instead.

PR #171, opened 2018-05-27
add SFC flag for decode.

fruitboy1226:
we need one VAAPI flag to disable/control SFC in decode. now we using VA_DEC_PROCESSING flag.
maybe we need a new one in the next.

Signed-off-by: Zhang, Owen <owen.zhang@intel.com>


mypopydev:
So this is a build flag, and can't change runtime? I don't think this is a good idea

PR #184, opened 2018-06-07
[ddi] Load libX11.so.6 instead of libX11.so

jonrecker:
libX11.so is a linker name so it is not available if the
libX11-dev package is not installed. As a result, attempting to
render to screen with X server will fail (e.g. sample_decode -r)

This fix changes the code to dlopen the actual library name
instead: libX11.so.6

The same patch has been applied to MediaSDK samples:
https://github.com/Intel-Media-SDK/MediaSDK/issues/210

And a similar issue was addressed in media-driver:
https://github.com/intel/media-driver/issues/150

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -673,7 +673,7 @@ DDI_MEDIA_FORMAT DdiMedia_OsFormatAlphaMaskToMediaFormat(int32_t fourcc, int32_t
 
 #ifndef ANDROID
 
-#define X11_LIB_NAME "libX11.so"
+#define X11_LIB_NAME "libX11.so.6"
 
 /*
  * Close opened libX11.so lib, free related function table.


XinfengZhang:
hi @jonrecker I am not sure whether it is correct. but on my machine , /usr/lib/libX11.so -> libX11.so.6.3.0
 /usr/lib64/libX11.so.6 -> libX11.so.6.3.0. you  mean that if libX11-dev is not installed. libX11.so is not available, but libX11.so.6 is available?

jonrecker:
Yes, that is correct. Installing only the libX11 package will install the actual library libX11.so.6.3.0, and also create a symbolic link libX11.so.6 -> libX11.so.6.3.0.

Installing the developer package libX11-dev will install some additional components and create the link libX11.so -> libX11.so.6.3.0, which is used by the linker.

So currently if the user does not install libX11-dev, media-driver will fail to load because the link for libX11.so does not exist. This is the same basic problem which was addressed in issue https://github.com/intel/media-driver/issues/150

Issue #157, opened 2018-05-02
Compilation fails with GCC 8.0.1


uartie:
libva (master) heads/master-0-gd6fd111e2062 
gmmlib (master) heads/master-0-ga5015343bc93 
intel-media-driver (master) heads/master-0-ge16ab37e9e70 

GCC 8.0.1 is default on recently released Fedora 28.

```
In file included from /opt/media/build/intel-media-driver/media_driver/agnostic/common/renderhal/renderhal.h:36,
                 from /opt/media/build/intel-media-driver/media_driver/agnostic/common/cm/cm_hal.h:29,
                 from /opt/media/build/intel-media-driver/media_driver/linux/common/cm/cm_surface_manager.h:31,
                 from /opt/media/build/intel-media-driver/media_driver/agnostic/common/cm/cm_buffer_rt.cpp:29:
/opt/media/build/intel-media-driver/media_driver/agnostic/common/hw/mhw_render.h: In constructor 'MhwRenderInterface::MhwRenderInterface(MhwMiInterface*, PMOS_INTERFACE, MEDIA_SYSTEM_INFO*, uint8_t)':
/opt/media/build/intel-media-driver/media_driver/agnostic/common/hw/mhw_render.h:601:60: error: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'MHW_RENDER_ENGINE_L3_CACHE_CONFIG' {aka 'struct _MHW_RENDER_ENGINE_L3_CACHE_CONFIG'}; use assignment or value-initialization instead [-Werror=class-memaccess]
         memset(&m_l3CacheConfig, 0, sizeof(m_l3CacheConfig));
                                                            ^
In file included from /opt/media/build/intel-media-driver/media_driver/agnostic/common/renderhal/renderhal.h:36,
                 from /opt/media/build/intel-media-driver/media_driver/agnostic/common/cm/cm_hal.h:29,
                 from /opt/media/build/intel-media-driver/media_driver/linux/common/cm/cm_surface_manager.h:31,
                 from /opt/media/build/intel-media-driver/media_driver/agnostic/common/cm/cm_buffer_rt.cpp:29:
/opt/media/build/intel-media-driver/media_driver/agnostic/common/hw/mhw_render.h:63:16: note: 'MHW_RENDER_ENGINE_L3_CACHE_CONFIG' {aka 'struct _MHW_RENDER_ENGINE_L3_CACHE_CONFIG'} declared here
[...]

PR #199, opened 2018-06-19
[ClearLinux] Fix a bunch of compile issues on ClearLinux

jwang11:
The PR fix below issues found on ClearLinux (gcc 8.1.1), #157 
*) -Werror=strict-aliasing compile issue
*) -Werror=unused-variable compile issues
*) -Wdelete-non-virtual-dtor issue
*) -Werror=re-order compile issue

Via this PR, the media-driver can build successfully on clearlinux with latest libva b3be72a5a110880f70626d7c3bed953cdde124b2


cmrtlib/agnostic/share/cm_queue_base.h
```diff
@@ -59,6 +59,7 @@ const CM_QUEUE_CREATE_OPTION CM_DEFAULT_QUEUE_CREATE_OPTION = { CM_QUEUE_TYPE_RE
 class CmQueue
 {
 public:
+    virtual ~CmQueue() = default;
```
HeFan2017: CmQueue and CmDevice are CM APIs which can be called directly by applications. We can't just insert a destructor here for backward compatibility. We will check whether it can maintain the backward compatibility if we add the destructor in protected domain.

jwang11: Ok, so you suggest to move destructor to protected: domain of class?

HeFan2017: Yes. Directly deleting CmQueue and CmDevice class is forbidden, so moving the destructor in protected: domain can avoid directly deleting these two classes, and can avoid this kind of warning.
We are running a sanity test with your changes + protected destructors. Once it passes, we can merge this change.

HeFan2017: about the protected destructor, you can refer to b9c82263251566ca129566c40c7a934fe5904162

cmrtlib/linux/hardware/cm_device_os.cpp
```diff
@@ -166,7 +166,6 @@ CmDevice_RT::~CmDevice_RT( void )
 {
     if (m_cmCreated)
     {
-        VAStatus vaStatus = vaTerminate(m_vaDisplay);
```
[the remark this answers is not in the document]
HeFan2017: I see. Please just remove the unused variable as we don't need to judge its return value in a destructor. But vaTerminate(m_vaDisplay) still needs to be called here.
So this line better like:
vaTerminate(m_vaDisplay);

cmrtlib/agnostic/share/cm_queue_base.h
HeFan2017: Thanks a lot

cmrtlib/linux/hardware/cm_device_os.cpp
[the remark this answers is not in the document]
jwang11: yes, you are right. vaTerminate function invoke should be kept.

cmrtlib/agnostic/share/cm_printf_host.cpp
```diff
@@ -626,10 +626,16 @@ bool PFParser::outputToken(const char *tkn, PCM_PRINT_HEADER header)
...
                 case CM_PRINT_DATA_TYPE_INT:
-                    m_args[m_numMultArg - m_argsExpected] = *((int*            )&(header->scalar64));
+                    {
+                        int* scalar_int = reinterpret_cast<int*>(&header->scalar64);
+                        m_args[m_numMultArg - m_argsExpected] = *scalar_int;
+                    }
                     break;
                 case CM_PRINT_DATA_TYPE_UINT:
-                    m_args[m_numMultArg - m_argsExpected] = *((unsigned int*   )&(header->scalar64));
+		    {
```
HeFan2017: seems tabs are used here. Can you replace them with 4 spaces? Please also keep the alignment. Thanks a lot

jwang11: My bad,  editor introduce a lot tabs that I don't aware. Fix soon.

cmrtlib/linux/hardware/cm_device_os.cpp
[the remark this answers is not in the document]
chivakker: shouldn't the solution of this be check the value of vaStatus and return error if applicable?

chivakker:
Also, please correct commit message and add "Fixes #xxx" where xxx is the number of the issues really fixed. #157 is not relevant and also remove clearlinux reference, this is not specific to it.

PR #216, opened 2018-06-29
Check whether external libs exist before building driver

walter-bai:
Check whether external libs exist before building driver, if not found, print an error message and abort cmake.

### DIFF
--- a/media_driver/cmake/media_utils.cmake
+++ b/media_driver/cmake/media_utils.cmake
@@ -59,4 +59,5 @@ macro (MediaAddCommonTargetDefines target)
     endif()
 endmacro()
 
+
 include( ${MEDIA_DRIVER_CMAKE}/ext/media_utils_ext.cmake OPTIONAL)
--- a/media_driver/cmake/media_utils.cmake
+++ b/media_driver/cmake/media_utils.cmake
@@ -17,6 +17,7 @@
 # OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
 # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
+include (FindPkgConfig)
 
 # Only can include subdirectory which has a media_srcs.cmake
 # the effect is like include(${CMAKE_CURRENT_LIST_DIR}/<subd>/media_srcs.cmake)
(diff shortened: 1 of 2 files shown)


media_driver/cmake/media_utils.cmake
```diff
@@ -59,5 +60,15 @@ macro (MediaAddCommonTargetDefines target)
     endif()
 endmacro()
 
+# find external libs, if not found, print an error message and abort cmake
+macro (find_external_libs EXTERNAL_LIBS)
```
dvrogozh: I think this macro is not needed. plkg_check_modules  will error out by itself if you call it with REQUIRED like this:

```
--   No package 'XYZ' found
CMake Error at /usr/share/cmake3/Modules/FindPkgConfig.cmake:418 (message):
  A required package was not found
Call Stack (most recent call first):
  /usr/share/cmake3/Modules/FindPkgConfig.cmake:585 (_pkg_check_modules_internal)
  CMakeLists.txt:484 (pkg_check_modules)
```

I.e. this macro gives useless over complexity and redefinies already written function (pkg_check_modules)

media_driver/cmake/media_utils.cmake
```diff
@@ -59,5 +60,15 @@ macro (MediaAddCommonTargetDefines target)
...
+    string (REPLACE " " ";" EXTERNAL_LIBS_LIST "${EXTERNAL_LIBS}")
+    foreach (LIB ${EXTERNAL_LIBS_LIST})
+        pkg_check_modules (${LIB} REQUIRED)
+        if (NOT ${LIB}_FOUND)
+            message (FATAL_ERROR "cannot find enternal lib: ${LIB}")
```
dvrogozh: Are you sure that this code will be executed at all after failure on REQUIRED? I very much doubt.

media_driver/cmake/media_utils.cmake
walter-bai: Thanks for your reminder, this macro is useless, I have removed it.

media_driver/media_top_cmake.cmake
```diff
@@ -21,11 +21,13 @@
 project( media )
 
 bs_set_if_undefined(LIB_NAME iHD_drv_video)
-set (LIB_NAME_EXTERNAL_LIBS "pciaccess m pthread dl rt")
+set (LIB_NAME_EXTERNAL_LIBS "pciaccess;m;pthread;dl;rt")
 
 include (FindPkgConfig)
 # find external libs, if not found, cmake will abort
-pkg_check_modules (pciaccess REQUIRED)
+foreach (LIB ${LIB_NAME_EXTERNAL_LIBS})
```
dvrogozh: I suggest to avoid introduction of additional variable (LIB_NAME_EXTERNAL_LIBS) and have explicit calls like:
```
pkg_check_modules(pciaccess)
pkg_check_modules(ptherad)
```
etc, This will make a script more explicit and easy to support

walter-bai:
I'll cancel this and create a new one.

PR #219, opened 2018-07-04
Check whether external libs exist before building

walter-bai:


media_driver/media_top_cmake.cmake
```diff
@@ -75,6 +75,14 @@ set_target_properties(${LIB_NAME_STATIC} PROPERTIES PREFIX "")
 
 MediaAddCommonTargetDefines(${LIB_NAME_OBJ})
 
+include (FindPkgConfig)
+# find external libs, if not found, cmake will abort
+pkg_check_modules (pciaccess REQUIRED)
```
dvrogozh: This is _not_ functional. Read: https://cmake.org/cmake/help/v3.6/module/FindPkgConfig.html. This change gives a lot of warnings like:
```
-- Checking for modules ''
Must specify package names on the command line
Must specify package names on the command line
```
And in reality you check nothing. What you have tried to do should look like:
```
pkg_check_modules (PKG_PCIACCESS REQUIRED pciaccess)
pkg_check_modules (PKG_M REQUIRED m)
```
And it will give the following output:
```
-- Checking for module 'pciaccess'
--   Found pciaccess, version 0.14
-- Checking for module 'm'
--   No package 'm' found
CMake Error at /usr/share/cmake3/Modules/FindPkgConfig.cmake:418 (message):
  A required package was not found
```

Now, I think I have never saw a system which has pkgconfig for the m, dl, rt and pthread . They simply always present on the system and there is no reason to check for them.

Finally, it actually don't make much sense to check for the library and don't use pkg-config output. With cmake you should have such things as PKG_PCIACCESS_CFLAGS, PKG_PCIACCESS_LDFLAGS, etc. which you _must_ use otherwise it does not make sense to check.

walter-bai: Thanks very much for catching that, how about this change:
```
include (FindPkgConfig)
pkg_check_modules (PKG_PCIACCESS REQUIRED pciaccess)
include_directories (BEFORE ${PKG_PCIACCESS_INCLUDE_DIRS})
link_directories (${PKG_PCIACCESS_LIBRARY_DIRS})
```
Here I only use PKG_PCIACCESS_INCLUDE_DIRS and PKG_PCIACCESS_LIBRARY_DIRS to specify pciaccess headers' and libraries' directories, there is already a command to handle the link:
```
bs_ufo_link_libraries_noBsymbolic(
    ${LIB_NAME}
    "${INCLUDED_LIBS}"
    "pciaccess m pthread dl rt"
)
```

dvrogozh: well, this sounds similar to what is used in other places of media-driver. You may also wish to use:
```
bs_ufo_link_libraries_noBsymbolic(
    ${LIB_NAME}
    "${INCLUDED_LIBS}"
    "${PKG_PCIACCESS_LIBRARIES} m pthread dl rt"
)
```
to fully rely on pkg-config.

Issue #198, opened 2018-06-15
provide a default building configuration


chivakker:
Please provide a default building configuration where the following compiles something that can be used.

workdir/media-driver
workdir/gmmlib

$ cd workdir/build_media
$ cmake -DCMAKE_INSTALL_PREFIX=<my own prefix>
$ make install

This is something MediaSDK project implemented and is really helpful.

oliver-sang:
ok, will work on this.
For CMAKE_INSTALL_PREFIX=, there are actually two libraries in media-driver to compile - iHD (media driver) and cmrtlib driver. iHD is the backend of libva, so in current cmake, the iHD install path is determined by libva config by default. cmrtlib is just a normal library to install to system path. So what's your suggestion how CMAKE_INSTALL_PREFIX controls the install behavior for both iHD and cmrtlib? Thanks

oliver-sang:
@chivakker how about cmrtlib installed directly into CMAKE_INSTALL_PREFIX and iHD installed into CMAKE_INSTALL_PREFIX/dri ?

chivakker:
@oliver-sang yeah that's the default being used with the drivers, it should be ok.  I am also wondering why we need cmrtlib compiled and installed by default if there's no opencl use cases needed for iHD to operate.  Maybe you want to attack this problem later but I thought it was worth mentioning.

dvrogozh:
> I am also wondering why we need cmrtlib compiled and installed by default if there's no opencl use cases needed for iHD to operate

cmrtlib is not related to opencl. For example, mediasdk uses it directly to improve performance or quality in a certain use cases.

Now few speculations:)...
I think cmrt sources are coupled with the driver simply because driver is implementation is based on it, I mean that driver don't use cmrtlib, but it uses core features (cm) which cmrtlib exposes. I.e. sources are in the driver repository, cmrtlib simply exposes them for users. I guess that if we will want to move cmrt out to its own repo we will simply end up duplicating the media driver... Though, it should be possible to make a more deep refactoring and extract out cm completely transforming it to the real driver dependency... Hm.... Need to discuss this with @XinfengZhang and @oliver-sang... Well, I would like to hear from these 2 guys confirmation on what I just wrote... In intel-vaapi-driver you also had this part actually which was a shaders layer and you used some asm-4gen or something similar to compile the asm shaders code - this is analogue of cmrt, though some principals are different...

oliver-sang:
@HeFan2017 could you help comment?

[...]

PR #223, opened 2018-07-06
Resolve 'provide a default building configuration'

esbetan:
Fixes: #198  
 - Default CMAKE_INSTALL_PREFIX to '/usr/local' but allow modifications at run-time
 - Default MEDIA_VERSION / UFO_VERSION to 1.0.0
 - Default gmmlib dependency paths until builds shift to dynamic linked
 - Update README.md to reflect simplified build steps
 - Update distro_cmake.sh & unified_cmake.sh scripts

Change-Id: I9250f6ff07c670a6cdf37ea516513b031508cabb

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,15 @@ include(os_release_info.cmake)
 
 get_os_release_info(os_name os_version)
 
+if(NOT DEFINED CMAKE_INSTALL_PREFIX OR CMAKE_INSTALL_PREFIX STREQUAL "")
+    set(CMAKE_INSTALL_PREFIX "/usr/local" CACHE PATH "..." FORCE)
+endif()
+message(STATUS "CMAKE_INSTALL_PREFIX = ${CMAKE_INSTALL_PREFIX}")
+
+if(NOT DEFINED MEDIA_VERSION OR MEDIA_VERSION STREQUAL "")
+    set(MEDIA_VERSION "1.0.0")
+endif()
+
 if("${os_name}" STREQUAL "clear-linux-os")
     # clear-linux-os distribution avoids /etc for distribution defaults.
     # Set this variable explicitly before including GNUInstallDirs.
--- a/Tools/bldsys/include/bs_dir_names.cmake
+++ b/Tools/bldsys/include/bs_dir_names.cmake
@@ -41,12 +41,12 @@ endif(NOT DEFINED GFX_DEVELOPMENT_DIR)
 #
 # These will be used later to help with separation of components.
 #
-bs_set_if_undefined(BS_DIR_SOURCE          "${GFX_DEVELOPMENT_DIR}/Source")
+bs_set_if_undefined(BS_DIR_SOURCE          "${CMAKE_SOURCE_DIR}/..")
 
-bs_set_if_undefined(BS_DIR_COMMON          "${BS_DIR_SOURCE}/Common")
-bs_set_if_undefined(BS_DIR_GMMLIB          "${BS_DIR_SOURCE}/GmmLib")
-bs_set_if_undefined(BS_DIR_INC             "${BS_DIR_SOURCE}/inc")
+bs_set_if_undefined(BS_DIR_COMMON          "${BS_DIR_SOURCE}/gmmlib/Source/Common")
+bs_set_if_undefined(BS_DIR_GMMLIB          "${BS_DIR_SOURCE}/gmmlib/Source/GmmLib")
+bs_set_if_undefined(BS_DIR_INC             "${BS_DIR_SOURCE}/gmmlib/Source/inc")
 bs_set_if_undefined(BS_DIR_INSTALL         "${BS_DIR_SOURCE}/install")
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -25,6 +25,15 @@ include(os_release_info.cmake)
 
 get_os_release_info(os_name os_version)
 
+if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT OR CMAKE_INSTALL_PREFIX STREQUAL "")
+    set(CMAKE_INSTALL_PREFIX "/usr" CACHE PATH "..." FORCE)
```
dvrogozh: I think it should be '/usr/local' as default. Usual practice is to step out from /usr and avoid rewriting system defaults (you should assume that media-driver is already installed from the system packages).

CMakeLists.txt
```diff
@@ -25,6 +25,15 @@ include(os_release_info.cmake)
...
+endif()
+message(STATUS "CMAKE_INSTALL_PREFIX = ${CMAKE_INSTALL_PREFIX}")
+
+if(NOT DEFINED MEDIA_VERSION OR MEDIA_VERSION STREQUAL "")
+    set(MEDIA_VERSION "1.0.0")
```
dvrogozh: Can we have explicit setting of the version? I.e. version stored here or in the file which we read and which is updated regularly once the version should be bumped. I that way developers will always roughly understand with which version of the code they work.

Tools/bldsys/include/bs_dir_names.cmake
```diff
@@ -41,12 +41,12 @@ endif(NOT DEFINED GFX_DEVELOPMENT_DIR)
 #
 # These will be used later to help with separation of components.
 #
-bs_set_if_undefined(BS_DIR_SOURCE          "${GFX_DEVELOPMENT_DIR}/Source")
+bs_set_if_undefined(BS_DIR_SOURCE          "${CMAKE_SOURCE_DIR}/..")
 
-bs_set_if_undefined(BS_DIR_COMMON          "${BS_DIR_SOURCE}/Common")
-bs_set_if_undefined(BS_DIR_GMMLIB          "${BS_DIR_SOURCE}/GmmLib")
-bs_set_if_undefined(BS_DIR_INC             "${BS_DIR_SOURCE}/inc")
+bs_set_if_undefined(BS_DIR_COMMON          "${BS_DIR_SOURCE}/gmmlib/Source/Common")
```
dvrogozh: I hope that everyone understands that we need to have pkg-config identification for _installed_ GmmLib? Whether it is dynamic or static it actually does not matter.

CMakeLists.txt
oliver-sang: good point! Hi Eric, please reconsider this. I think we actually don't need to do anything, just depending on GNUInstallDirs default setting on a distribution should be fine. If we really need a package which could overwrite system version, just passing CMAKE_INSTALL_PREFIX in cmake command line should be fine.

Tools/bldsys/include/bs_dir_names.cmake
oliver-sang: This is still a temp solution before we can change to use dynamic link GmmLib.

esbetan: Yep, this is temporary. This should be removed when we've switched over to pkg-config.

CMakeLists.txt
esbetan: Ok, i will edit this to just deal with edge case mentioned in #198, ie. CMAKE_INSTALL_PREFIX=

Tools/bldsys/include/bs_dir_names.cmake
dvrogozh: ok, I just wanted to double check target goal:).

CMakeLists.txt
esbetan: Is this location ok? And is this version string ok? Seems like previously, version wasn't defaulted and readme always stuck with '2.0.0'.

esbetan:
Editted my commit with CMAKE_INSTALL_PREFIX edit. Now it only reset if user passes in empty string.

esbetan:
@dvrogozh @oliver-sang Any additional thoughts on this PR?

CMakeLists.txt
dvrogozh: /usr/local location is definitely ok.

CMakeLists.txt
```diff
@@ -25,6 +25,15 @@ include(os_release_info.cmake)
 
 get_os_release_info(os_name os_version)
 
+if(NOT DEFINED CMAKE_INSTALL_PREFIX OR CMAKE_INSTALL_PREFIX STREQUAL "")
```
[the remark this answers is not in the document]
dvrogozh: but what user wants to achieve with this? and how we should interpret this? From my perspective empty prefix is simply wrong and we either are free to overwrite or return a fatal error.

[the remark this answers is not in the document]
esbetan: I added this to address issue #198

esbetan:
@oliver-sang Done, distro_cmake.sh updated.

PR #226, opened 2018-07-10
update README for 18.2 release

oliver-sang:
Change-Id: Ie644b3bf80c657643b1a1a4601af83414dcde92b

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,24 +101,52 @@ BDW (Broadwell)
 
 SKL (Skylake)
 
-BXT (Broxton) / APL (Apollolake)
+BXT (Broxton) / APL (Apollo Lake)
+
+KBL (Kaby Lake)
+
+CFL (Coffee Lake)
 
 CNL (Cannonlake)
 
 ## Supported Codecs
 
-| CODEC      | BDW | SKL | BXT/APL | CNL |
-|------------|-----|-----|---------|-----|
-| H.264      | D/E | D/E |   D/E   | D/E |
-| MPEG-2     | D/E | D/E |    D    | D/E |
-| VC-1       |  D  |  D  |    D    |  D  |
-| JPEG       |  D  | D/E |   D/E   | D/E |
-| VP8        |  D  |  D  |    D    | D/E |
-| HEVC       |     | D/E |   D/E   | D/E |
-| HEVC 10bit |     |     |    D    | D/E |
-| VP9        |     |     |    D    | D/E |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,5 +166,11 @@ power encoding, and VP9 low power encoding require the kernel patch for GuC supp
 
 5. CNL: HEVC encoding does not support P frame.
 
+6. CFL: BRC functionalities requiring HuC for AVC low power encoding require 4.15 or later kernel to work.
```
dvrogozh: 4.11 and 4.15 looks ok. I think we need to mention kernel boot parameter to access HuC related features. This parameter looks to be i915.enable_guc=2 (bitmask, loading GuC is a prerequisite for HuC).

Can you reformulate what "BRC functionalities requiring HuC" means? I would like to see something like:

> The following features require HuC FW support:
> * CBR, VBR, .... (all BRC modes, CQP, ... are not affected) modes of AVC low power encoding
> * other feature, if any
> * other feature, if any
>
> As of now HuC FW support is disable in Linux kernels by default. To enable it load the kernel with `i915.enable_guc=2` (bitmask, loading GuC is a prerequisite for HuC) parameter. Mind that HuC FW support presents in the following kernels for the specified platforms:
> * APL, KBL: starting from kernel 4.11
> * CFL: starting from kernel 4.15

Thus, I would group all the notes about HuC together. By the way, why you don't mention SKL?

CNL: I don't understand where I can get HuC FW! I don't see it in the git://anongit.freedesktop.org/drm/drm-firmware. Thus, maybe we need to add a note for CNL like:
> CBR, VBR, etc. modes of AVC low power encoding are not supported (HuC is not available for CNL)

README.md
```diff
@@ -101,24 +101,52 @@ BDW (Broadwell)
...
+E1 - VME based encoding
+
+E2 - Low power encoding
+
+## Supported Video Processing
+
+| Video Processing                             | BDW | SKL | BXT/APL | KBL | CFL | CNL |
+|----------------------------------------------|-----|-----|---------|-----|-----|-----|
+| Blending                                     |  Y  |  Y  |    Y    |  Y  |  Y  |  Y  |
+| CSC (Color Space Conversion)                 |  Y  |  Y  |    Y    |  Y  |  Y  |  Y  |
```
mypopydev: Can you list the full list for CSC input/output format, e,g  NV12 -> RGB...

PR #227, opened 2018-07-12
[ffmpeg-vaapi][mpeg2 encode] make ffmpeg's mpeg2 vaapi

jsunintel:
encode work by f_code and macroblock_address changes.

Signed-off-by: Jing SUN <jing.a.sun@intel.com>

### DIFF
--- a/media_driver/agnostic/common/codec/hal/codechal_encode_mpeg2.cpp
+++ b/media_driver/agnostic/common/codec/hal/codechal_encode_mpeg2.cpp
@@ -1825,28 +1825,41 @@ MOS_STATUS CodechalEncodeMpeg2::SetPictureStructs()
 
     if (m_pictureCodingType == I_TYPE)
     {
-        m_picParams->m_fcode00 = fcodeX;
-        m_picParams->m_fcode01 = fcodeY;
-
+        if ((m_picParams->m_fcode00 > fcodeX) ||
+            (m_picParams->m_fcode01 > fcodeY) ||
+            (m_picParams->m_fcode00 == 0) ||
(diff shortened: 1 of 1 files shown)


XinfengZhang:
please refer https://github.com/intel/media-driver/blob/master/media_driver/agnostic/common/codec/hal/codechal_encode_mpeg2.cpp#L1823
and the value should can be smaller than the fcodeX fcodeY

XinfengZhang:
yes, sure, they are same, but this value means the maximum search area, maybe customer want to set smaller value to narrow the search range. thanks

XinfengZhang:
and why not change the codechal code to avoid duplicated statements

jsunintel:
The values recommended by the H264 standard are as below:

    switch (avctx->level) {
    case 4: // High.
    case 6: // High 1440.
        priv->f_code_horizontal = 9;
        priv->f_code_vertical   = 5;
        break;
    case 8: // Main.
        priv->f_code_horizontal = 8;
        priv->f_code_vertical   = 5;
        break;
    case 10: // Low.
    default:
        priv->f_code_horizontal = 7;
        priv->f_code_vertical   = 4;
        break;
    }

But those don't meet the HW requirement. Hence, we have a set of values as shown in this patch. MSDK follows this and set such values in its lib as well.  We have to overwrite the values chosen by the ffmpeg tool in the UMD, to make the encode work.

Also, SetPictureStructs is for initialization, and we can't ask the applications not to set the f_code values at runtime for each frame. That' why I added the code lines there.

jsunintel:
Thanks a lot for Carl's review comments! Just updated it and added ==0 checking based on his kind reminder. Unit tests passed.

Issue #228, opened 2018-07-12
vaCreateBuffer() issue when source data buffer is provided.


djie1:
When vaCreateBuffer() is called with the source data buffer provided (data != NULL), and there are multiple slices in the frame. Issue would happen that the existing BitStreamBuffObject is oversized for the second or following slices.
As for the second or following slices, buf->pData is assigned one external buffer, but calling mos_bo_subdata() would keep copying the data to the internal buffer.

PR #229, opened 2018-07-12
[vaCreateBuffer] one fixing for issue in vaCreateBuffer().

djie1:
Here is the case, when vaCreateBuffer() is called with source data
pointer provided, and there are multiple slices in the frame, buf->pData
is assigned a different buffer for the second or following slices if
bIsSliceOverSize is true, but the calling of mos_bo_subdata() for
the second slices would copy the data to the buffer of the first slice.


djie1:
This workaround could fix #228

Issue #233, opened 2018-07-13
[ffmpeg-vaapi][vpp]vpp scaling with 4K input surface failure.


dlin2:
### COMMAND:
 ffmpeg -vaapi_device /dev/dri/renderD128 -hwaccel vaapi -i /4K.h264 -c:v h264_vaapi -vf 'format=nv12|vaapi,hwupload,scale_vaapi=w=720:h=480' vpp_scaling.h264
### ERROR INFO
[LIBVA]:CRITICAL - DdiVp_GetColorSpace:1450: Invalid surface color standard
[VP]:   CRITICAL - RenderPhase:5752: Failed to create filter description.
[VP]:   CRITICAL - RenderMultiPhase:1997: MOS returned error, eStatus = 0x19
[VP]:   CRITICAL - Render:2329: MOS returned error, eStatus = 0x19
[VP]:   CRITICAL - RenderComposite:825: MOS returned error, eStatus = 0x19
[VP]:   CRITICAL - RenderPass:714: MOS returned error, eStatus = 0x19
[VP]:   CRITICAL - Render:1053: MOS returned error, eStatus = 0x19
[VP]:   CRITICAL - DdiVp_EndPicture:2851: Failed to call render function.
[Parsed_scale_vaapi_2 @ 0x4d82400] Failed to start picture processing: 1 (operation failed).
[VP]:   CRITICAL - RenderPhase:5752: Failed to create filter description.
[VP]:   CRITICAL - RenderMultiPhase:1997: MOS returned error, eStatus = 0x19
[VP]:   CRITICAL - Render:2329: MOS returned error, eStatus = 0x19
[VP]:   CRITICAL - RenderComposite:825: MOS returned error, eStatus = 0x19
[VP]:   CRITICAL - RenderPass:714: MOS returned error, eStatus = 0x19
[VP]:   CRITICAL - Render:1053: MOS returned error, eStatus = 0x19
[VP]:   CRITICAL - DdiVp_EndPicture:2851: Failed to call render function.
Error while filtering: Input/output error
Failed to inject frame into filter network: Input/output error
Error while processing the decoded data fo

dlin2:
49d7247 fixed the issue.

PR #234, opened 2018-07-13
[VP] fixed vpp scaling failure issue for 4K input surface

dlin2:


dlin2:
this is a fix for #233

XinfengZhang:
49d724783c66c4b516645ca751578025329ca086 should be duplicated with this PR

Issue #190, opened 2018-06-14
[ffmpeg-vaapi][vp8 decode] almost all vp8 decode test cases failed when using ffmpeg-vaapi


FocusLuo:
These test cases can work on intel-vaapi-driver
One CMD line shown as below:
ffmpeg -hwaccel vaapi -hwaccel_device /dev/dri/renderD128 -v verbose -y -i ./vp8_vpg/vp8_rev0_I_frame_H_Anim.webm -pix_fmt yuv420p -f md5 ffmpeg_vaapi.md5

The log info shown as below:
ffmpeg version N-91230-gb748772 Copyright (c) 2000-2018 the FFmpeg developers
  built with gcc 5.4.0 (Ubuntu 5.4.0-6ubuntu1~16.04.9) 20160609
  configuration: --enable-vaapi --prefix=/opt/X11R7/ffmpeg_mainline
  libavutil      56. 18.102 / 56. 18.102
  libavcodec     58. 19.105 / 58. 19.105
  libavformat    58. 17.100 / 58. 17.100
  libavdevice    58.  4.100 / 58.  4.100
  libavfilter     7. 25.100 /  7. 25.100
  libswscale      5.  2.100 /  5.  2.100
  libswresample   3.  2.100 /  3.  2.100
Input #0, matroska,webm, from '/media//vp8_vpg/vp8_rev0_I_frame_H_Anim.webm':
  Metadata:
    encoder         : vpxenc v0.9.6
  Duration: 00:00:00.03, start: 0.000000, bitrate: 146 kb/s
    Stream #0:0(eng): Video: vp8, 1 reference frame, yuv420p(progressive), 32x32, SAR 1:1 DAR 1:1, 30 fps, 30 tbr, 1k tbn, 1k tbc (default)
[AVHWDeviceContext @ 0x25fba80] Opened VA display via DRM device /dev/dri/renderD128.
[AVHWDeviceContext @ 0x25fba80] libva: VA-API version 1.1.0
[AVHWDeviceContext @ 0x25fba80] libva: va_getDriverName() returns 0
[AVHWDeviceContext @ 0x25fba80] libva: User requested driver 'iHD'
[AVHWDeviceContext @ 0x25fba80] libva: Trying to open /opt/X11R7/vpg_driver/lib/dri//iHD_drv_video.so
[AVHWDeviceContext @ 0x25fba80] libva: Found init function __vaDriverInit_1_1
[...]

PR #239, opened 2018-07-18
fix vp8 decode issue #190, change the first partition size

XinfengZhang:
when the count in VABoolCoderContextVPX is zero, actually it means there are 8 bit in bool coder

correct first partition_size,when there is 8 bits in bool coder

### DIFF
--- a/media_driver/linux/common/codec/ddi/media_ddi_decode_vp8.cpp
+++ b/media_driver/linux/common/codec/ddi/media_ddi_decode_vp8.cpp
@@ -87,9 +87,8 @@ VAStatus DdiDecodeVP8::ParseSliceParams(
     picParams->uiFirstMbByteOffset = slcParam->slice_data_offset + ((slcParam->macroblock_offset + 8) >> 3);
 
     memcpy_s(picParams->uiPartitionSize, sizeof(picParams->uiPartitionSize), slcParam->partition_size, sizeof(picParams->uiPartitionSize));
-
-    //partition 0 size in command buffer includes the one byte in bool decoder if remaining bits of bool decoder is not zero.
-    picParams->uiPartitionSize[0] += (slcParam->macroblock_offset & 0x7) ? 1 : 0;
+    //partition 0 size in command buffer includes the one byte in bool decoder if remaining bits of bool decoder is zero.
+    picParams->uiPartitionSize[0] -= (slcParam->macroblock_offset & 0x7) ? 0 : 1;
 
     return VA_STATUS_SUCCESS;
 }


media_driver/linux/common/codec/ddi/media_ddi_decode_vp8.cpp
```diff
@@ -84,12 +84,17 @@ VAStatus DdiDecodeVP8::ParseSliceParams(
     picParams->CodedCoeffTokenPartition = (num_token_partitions != 8) ? (num_token_partitions >> 1) : 3;
     //macroblock_offset is in unit of bit.it should be always the next byte, the byte is divided to two parts
     //used bits and remaining bits, if used bits == 8, uiFirstMbByteOffset should add 1, so use 8 to do the ceil operator
```
dmitryermilov: what about the comment above, is it still actual ? especially, I'm confused about this "if used bits == 8, uiFirstMbByteOffset should add 1, so use 8 to do the ceil operator"

media_driver/linux/common/codec/ddi/media_ddi_decode_vp8.cpp
```diff
@@ -84,12 +84,17 @@ VAStatus DdiDecodeVP8::ParseSliceParams(
...
-    picParams->uiFirstMbByteOffset = slcParam->slice_data_offset + ((slcParam->macroblock_offset + 8) >> 3);
+    picParams->uiFirstMbByteOffset = slcParam->slice_data_offset + ((slcParam->macroblock_offset + 7) >> 3);
 
     memcpy_s(picParams->uiPartitionSize, sizeof(picParams->uiPartitionSize), slcParam->partition_size, sizeof(picParams->uiPartitionSize));
 
     //partition 0 size in command buffer includes the one byte in bool decoder if remaining bits of bool decoder is not zero.
-    picParams->uiPartitionSize[0] += (slcParam->macroblock_offset & 0x7) ? 1 : 0;
+    //bool coder count should be 0-7, if it is 0 (ucP0EntropyCount == 0), actually it means there are 8 bit is loaded in bool coder.
+    if(picParams->ucP0EntropyCount == 0)
```
lakulako: Doubt it works. Because after Vp8EntropyState::DecodeFill() m_count is between 16 and 24. After some decoding it is 0<=m_count<24. Then in ParseFrameHead() I see vp8PicParams->ucP0EntropyCount = 8 - (m_count & 0x07); here byte-offset part of m_count is lost. The range is [1;8].
Then here in 'if (picParams->ucP0EntropyCount == 0)' byte offset is partially lost and the condition can never be true.

lakulako: Alternative is
https://github.com/Intel-Media-SDK/MediaSDK/pull/493
https://github.com/lakulako/media-driver/pull/1

XinfengZhang: mostlikely , we should calculate the value and bool coder remaining bits together

PR #257, opened 2018-08-01
[JPEG DEC] change the SFC flag for jpeg decode.

fruitboy1226:
only the input fmt is RGB, will go SFC path, otherwise disable SFC.

Change-Id: Ib07ed5aebeddbe75b7375809316ab22d989f7c05
Signed-off-by: Owen Zhang <owen.zhang@intel.com>


Sherry-Lin:
@fruitboy1226 Could you please rebase this patch? Thanks.

PR #259, opened 2018-08-02
compile: fix undeclared type MEDIA_FEATURE_TABLE on clang

xuguangxin:


DenWolf:
👍 - this change solved the problem with compilation (undeclared type MEDIA_FEATURE_TABLE) - Android P master tree

PR #263, opened 2018-08-03
[VP]Add platform information in DdiMedia_QueryVideoProcPipelineCaps

### LINKED ISSUE #201: Driver doesn't report VPP maxWidth/maxHeight via vaQueryVideoProcPipelineCaps
Driver doesn't report VPP maxWidth/maxHeight via vaQueryVideoProcPipelineCaps for EUs (e.g. scaling) and VEBOX usage cases. See code below:
https://github.com/intel/media-driver/blob/master/media_driver/linux/common/ddi/media_libva.cpp#L5343-L5363

The driver fills width/height/pixel_format caps **only** for SFC cases.

Jister:
Fix #201, add platform information in DdiMedia_QueryVideoProcPipelineCaps to distinguish Gen8 and Gen9+ max width and height

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -5311,6 +5311,7 @@ DdiMedia_QueryVideoProcPipelineCaps(
 
     if ((context & DDI_MEDIA_MASK_VACONTEXT_TYPE) == DDI_MEDIA_VACONTEXTID_OFFSET_DECODER)
     {
+        //Decode+SFC, go SFC path, the restriction here is the capability of SFC
         pipeline_caps->num_input_pixel_formats    = 1;
         pipeline_caps->input_pixel_format[0]      = VA_FOURCC_NV12;
         pipeline_caps->num_output_pixel_formats   = 1;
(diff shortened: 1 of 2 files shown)


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -5334,12 +5334,21 @@ DdiMedia_QueryVideoProcPipelineCaps(
     }
     else if ((context & DDI_MEDIA_MASK_VACONTEXT_TYPE) == DDI_MEDIA_VACONTEXTID_OFFSET_VP)
     {
-        pipeline_caps->max_input_width            = VP_MAX_PIC_WIDTH;
-        pipeline_caps->max_input_height           = VP_MAX_PIC_HEIGHT;
+        if(mediaCtx->platform.eRenderCoreFamily == IGFX_GEN8_CORE)
```
dvrogozh: if we will suddenly decide to support something old like HSW, we will fall to the Gen9+ case:(. Well, I understand that this is unlikely, but still catches an eye. What was max resolution for the previous platforms? If they all supported the same, it may be easier to add it here. Just in case...

Don't consider this as a critical comment...

media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -5334,12 +5334,21 @@ DdiMedia_QueryVideoProcPipelineCaps(
     }
```
dvrogozh: Question to the above if:
`if ((context & DDI_MEDIA_MASK_VACONTEXT_TYPE) == DDI_MEDIA_VACONTEXTID_OFFSET_DECODER) {}`
I wonder whether this if can be hit on BDW and what it will actually report? There is no condition like if isSFC()... I wonder whether VACONTEXTID_OFFSET_DECODER assumes SFC or it may work with the AVS as well? If AVS possible, then a lot of SFC defines look suspicious. If only SFC is possible, I suggest to comment this explicitly.

media_driver/linux/common/ddi/media_libva.cpp
Jister: For previous platforms, the max supported resolution should be same as Gen8, i.e. 16352.  The driver now only support Gen8+ platform, so we didn't consider previous platform here.

media_driver/linux/common/ddi/media_libva.cpp
Jister: This if will be hit when the case is decode+SFC, it will only go through SFC path, so here the restriction here is the max resolution of SFC. I'll comment it here to make it more explicitly, thanks.

Jister:
Sure, thank you for the suggestion.

PR #270, opened 2018-08-08
switch to dynamic gmmlib

dvrogozh:
Fixes: #246, #35

Signed-off-by: Yang, Ning <ning.yang@intel.com>
Signed-off-by: Dmitry Rogozhkin <dmitry.v.rogozhkin@intel.com>

PR #274, opened 2018-08-14
[upstream] switch to dynamic gmm lib

yangnin:
Signed-off-by: Yang, Ning <ning.yang@intel.com>


media_driver/media_top_cmake.cmake
```diff
@@ -81,19 +81,34 @@ bs_ufo_link_libraries_noBsymbolic(
...
-    # dep libs (gmmlib for now) can be passed through INCLUDED_LIBS, but if not, we need try to setup dep through including dep projects
-    if (NOT TARGET gmm_umd)
-        add_subdirectory("${BS_DIR_GMMLIB}" "${CMAKE_BINARY_DIR}/gmmlib")
+
+    # checking gmm dependencies
+    pkg_check_modules(LIBGMM igdgmm)
+
+    if(LIBGMM_FOUND)
+        include_directories(BEFORE ${LIBGMM_INCLUDE_DIRS})
+        link_directories(${LIBGMM_LIBRARY_DIRS})
```
dvrogozh: You can't call link_directories() after add_library(): it will not take effect. You can see this if you will configure and install gmmlib to the custom location. I get this error with your PR:
`/usr/bin/ld: cannot find -ligdgmm`

I have fixed this in my PR: https://github.com/intel/media-driver/pull/270. Please, take a look

media_driver/media_top_cmake.cmake
```diff
@@ -81,19 +81,34 @@ bs_ufo_link_libraries_noBsymbolic(
...
+
+        if (TARGET igfx_gmmumd_dll)
+            target_link_libraries ( ${LIB_NAME}
+                igfx_gmmumd_dll
+            )
+        endif()
     endif()
 
-include(${MEDIA_DRIVER_CMAKE}/ext/media_feature_include_ext.cmake OPTIONAL)
+    target_compile_definitions(${LIB_NAME} PUBLIC GMM_LIB_DLL)
```
dvrogozh: This is not needed because igdgmm.pc have this:
https://github.com/intel/gmmlib/blob/4f787cc7c47b5b79ff20c264852f1213f21e898d/Source/GmmLib/igdgmm.pc.in#L9

media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -1398,12 +1398,15 @@ VAStatus DdiMedia__Initialize (
     output_dri_init(ctx);
 #endif
 
+    void *gmm_handle = dlopen(GMM_UMD_DLL, RTLD_NOW);
```
dvrogozh: You _link_ against igdgmm, you don't need to dlopen() it and search for symbols.

PR #278, opened 2018-08-16
Fix abort issue

xhaihao:


xhaihao:
Rebased against the latest intel:master branch

PR #280, opened 2018-08-16
Fix a ULT fail issue when driver dynamically links gmmlib.

walter-bai:
Fix a ULT fail issue when driver dynamically links gmmlib.

### DIFF
--- a/media_driver/linux/ult/ult_app/driver_loader.h
+++ b/media_driver/linux/ult/ult_app/driver_loader.h
@@ -87,7 +87,7 @@ struct DriverSymbols
     void Clear()
     {
         auto p = (const void **)this;
-        for (auto i = 0; i < sizeof(this) / sizeof(const void *); i++)
+        for (auto i = 0; i < sizeof(*this) / sizeof(const void *); i++)
         {
             p[i] = nullptr;
         }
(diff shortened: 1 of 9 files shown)


media_driver/linux/ult/ult_app/driver_loader.h
```diff
@@ -87,7 +87,7 @@ struct DriverSymbols
     void Clear()
```
dvrogozh: Folks, I strongly suggest to remove constructor and Clear() entirely, You have class with all public members C++11 already has a way to reset a structure to default value-initialization which is: m_drvSyms = {}.

media_driver/linux/ult/ult_app/driver_loader.h
```diff
@@ -96,7 +96,7 @@ struct DriverSymbols
     bool Initialized() const
     {
         auto p = (const void * const *)this;
-        for (auto i = 0; i < sizeof(this) / sizeof(const void *); i++)
+        for (auto i = 0; i < sizeof(*this) / sizeof(const void *); i++)
```
dvrogozh: Folks, please, remove this and check pointers directly. This is simply a freak over complex implementation of very simple thing: check 5 pointers that they are not NULL. You don't need any loop here!!

media_driver/linux/ult/ult_app/driver_loader.h
walter-bai: Thank you for catching this, "m_drvSyms = {}" is much better than a Clear function.

media_driver/linux/ult/ult_app/driver_loader.h
walter-bai: Yes, checking 5 pointers is simple and easy to understand than previous code.

yangnin:
It has been combined with PR#270. close it.

PR #283, opened 2018-08-20
Enable virtual engine (single pipe) on Gen11.

zxye:
To enable virtual engine for Gen11.

### DIFF
--- a/media_driver/linux/common/os/i915/include/mos_bufmgr.h
+++ b/media_driver/linux/common/os/i915/include/mos_bufmgr.h
@@ -60,6 +60,7 @@ struct mos_linux_context {
 #ifndef ANDROID
     struct _MOS_OS_CONTEXT    *pOsContext;
 #endif
+    struct drm_i915_gem_vm_control* vm;
 };
 
 struct mos_linux_bo {
@@ -240,6 +241,27 @@ int mos_get_aperture_sizes(int fd, size_t *mappable, size_t *total);
 int mos_bufmgr_gem_get_devid(struct mos_bufmgr *bufmgr);
 
 struct mos_linux_context *mos_gem_context_create(struct mos_bufmgr *bufmgr);
+struct mos_linux_context *mos_gem_context_create_ext(
+                            struct mos_bufmgr *bufmgr,
+                            __u32 flags);
+struct mos_linux_context *mos_gem_context_create_shared(
+                            struct mos_bufmgr *bufmgr,
+                            mos_linux_context* ctx,
+                            __u32 flags);
+struct drm_i915_gem_vm_control* mos_gem_vm_create(struct mos_bufmgr *bufmgr);
+void mos_gem_vm_destroy(struct mos_bufmgr *bufmgr, struct drm_i915_gem_vm_control* vm);
+
+#define MAX_ENGINE_INSTANCE_NUM 8
+
+int mos_query_engines(int fd,
+                      __u16 engine_class,
+                      __u64 caps,
+                      unsigned int *nengine,
+                      struct i915_engine_class_instance *ci);
+int mos_set_context_param_load_balance(struct mos_linux_context *ctx,
+                         struct i915_engine_class_instance *ci,
+                         unsigned int count);
+
 void mos_gem_context_destroy(struct mos_linux_context *ctx);
 int mos_gem_bo_context_exec(struct mos_linux_bo *bo, struct mos_linux_context *ctx,
                   int used, unsigned int flags);
--- a/media_driver/linux/common/os/i915/mos_bufmgr.c
+++ b/media_driver/linux/common/os/i915/mos_bufmgr.c
@@ -4691,3 +4691,225 @@ exit:
 
     return bufmgr_gem != nullptr ? &bufmgr_gem->bufmgr : nullptr;
 }
+
+struct mos_linux_context *
+mos_gem_context_create_ext(struct mos_bufmgr *bufmgr, __u32 flags)
+{
+    struct mos_bufmgr_gem *bufmgr_gem = (struct mos_bufmgr_gem *)bufmgr;
+    struct drm_i915_gem_context_create_ext create;
+    struct mos_linux_context *context = nullptr;
+    int ret;
+
+    context = (struct mos_linux_context *)calloc(1, sizeof(*context));
+    if (!context)
+        return nullptr;
+
+    memclear(create);
+    create.flags = flags;
+    create.extensions = 0;
+    ret = drmIoctl(bufmgr_gem->fd, DRM_IOCTL_I915_GEM_CONTEXT_CREATE_EXT, &create);
+    if (ret != 0) {
+        MOS_DBG("DRM_IOCTL_I915_GEM_CONTEXT_CREATE failed: %s\n",
+            strerror(errno));
+        free(context);
+        return nullptr;
+    }
+
+    context->ctx_id = create.ctx_id;
+    context->bufmgr = bufmgr;
+
+    return context;
+}
+
+struct drm_i915_gem_vm_control* mos_gem_vm_create(struct mos_bufmgr *bufmgr)
+{
+    struct mos_bufmgr_gem *bufmgr_gem = (struct mos_bufmgr_gem *)bufmgr;
+    struct drm_i915_gem_vm_control *vm = nullptr;
+    int ret;
+
+    vm = (struct drm_i915_gem_vm_control *)calloc(1, sizeof(*vm));
+    memset(vm, 0, sizeof(*vm));
+
+    ret = drmIoctl(bufmgr_gem->fd, DRM_IOCTL_I915_GEM_VM_CREATE, vm);
+    if (ret != 0) {
+        MOS_DBG("DRM_IOCTL_I915_GEM_VM_CREATE failed: %s\n",
+            strerror(errno));
+        return nullptr;
+    }
+
+    return vm;
+}
+
+void mos_gem_vm_destroy(struct mos_bufmgr *bufmgr, struct drm_i915_gem_vm_control* vm)
+{
+    struct mos_bufmgr_gem *bufmgr_gem = (struct mos_bufmgr_gem *)bufmgr;
+    assert(vm);
+    int ret;
+
+    ret = drmIoctl(bufmgr_gem->fd, DRM_IOCTL_I915_GEM_VM_DESTROY, vm);
+    if (ret != 0) {
+        MOS_DBG("DRM_IOCTL_I915_GEM_VM_DESTROY failed: %s\n",
+            strerror(errno));
+    }
+    free(vm);
+}
+
+struct mos_linux_context *
+mos_gem_context_create_shared(struct mos_bufmgr *bufmgr, mos_linux_context* ctx, __u32 flags)
+{
+    struct mos_bufmgr_gem *bufmgr_gem = (struct mos_bufmgr_gem *)bufmgr;
+    struct drm_i915_gem_context_create_ext create;
+    struct mos_linux_context *context = nullptr;
+    int ret;
+
+    if (ctx == nullptr || ctx->vm == nullptr)
+        return nullptr;
+
+    context = (struct mos_linux_context *)calloc(1, sizeof(*context));
+    if (!context)
+        return nullptr;
+
+    memclear(create);
+    create.flags = flags;
+    create.extensions = 0;
+    ret = drmIoctl(bufmgr_gem->fd, DRM_IOCTL_I915_GEM_CONTEXT_CREATE_EXT, &create);
+    if (ret != 0) {
+        MOS_DBG("DRM_IOCTL_I915_GEM_CONTEXT_CREATE failed: %s\n",
+            strerror(errno));
+        free(context);
+        return nullptr;
+    }
+
+    context->ctx_id = create.ctx_id;
+    context->bufmgr = bufmgr;
+
+    ret = mos_set_context_param(context,
+                0,
+                I915_CONTEXT_PARAM_VM,
+                ctx->vm->vm_id);
+    if(ret != 0) {
+        MOS_DBG("I915_CONTEXT_PARAM_VM failed: %s\n",
+            strerror(errno));
+        free(context);
+        return nullptr;
+    }
+
+    return context;
+}
+
+int mos_query_engines(int fd,
+                      __u16 engine_class,
+                      __u64 caps,
+                      unsigned int *nengine,
+                      struct i915_engine_class_instance *ci)
+{
+    struct drm_i915_query query;
+    struct drm_i915_query_item query_item;
+    struct drm_i915_query_engine_info *engines = nullptr;
+    int ret, len;
+
+    memclear(query_item);
+    query_item.query_id = DRM_I915_QUERY_ENGINE_INFO;
+    query_item.length = 0;
+    memclear(query);
+    query.num_items = 1;
+    query.items_ptr = (uintptr_t)&query_item;
+
+    ret = drmIoctl(fd, DRM_IOCTL_I915_QUERY, &query);
+    if (ret)
+    {
+        goto fini;
+    }
+    len = query_item.length;
+
+    engines = (drm_i915_query_engine_info *)malloc(len);
+    memset(engines,0,len);
+    memclear(query_item);
+    query_item.query_id = DRM_I915_QUERY_ENGINE_INFO;
+    query_item.length = len;
+    query_item.data_ptr = (uintptr_t)engines;
+    memclear(query);
+    query.num_items = 1;
+    query.items_ptr = (uintptr_t)&query_item;
+
+    ret = drmIoctl(fd, DRM_IOCTL_I915_QUERY, &query);
+    if (ret)
+    {
+        goto fini;
+    }
+
+    int i, num;
+    for (i = 0, num = 0; i < engines->num_engines; i++) {
+        struct drm_i915_engine_info *engine =
+            (struct drm_i915_engine_info *)&engines->engines[i];
+        if ( engine_class == engine->engine.engine_class
+             && ((caps & engine->capabilities) == caps ))
+        {
+            ci->engine_class = engine_class;
+            ci->engine_instance = engine->engine.engine_instance;
+            ci++;
+            num++;
+        }
+        if (num > *nengine)
+        {
+            fprintf(stderr,"%s: Number of engine instances out of range, %d,%d\n",
+                    __FUNCTION__, num, *nengine);
+            goto fini;
+        }
+    }
+    *nengine = num;
+
+
+fini:
+    if (engines)
+        free(engines);
+    return ret;
+}
+
+int mos_set_context_param_load_balance(struct mos_linux_context *ctx,
+                     struct i915_engine_class_instance *ci,
+                     unsigned int count)
+{
+    int ret;
+    uint32_t size;
+    struct i915_context_engines_load_balance* balancer = nullptr;
+    struct i915_context_param_engines* set_engines = nullptr;
+
+    assert(ci);
+
+    /* I915_DEFINE_CONTEXT_ENGINES_LOAD_BALANCE */
+    size = sizeof(struct i915_context_engines_load_balance) + count * sizeof(*ci);
+    balancer = (struct i915_context_engines_load_balance*)malloc(size);
+    if (NULL == balancer)
+    {
+        ret = -ENOMEM;
+        goto fini;
+    }
+    memset(balancer, 0, size);
+    balancer->base.name = I915_CONTEXT_ENGINES_EXT_LOAD_BALANCE;
+    balancer->num_siblings = count;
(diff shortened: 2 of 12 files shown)


zxye:
This PR requires the i915 virtual engine and GT discovery uAPI provided  by patch series  https://patchwork.freedesktop.org/series/49923/ and https://patchwork.freedesktop.org/series/39958/

zxye:
This PR works with this i915 branch: https://cgit.freedesktop.org/~tursulin/drm-intel/log/?h=media

media_driver/agnostic/common/codec/hal/codechal_decode_scalability.cpp
```diff
@@ -1499,7 +1499,10 @@ MOS_STATUS CodecHalDecodeScalability_SetHintParams(
     else
     {
         VEParams.bScalableMode = false;
-        CODECHAL_DECODE_CHK_STATUS_RETURN(pVEInterface->pfnVESetHintParams(pVEInterface, &VEParams));
+        if (pVEInterface->pfnVESetHintParams)
```
dvrogozh: This looks to be a patch which can be merged in right now.

media_driver/agnostic/common/codec/hal/codechal_encode_scalability.cpp
```diff
@@ -121,6 +121,8 @@ MOS_STATUS CodecHalEncodeScalability_PopulateHintParams(
     CODECHAL_ENCODE_CHK_NULL_RETURN(cmdBuffer);
     PMOS_CMD_BUF_ATTRI_VE pAttriVe = (PMOS_CMD_BUF_ATTRI_VE)(cmdBuffer->Attributes.pAttriVe);
 
+    if (pAttriVe)
```
dvrogozh: Again, a patch we should consider merge in right now.

Also, you have 2 patches with the same name. That's bad. Either squash them or add something to separate titles, for example, add "decode" and "encode" prefix:
`encode: fix NULL pointer access issue for encode virtual engine enabling`

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4631,3 +4631,132 @@ mos_bufmgr_gem_init(int fd, int batch_size)
...
+    balancer.mask = ~0ull;
+
+    struct set_engines {
+        uint64_t extension;
+        struct class_instance engines[MAX_VDBOX_NUM];
+    } set_engines;
+    set_engines.extension = (uintptr_t)(&balancer);
+    memcpy(set_engines.engines, ci, count * sizeof(struct class_instance));
+    return mos_set_context_param(ctx,
+                          sizeof(set_engines.extension) + count*sizeof(struct class_instance),
```
dvrogozh: how count relates to MAX_VDBOX_NUM? Shouldn't we have buffer overflow protection here?

media_driver/linux/common/os/i915/include/mos_bufmgr.h
```diff
@@ -239,6 +239,21 @@ int mos_get_aperture_sizes(int fd, size_t *mappable, size_t *total);
 int mos_bufmgr_gem_get_devid(struct mos_bufmgr *bufmgr);
 
 struct mos_linux_context *mos_gem_context_create(struct mos_bufmgr *bufmgr);
+struct mos_linux_context *mos_gem_context_create_v2(struct mos_bufmgr *bufmgr, __u32 flags);
+struct class_instance {
+	uint16_t engine_class;
+	uint16_t instance;
+};
+#define MAX_VDBOX_NUM 64
```
dvrogozh: why 64? where this comes from? How many vdbox engines does i915 uAPI implies? or they get a count from us somehow for the buffers to fill?

media_driver/agnostic/common/codec/hal/codechal_decode_scalability.cpp
zxye: Yes, I will submit a PR to check-in this patch first.

media_driver/agnostic/common/codec/hal/codechal_encode_scalability.cpp
zxye: ok.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: count is returned from KMD which should always <= the array size MAX_VDBOX_NUM. But yet, should add protection here.

zxye: MAX_VDBOX_NUM should not be used here. Will revise the patch.

media_driver/linux/common/os/i915/include/mos_bufmgr.h
zxye: yes, i915 gets a buffer from UMD to fill when UMD queries engine instances. MAX_VDBOX_NUM is used to define the buffer size when we query the VDBOX engine instances.

media_driver/agnostic/common/codec/hal/codechal_decode_scalability.cpp
zxye: Done with PR#481 https://github.com/intel/media-driver/pull/481

media_driver/agnostic/common/codec/hal/codechal_encode_scalability.cpp
zxye: Done with PR   https://github.com/intel/media-driver/pull/481

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: done.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4691,3 +4691,246 @@ mos_bufmgr_gem_init(int fd, int batch_size)
...
+        return nullptr;
+
+    context = (struct mos_linux_context *)calloc(1, sizeof(*context));
+    if (!context)
+        return nullptr;
+
+    memclear(create);
+    create.flags = flags;
+    create.extensions = 0;
+    ret = drmIoctl(bufmgr_gem->fd, DRM_IOCTL_I915_GEM_CONTEXT_CREATE, &create);
```
tursulin: Mismatch in ioctl and data structure (_ext vs no _EXT). I think it doesn't matter though until  you attempt to use the extension pointer. For instance if you decided to move the below setting of VM into this single ctx create ioctl.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4691,3 +4691,246 @@ mos_bufmgr_gem_init(int fd, int batch_size)
...
+    struct drm_i915_query_engine_info *engines;
+    int ret, len;
+
+    engines = (drm_i915_query_engine_info *)malloc(4096);
+    memset(engines,0,4096);
+    memclear(query_item);
+    query_item.query_id = DRM_I915_QUERY_ENGINE_INFO;
+    query_item.length = 0;
+    query_item.data_ptr = (uintptr_t)engines;
+    memclear(query);
```
tursulin: Code looks correct so this is just an observation. Technically you don't need to set the data_ptr to anything valid when query_item.length is zero for the length query. Equally you could move the above malloc(4096) to malloc(len) after the first query.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4691,3 +4691,246 @@ mos_bufmgr_gem_init(int fd, int batch_size)
...
+    {
+        goto fini;
+    }
+
+    int i, num;
+    for (i = 0, num = 0; i < engines->num_engines; i++) {
+        struct drm_i915_engine_info *engine =
+            (struct drm_i915_engine_info *)&engines->engines[i];
+        if ( engine_class == engine->engine_class
+             && ((caps & engine->capabilities) || caps == 0))
```
tursulin: For future proofing, shouldn't the capability part of the check strictly speaking be "((caps & engine->capabilities) == caps)" so that it is match-all instead of match-any?

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4691,3 +4691,246 @@ mos_bufmgr_gem_init(int fd, int batch_size)
...
+        {
+            ci->engine_class = engine_class;
+            ci->engine_instance = engine->engine_instance;
+            ci++;
+            num++;
+        }
+    }
+    if (num > *nengine)
```
tursulin: This check would be better in the for loop above so it avoids overwriting memory before it decides to fail.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4691,3 +4691,246 @@ mos_bufmgr_gem_init(int fd, int batch_size)
...
+    uint32_t* balancer = (uint32_t*)malloc(size);
+    if (NULL == balancer)
+    {
+        return -ENOMEM;
+    }
+    memset(balancer, 0, size);
+    ((struct i915_context_engines_load_balance*) balancer)->base.name = I915_CONTEXT_ENGINES_EXT_LOAD_BALANCE;
+    ((struct i915_context_engines_load_balance*) balancer)->num_siblings = count;
+    memcpy(((struct i915_context_engines_load_balance*) balancer)->engines, ci, count * sizeof(*ci));
+
```
tursulin: Why is balancer a pointer to uint32_t? If it was a pointer to struct i915_context_engines_load_balance then you wouldn't need all these casts.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4691,3 +4691,246 @@ mos_bufmgr_gem_init(int fd, int batch_size)
...
+    /* set_engines->engines[0],
+       KMD requires the first entry of the engine map to be:
+       class=I915_ENGINE_CLASS_INVALID:instance=I915_ENGINE_CLASS_INVALID_NONE
+       */
+    ((struct i915_context_param_engines*)set_engines)->engines[0].engine_class = I915_ENGINE_CLASS_INVALID;
+    ((struct i915_context_param_engines*)set_engines)->engines[0].engine_instance = I915_ENGINE_CLASS_INVALID_NONE;
+
+    /* set_engines->engines[1]..[count], the "real" engine map */
+    memcpy(&((struct i915_context_param_engines*)set_engines)->engines[1], ci, count * sizeof(*ci));
+
```
tursulin: Same question of data type as with balancer.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4691,3 +4691,246 @@ mos_bufmgr_gem_init(int fd, int batch_size)
...
+    /* set_engines->engines[0],
+       KMD requires the first entry of the engine map to be:
+       class=I915_ENGINE_CLASS_INVALID:instance=I915_ENGINE_CLASS_INVALID_NONE
+       */
+    ((struct i915_context_param_engines*)set_engines)->engines[0].engine_class = I915_ENGINE_CLASS_INVALID;
+    ((struct i915_context_param_engines*)set_engines)->engines[0].engine_instance = I915_ENGINE_CLASS_INVALID_NONE;
+
+    /* set_engines->engines[1]..[count], the "real" engine map */
+    memcpy(&((struct i915_context_param_engines*)set_engines)->engines[1], ci, count * sizeof(*ci));
```
tursulin: Are you using map entries other than the VENG during submission? If not I'd recommend not configuring them and just have the VENG in the map.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4691,3 +4691,246 @@ mos_bufmgr_gem_init(int fd, int batch_size)
...
+    if (NULL == set_engines)
+    {
+        return -ENOMEM;
+    }
+
+    /* set_engines->extension */
+    ((struct i915_context_param_engines*) set_engines)->extensions = (uintptr_t)(balancer);
+
+    /* set_engines->engines[0],
+       KMD requires the first entry of the engine map to be:
```
tursulin: Strictly speaking it doesn't need to be the first entry. You can deserve any but first is perhaps most convenient if you want to use I915_EXEC_DEFAULT to mean VENG.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4691,3 +4691,246 @@ mos_bufmgr_gem_init(int fd, int batch_size)
...
+    struct i915_context_param_engines {
+    __u64 extensions;
+    struct i915_engine_class_instance engines[count];
+    }
+    */
+    size = sizeof(uint64_t) + (count + 1) * sizeof(*ci);
+    uint32_t* set_engines = (uint32_t*) malloc(size);
+    if (NULL == set_engines)
+    {
+        return -ENOMEM;
```
tursulin: Memory allocated the the balancer is leaked here.

media_driver/linux/common/os/mos_gpucontext_specific.cpp
```diff
@@ -112,6 +115,123 @@ MOS_STATUS GpuContextSpecific::Init(OsContext *osContext)
...
+            {
+                MOS_OS_ASSERTMESSAGE("Failed to set balancer extension.\n");
+                return MOS_STATUS_UNKNOWN;
+            }
+
+            if (createOption->SSEUValue != 0)
+            {
+                struct drm_i915_gem_context_param_sseu sseu;
+                MOS_ZeroMemory(&sseu, sizeof(sseu));
+                sseu.engine.engine_class = I915_ENGINE_CLASS_RENDER;
```
tursulin: Class is ignored in index addressing.

media_driver/linux/common/os/mos_gpucontext_specific.cpp
```diff
@@ -112,6 +115,123 @@ MOS_STATUS GpuContextSpecific::Init(OsContext *osContext)
...
+                sseu.engine.engine_instance = 0;
```
tursulin: Would it be more self-documenting to replace 0 with m_i915ExecFlag? I am thinking how to designate the VENG slot.

media_driver/linux/common/os/mos_gpucontext_specific.cpp
```diff
@@ -112,6 +115,123 @@ MOS_STATUS GpuContextSpecific::Init(OsContext *osContext)
...
+            if (m_createOptionEnhanced->UsingSFC)
+            {
+                caps |= I915_VIDEO_AND_ENHANCE_CLASS_CAPABILITY_SFC;
+            }
+
+            MOS_ZeroMemory(engine_map, sizeof(engine_map));
+            if (mos_query_engines(osInterface->pOsContext->fd,engine_class,caps,&nengine,engine_map))
+            {
+                MOS_OS_ASSERTMESSAGE("Failed to query engines.\n");
+                return MOS_STATUS_UNKNOWN;
```
tursulin: So the design is that certain SKUs (Icelake) will need new uAPI or nothing will work? Is fall back to non-VENG mode not feasible in this case?

media_driver/linux/common/os/mos_gpucontext_specific.cpp
```diff
@@ -112,6 +115,123 @@ MOS_STATUS GpuContextSpecific::Init(OsContext *osContext)
...
+            }
+
+            if (mos_set_context_param_load_balance(m_i915Context, engine_map, nengine))
+            {
+                MOS_OS_ASSERTMESSAGE("Failed to set balancer extension.\n");
+                return MOS_STATUS_UNKNOWN;
+            }
+        }
```
tursulin: Looks like the MOS_GPU_NODE_VIDEO and MOS_GPU_NODE_VE branches are completely identical apart from the engine class. So you could consolidate them into one branch and just toggle the class based on GpuNode.

media_driver/linux/common/os/mos_gpucontext_specific.cpp
```diff
@@ -909,3 +1049,44 @@ MOS_STATUS GpuContextSpecific::AllocateGPUStatusBuf()
     m_statusBufferResource = graphicsResource;
     return MOS_STATUS_SUCCESS;
 }
+
+#if 0
```
tursulin: I was confused about this code until I noticed the #if 0. :) You'll know if you need this in the pull request or not.

media_driver/linux/common/os/mos_os_specific.c
```diff
@@ -1288,13 +1293,27 @@ MOS_STATUS Linux_InitContext(
...
-       if (pContext->intel_context == nullptr)
-       {
-            MOS_OS_ASSERTMESSAGE("Failed to create drm intel context");
-            return MOS_STATUS_UNKNOWN;
-       }
+        if (pContext->intel_context == nullptr)
+        {
+             MOS_OS_ASSERTMESSAGE("Failed to create drm intel context");
+             return MOS_STATUS_UNKNOWN;
+        }
```
tursulin: This block looks like just whitespace noise.

media_driver/linux/ult/libdrm_mock/xf86drm_mock.c
```diff
@@ -357,6 +366,12 @@ mosdrmIoctl(int fd, unsigned long request, void *arg)
             ret = -1;
         }
         break;
+        case DRM_IOCTL_I915_GEM_VM_CREATE:
+        case DRM_IOCTL_I915_GEM_VM_DESTROY:
+        {
+            ret = 0;
```
tursulin: For full compatibility with i915 you might want to set vm_id to non-zero. (i915 will never provide zero vm_id.) I know it is not documented, but it may make sense regardless so there is some differentiation between unitialized struct and returned data.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: Thanks Tvrtko! This is a bug. Fixed.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: yes, fixed.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: fixed.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: fixed.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: fixed.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: yeah, fixed.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: fixed.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: updated the comments.

media_driver/linux/common/os/mos_gpucontext_specific.cpp
zxye: So for virtual engine, only .flags and .engine.engine_instance needed?

                sseu.flags = I915_CONTEXT_SSEU_FLAG_ENGINE_INDEX;
                sseu.engine.engine_instance = m_i915ExecFlag;

media_driver/linux/common/os/mos_gpucontext_specific.cpp
zxye: The code here is only executed when (osInterface->ctxBasedScheduling) is true. non-VENG mode will not run into here.

media_driver/linux/common/os/mos_gpucontext_specific.cpp
zxye: yes, fixed.

media_driver/linux/common/os/mos_gpucontext_specific.cpp
zxye: removed.

media_driver/linux/common/os/mos_os_specific.c
zxye: Yeah, original source indented with 3 spaces.  To make the patch clear, I will keep it unchanged in this commit.

media_driver/linux/ult/libdrm_mock/xf86drm_mock.c
zxye: fixed.

media_driver/linux/common/os/mos_gpucontext_specific.cpp
tursulin: Correct.

PR #293, opened 2018-08-29
Revert "[Decode] MPEG2 Decoder Fix slice_vertical_position handling i…

vshampor:
…n frames with fields"

This reverts commit 5900867e156d67e3e9d931ae6f6f161392dab8ed.

Reasoning:
1) The commit in question does not seem to fix anything, but in fact
breaks MPEG2 decode, as shown by the MediaSDK testing
2) The affected slice param structure never even leaves the driver
code
3) The commit message seems unrelated to what it actually does.

### DIFF
--- a/media_driver/linux/common/codec/ddi/media_ddi_decode_mpeg2.cpp
+++ b/media_driver/linux/common/codec/ddi/media_ddi_decode_mpeg2.cpp
@@ -103,7 +103,7 @@ VAStatus DdiDecodeMPEG2::ParseSliceParams(
     for (slcCount = 0; slcCount < numSlices; slcCount++)
     {
         codecSlcParams->m_sliceHorizontalPosition = slcParam->slice_horizontal_position;
-        codecSlcParams->m_sliceVerticalPosition   = slcParam->slice_vertical_position;
+        codecSlcParams->m_sliceVerticalPosition   = slcParam->slice_vertical_position / (1 + isField);
 
         codecSlcParams->m_sliceDataSize   = (slcParam->slice_data_size) * 8;
         codecSlcParams->m_sliceDataOffset = slcParam->slice_data_offset + sliceBaseOffset;


XinfengZhang:
@vshampor this fix is for MPEG2 interlace decode, 
Unfortunately ,There are no comments on libva.
Also no special description in MPEG2 spec. 
slice_vertical_position -- This is given by the last eight bits of the slice_start_code. It is an unsigned
integer giving the vertical position in macroblock units of the first macroblock in the slice.

So it depend on MB definition:
In frame pictures
• In the case of frame DCT coding, each block shall be composed of lines from the two fields
alternately. This is illustrated in Figure 6-13.
• In the case of field DCT coding, each block shall be composed of lines from only one of the
two fields. This is illustrated in Figure 6-14.
Field picture
In field pictures, each picture only contains lines from one of the fields. In this case each block
consists of lines taken from successive lines in the picture

so, for a 480P picture( 30 MB rows)
if frame picture:
frame coding: it should be 0-29
frame field coding: it should be 0-29
if field picture:
it should be 0-14

is it correct? If yes, lets follow mpeg2 spec.

vshampor:
Well, I am starting to understand the motivation here but I am still not convinced that this one-liner fix I move to revert actually removes the problems instead of adding to them. 

The fix would be valid if the `slcParam->slice_vertical_position` had contained the frame/field-dependent value (i.e 0-29 for frame coding and 0-14 for field picture in your example); it would also have been valid if the `slcParam->slice_vertical_position` had remained independent of frame/field coding (i.e. 0-29 in all cases), but all references to the auxiliary field `codecSlcParams->m_sliceVerticalPosition` in the surrounding code had been updated to accomodate for frame-field coding via an additional 1/(1 + isField) factor.

As it is now, the  `slcParam->slice_vertical_position` seems to contain frame/field independent data, since after @jcescamilla 's fix we have a CRC mismatch with the reference MPEG2 decoder. Therefore, this one-liner commit in question simply won't do: it should be extended with the logic to provide the frame/field dependent `slcParam->slice_vertical_position` values.. Until that is done, the current commit should be reverted.

vshampor:
Greetings!

It turned out that we had an infrastructure problem in the previous MediaSDK validation and an older version of the library was used instead of the proper one. The most recent, correct validation does not show any problems with the MPEG2 decode. Furthermore, I found the MediaSDK library commit (https://github.com/Intel-Media-SDK/MediaSDK/commit/321f59e96dcc1165e05a701e3aff295582de99c8) that aligns the slice parsing to the behaviour introduced by @jcescamilla 's commit, so no action is required and I will close this pull request.

However, I still do not understand why the formula for the `codecSlcParams->m_sliceVerticalPosition` value had been changed, but other [references](https://github.com/intel/media-driver/blob/5900867e156d67e3e9d931ae6f6f161392dab8ed/media_driver/linux/common/codec/ddi/media_ddi_decode_mpeg2.cpp#L59-L60) to the same field had not been updated.

Issue #272, opened 2018-08-13
The vpp will ignore alpha value in output_background_color and always clear background to black.


xuguangxin:
According to libva definition, the output_background_color  is an ARGB value. If we set alpha to 0.  It will use the original background value. This feature is very useful if we want to draw mutli video on different areas. But the current driver will ignore the alpha and always clear the background to black. 

How to reproduce:
1. follow https://github.com/intel/libyami/wiki/Build to build libyami and the media-driver
2. configure libyami-utils using --enable-dmabuf
3. kill x server
4. run libyami-utils/exmpale/grid xxx.264 -r 4 -c 5

You will see only right conner has a video playback. other areas are black

Expected behaviors:
We can see 4x5(20) videos playback on screen

FurongZhang:
Thanks for the details of issue. Will try to reproduce the issue first.

xuguangxin:
hi @FurongZhang , any update on this? thanks

Jister:
Hi, Guangxin, I have create a pull request to fix the issue, please refer to #296

PR #296, opened 2018-08-30
[VP]Fix issue #272.

Jister:
Add output region of target surface

### DIFF
--- a/media_driver/linux/common/vp/ddi/media_libva_vp.c
+++ b/media_driver/linux/common/vp/ddi/media_libva_vp.c
@@ -861,6 +861,14 @@ DdiVp_SetProcPipelineParams(
         pVpHalSrcSurf->rcDst.left   = pPipelineParam->output_region->x;
         pVpHalSrcSurf->rcDst.right  = pPipelineParam->output_region->x + pPipelineParam->output_region->width;
         pVpHalSrcSurf->rcDst.bottom = pPipelineParam->output_region->y + pPipelineParam->output_region->height;
+        // According to libva  definition, if alpha in output background color is zero, then colorfill is not needed
+        if ((pPipelineParam->output_background_color >> 24) == 0)
+        {
+            pVpHalTgtSurf->rcDst.top    = pPipelineParam->output_region->y;
+            pVpHalTgtSurf->rcDst.left   = pPipelineParam->output_region->x;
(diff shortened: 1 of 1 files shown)


xuguangxin:
@Jister , great. this fixed my issue. 

thanks

xuguangxin:
@Jister , one thing I want double confirm. if we really follow alpha blending define. Why we need specail handle alpha value 0? Is this mean the alpha value does not work ?

Jister:
@xuguangxin Actually this issue is not an alpha blending case, multi-input surfaces are written to different region of the same output surface after scaling, if color fill is executed, the surface data outside of the region will sure be overwritten by the data of background color, so here the fix handle alpha value 0 is to judge whether color fill need to be executed. For alpha blending usage, you need to call blending API to do it.

Jister:
@xuguangxin By the way, for color fill cases, currently our driver will always ignore alpha channel and fill the alpha with 0xFF, this is another story, and I have fixed it and will check in soon.

xuguangxin:
No, the definition for the background is not color fill, it's alpha blending. https://github.com/intel/libva/blob/master/va/va_vpp.h#L871.
alpha value 0 and 0xff is just a special case for the alpha blending.

thanks

Jister:
As libva says "Background color used to **fill in pixels** that reside outside of the specified \ref output_region.", that means colorfill and colorfill will always use this value. For alpha blending case, when primary layer need to do color fill (i.e. primary layer don't represent the whole surface), we will also use this value as background color and blend the primary and background together, but when it's 0, the background will be transparent, so we don't need to fill it. 
For sublayer, we'll use the alpha value in VABlendState to do blending. You can refer to:
https://github.com/intel/libva/blob/30c751f7b74c43021f33361c1baf2aa92acd0317/va/va_vpp.h#L482

Thanks

xuguangxin:
@Jister 
In this case, the alpha value has no much mean.
It only can be 0x0 or 0xff.
@xhaihao  and @XinfengZhang, could you explain more about the API.
If we can only accept 0x0 and 0xff for the alpha. we'd better document it down in headers.

thanks

Jister:
The change affect blending cases, will backout and create another fix.

xuguangxin:
Please consider use Alpha blending for this API. else we need ask @xhaihao and @XinfengZhang change the libva API description.

Jister:
Created another pull request #324

PR #311, opened 2018-09-11
Fix for VmRSS memory growth on VPP Close/Init.

Stephanie-Qi:
Signed-off-by: Qi, Stephanie <stephanie.qi@intel.com>

### DIFF
--- a/media_driver/agnostic/common/os/mos_gpucontextmgr.cpp
+++ b/media_driver/agnostic/common/os/mos_gpucontextmgr.cpp
@@ -132,10 +132,31 @@ GpuContext *GpuContextMgr::CreateGpuContext(
 
     MOS_LockMutex(m_gpuContextArrayMutex);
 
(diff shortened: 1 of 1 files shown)


media_driver/agnostic/common/os/mos_gpucontextmgr.cpp
```diff
@@ -132,10 +132,31 @@ GpuContext *GpuContextMgr::CreateGpuContext(
...
+    {
+        if(curGpuContext == nullptr)
+        {
+            gpuContextHandle = index;
+            break;
+        }
+        index++;
+    }
+
+    gpuContextHandle = m_gpuContextArray.size() ? index : 0;
```
LuoIvan: Why we need reassign to gpuContextHandle here?

Stephanie-Qi: I have removed reassign.

PR #319, opened 2018-09-18
Enable QVBR for AVC DP encoding

ChenXiaomin36:
Change-Id: I1a22aa036bf9438c491e8749087c8aa12217dbb0
Signed-off-by: xiaominc <xiaomin.chen@intel.com>

### DIFF
--- a/media_driver/linux/common/codec/ddi/media_ddi_encode_avc.cpp
+++ b/media_driver/linux/common/codec/ddi/media_ddi_encode_avc.cpp
@@ -265,6 +265,12 @@ VAStatus DdiEncodeAvc::ParseMiscParamRC(void *data)
         seqParams->MinBitRate    = (uint32_t)((uint64_t)seqParams->TargetBitRate * (2 * encMiscParamRC->target_percentage - 100) / 100);
         seqParams->TargetBitRate = (uint32_t)((uint64_t)seqParams->TargetBitRate * encMiscParamRC->target_percentage / 100);
         vuiParam->cbr_flag       = 0x0;
+
+        if (VA_RC_QVBR == m_encodeCtx->uiRCMethod)
+        {
+            seqParams->ICQQualityFactor = encMiscParamRC->quality_factor;
+        }
+
         if ((m_encodeCtx->uiTargetBitRate != seqParams->TargetBitRate) ||
             (m_encodeCtx->uiMaxBitRate != seqParams->MaxBitRate))
         {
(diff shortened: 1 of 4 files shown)


ChenXiaomin36:
Here are the answers:
1. Now after this patch, only AVC encoder will support QVBR.
2. Once this patch for AVC DP is merged, I will submit another PR for AVC lowpower encoder.
Thanks! @dvrogozh

dvrogozh:
@ChenXiaomin36 : I see some strange behaviour on enabled qvbr. Could you, please, check on your side? Consider the following mediasdk command lines with _different_ bitrate settings (compare: -b 2500 and -b 4000, -MaxKbps 3000 and MaxKbps 4800):
```
./share/mfx/samples/sample_encode h264 -i 720p.yuv -o 30_se.h264 -w 1280 -h 720 -f 30 -b 2500 -u 1 -hw -vaapi -n 600 -r 1 -g 30 -MaxKbps 3000 -qvbr 30

./share/mfx/samples/sample_encode h264 -i 720p.yuv -o 48_se.h264 -w 1280 -h 720 -f 30 -b 4000 -u 1 -hw -vaapi -n 600 -r 1 -g 30 -MaxKbps 4800 -qvbr 30
```

The strange thing is that if you will check the quality of output streams, you will see that PSNR values are identical. Are there any limit in the driver which forces bitrates to be equal to this limit if user-specified bitrate is lower than this limit? I am looking into such a limit in mediasdk, but it seems this can be something on the driver side, because settings which mediasdk pass to VAEncMiscParameterRateControl look correct (my custom debug prints below added here: https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/shared/src/mfx_h264_encode_vaapi.cpp#L210):
```
>>> rate_param->bits_per_second=2999296, rate_param->window_size=300000, rate_param->target_percentage=83, rate_param->quality_factor=30

>>> rate_param->bits_per_second=4799488, rate_param->window_size=480000, rate_param->target_percentage=83, rate_param->quality_factor=30
```

Note: you need this PR Intel-Media-SDK/MediaSDK#656 to try QVBR under mediasdk.

dvrogozh:
@ChenXiaomin36 : any feedback on the noted behavior?

ChenXiaomin36:
@dvrogozh : I have tried the command lines you mentioned, compared different bitrate settings. I also got the result that PSNR values were the same for these two output bitstreams. I need to look into the driver codes,  perhaps even lower level realization for QVBR. I will give out the feedback once I have some findings. Thanks!

dvrogozh:
@ChenXiaomin36 : I have submitted #379 to track issue with the same quality for the different QVBR settings.

PR #327, opened 2018-09-26
Add ubuntu 18.04 dockerfile

Sherry-Lin:
Includes libva/libva-utils/gmm/media-driver/MediaSDK


uartie:
From https://docs.docker.com/engine/reference/commandline/build/#text-files
> In most cases, it’s best to put each Dockerfile in an empty directory. Then, add to that directory only the files needed for building the Dockerfile. To increase the build’s performance, you can exclude files and directories by adding a .dockerignore file to that directory as well. For information on creating one, see the .dockerignore file.

The Dockerfile here is at the top-level of the project directory.  Thus, the entire source tree (including .git directory) will be included in the docker build context.  IIRC, when docker builds a Dockerfile, the build context is copied to Docker's graph directory and builds from there.  Hence, the entire source tree will get copied and slow down the overall docker build performance.  At the very least, you should include a .dockerignore file (https://docs.docker.com/engine/reference/builder/#dockerignore-file).  Otherwise, create a separate directory with Dockerfile and only the files needed by it.

Dockerfile
```diff
@@ -0,0 +1,91 @@
...
+    ln -sf /usr/bin/ccache c++ && \
+    ln -sf /usr/bin/ccache clang && \
+    ln -sf /usr/bin/ccache clang++ && \
+    ln -sf /usr/bin/ccache clang-4.0 && \
+    ln -sf /usr/bin/ccache clang++-4.0
+
+RUN cd /opt/src && \
+    git clone https://github.com/intel/libva && \
+    cd libva && \
+    git checkout 2.2.1.pre1-20180921 && \
```
dvrogozh: What's the purpose of this docker file? If that's to prepare _latest_ media environment, then why it is full of specific tags? If that's to reflect certain mediasdk stack release, then: 1) why that's in master branch, 2) why mediasdk tag is not used?

Sherry-Lin: Besides of quarterly release, we have other pre-release in weekly or bi-weekly cadence which is qualified by CI suite. This dockerfile is designed to create an automated build in hub.docker.com based on these pre-release tag, then user could docker pull the image directly.

Sherry-Lin: For mediasdk, I'm not sure which version is targeting to the build we verified 7.0 .0484-OS-OS1.02. Could you please provide the tag or commit ID?

Sherry-Lin:
Thanks for your suggestions. I moved it to Tools/dockers/ubuntu.

Sherry-Lin:
@dvrogozh could you please check the patch to see any issues? If no, could you please approve it? Thanks.

PR #357, opened 2018-10-11
add missing break in DdiMedia_CreateImage

### LINKED ISSUE #321: [gst-vaapi][ICL] vaCreateImage can not used uyvy format to create image.
By using uyvy format in vaCreateImage, un-implemented status will be return. This format required by Iotg, may need your help to support this format.

XinfengZhang:


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -3930,6 +3930,7 @@ VAStatus DdiMedia_CreateImage(
         case VA_FOURCC_UYVY:
```
wangzj0601: duplicated case value

PR #370, opened 2018-10-19
android: replace deprecated cutils/log.h with log/log.h

xuguangxin:
cutils/log.h deprecated by google on android P


XinfengZhang:
LGTM

PR #394, opened 2018-11-01
[CM] Add an option in CMakeLists to include/exclude cmrtlib build

HeFan2017:
Add an option in CMakeLists to include/exclude cmrtlib build together with media driver

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,9 +40,13 @@ if("${os_name}" STREQUAL "clear-linux-os")
     set(CMAKE_INSTALL_SYSCONFDIR "usr/share/defaults/etc")
 endif()
 
+option (BUILD_CMRTLIB "Build and Install cmrtlib together with media driver" ON)
+
 include(GNUInstallDirs)
 
-add_subdirectory(cmrtlib)
+if (BUILD_CMRTLIB)
+    add_subdirectory(cmrtlib)
+endif()
 
 add_subdirectory(media_driver)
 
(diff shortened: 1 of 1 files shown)


dvrogozh:
@chivakker : FYI, please, review.

CMakeLists.txt
```diff
@@ -40,9 +40,13 @@ if("${os_name}" STREQUAL "clear-linux-os")
     set(CMAKE_INSTALL_SYSCONFDIR "usr/share/defaults/etc")
 endif()
 
+option (BUILD_CMRTLIB "Build and Install cmrtlib together with media driver" ON)
```
dvrogozh: @chivakker: this commit looks ok to me. How about you?

cmrtlib/linux/CMakeLists.txt
```diff
@@ -79,7 +79,8 @@ set(CMRT_DEFINES "")
 cmrt_include_directory(${CMAKE_CURRENT_LIST_DIR}/../agnostic)
 cmrt_include_directory(${CMAKE_CURRENT_LIST_DIR}/../linux)
 
-setup_library(igfxcmrt "${CMRT_SOURCES}" TRUE "igfxcmrt${TARGET_MODIFIER}")
+#setup_library(igfxcmrt "${CMRT_SOURCES}" TRUE "igfxcmrt${TARGET_MODIFIER}")
```
dvrogozh: This commit is already in a separate PR #413. Could you, please, drop the commit from the current PR?

HeFan2017: done, thanks

XinfengZhang:
@chivakker if you have no concern, we will check in it.

CMakeLists.txt
chivakker: Can default be OFF?  Users should intentionally know they are enabling it.  Otherwise it looks okay.

HeFan2017: cmrtlib and media_driver are both in this repo, so we still prefer this be ON by default.

PR #398, opened 2018-11-02
VeboxSendVeboxCmd: do not goto finish if caller pass NULl for pRender

xuguangxin:
goto finish will use CmdBuffer, it will init later by VeboxSendVeboxCmd_Prepare


xuguangxin:
@XinfengZhang , any feedback on this?
thanks

PR #399, opened 2018-11-02
CodechalMmcDecodeMpeg2: follow other codec, do not check hwInterface for NULL

xuguangxin:
No need check the hwInterface  here
if hwInterface  is NULL, CodecHalMmcState::m_hwInterface will be null.
Many fuction calls will crash for release version(CODECHAL_HW_ASSERT is no op in release build)


xuguangxin:
@XinfengZhang , any feedback on this?

XinfengZhang:
LGTM

XinfengZhang:
will merge

PR #407, opened 2018-11-07
Fix transcoding performance issue on BDW.

TangZhiZhen:
Signed-off-by: ZhiZhen Tang <zhizhen.tang@intel.com>


TangZhiZhen:
This patch is porting from PV5 HF1.

TangZhiZhen:
yes, the perf drop only on BDW

cchunbo:
> yes, the perf drop only on BDW
but the change will impact all platforms

TangZhiZhen:
yes, we merged it on PV5 HF1 and run full test, it isn't impact on SKL platform.

TangZhiZhen:
This issue was caused by CL722640. CL722640 can improve HEVC VME performance, but it is applied for all codecs. I changed this patch only for HEVC codec.

PR #413, opened 2018-11-12
Add version information to the cmrt library

HeFan2017:


HeFan2017:
@chivakker

cmrtlib/linux/CMakeLists.txt
```diff
@@ -90,8 +91,26 @@ endif()
...
+list(GET CMRT_VERSION_LIST 0 CMRT_VERSION_MAJOR)
+
+set_target_properties(igfxcmrt PROPERTIES VERSION ${CMRT_VERSION})
+set_target_properties(igfxcmrt PROPERTIES SOVERSION ${CMRT_VERSION_MAJOR})
 set_target_properties( igfxcmrt PROPERTIES PREFIX "")
 target_link_libraries( igfxcmrt dl va rt ${GCC_SECURE_LINK_FLAGS})
 
+ADD_CUSTOM_TARGET(link_target ALL COMMAND ${CMAKE_COMMAND} -E create_symlink libigfxcmrt${TARGET_MODIFIER}.so igfxcmrt${TARGET_MODIFIER}.so)
+
+if(DEFINED CMAKE_INSTALL_LIBDIR AND (NOT CMAKE_INSTALL_LIBDIR STREQUAL ""))
```
dvrogozh: These checks are absolutely useless and should be dropped. CMAKE_INSTALL_LIBDIR _will_ be defined once we include GNUInstallDirs which we do unconditionally here:
https://github.com/intel/media-driver/blob/f9b1c19c36ead3cca95295581f0fee8b8b99a137/CMakeLists.txt#L43
If user did not specify CMAKE_INSTALL_LIBDIR, then system default will be used. So, drop the condition, please.

cmrtlib/linux/CMakeLists.txt
```diff
@@ -90,8 +91,26 @@ endif()
...
+    install(TARGETS igfxcmrt LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT media NAMELINK_SKIP)
+    install(TARGETS igfxcmrt LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT media NAMELINK_ONLY)
```
dvrogozh: To be honest I do not favor self-packaging support in the repositories. I believe this is responsibility of the Linux distributions maintainers. I actually would like to see all the packaging code to be completely removed going forward. For these install commands this would mean remove code starting from "COMPONENT" keyword.

However, since packaging is currently here, you have wrong assumption that all targets should go to the same package. No, usually 2 packages are created:
1. runtime package which should hold libigfxcmrt.so.x.y.z and libigfxcmrt.so.x -> libigfxcmrt.so.x.y.z
2. development package which should hold libigfxcmrt.so and *.h and *.pc files

Please, decide how you will proceed with this....

cmrtlib/linux/CMakeLists.txt
```diff
@@ -79,7 +79,8 @@ set(CMRT_DEFINES "")
 cmrt_include_directory(${CMAKE_CURRENT_LIST_DIR}/../agnostic)
 cmrt_include_directory(${CMAKE_CURRENT_LIST_DIR}/../linux)
 
-setup_library(igfxcmrt "${CMRT_SOURCES}" TRUE "igfxcmrt${TARGET_MODIFIER}")
+#setup_library(igfxcmrt "${CMRT_SOURCES}" TRUE "igfxcmrt${TARGET_MODIFIER}")
```
dvrogozh: don't comment - delete if you don't need this.

cmrtlib/linux/CMakeLists.txt
```diff
@@ -90,8 +91,26 @@ endif()
 set_target_properties( igfxcmrt PROPERTIES 
   FOLDER CM_RT 
   COMPILE_DEFINITIONS "${CMRT_DEFINES}" 
-  OUTPUT_NAME "igfxcmrt${TARGET_MODIFIER}"
+  OUTPUT_NAME "libigfxcmrt${TARGET_MODIFIER}"
```
dvrogozh: "lib" looks strange. I think you need to remove it from here + remove set(PREFIX "") below. Linux will do the trick for you.

cmrtlib/linux/CMakeLists.txt
```diff
@@ -90,8 +91,26 @@ endif()
...
+if(NOT DEFINED MEDIA_VERSION OR MEDIA_VERSION STREQUAL "")
+    set(MEDIA_VERSION "1.0.0")
+endif()
+set(CMRT_VERSION ${MEDIA_VERSION})
+string(REPLACE "." ";" CMRT_VERSION_LIST ${CMRT_VERSION})
+list(GET CMRT_VERSION_LIST 0 CMRT_VERSION_MAJOR)
+
+set_target_properties(igfxcmrt PROPERTIES VERSION ${CMRT_VERSION})
+set_target_properties(igfxcmrt PROPERTIES SOVERSION ${CMRT_VERSION_MAJOR})
 set_target_properties( igfxcmrt PROPERTIES PREFIX "")
```
dvrogozh: remove

cmrtlib/linux/CMakeLists.txt
```diff
@@ -90,8 +91,26 @@ endif()
...
 target_link_libraries( igfxcmrt dl va rt ${GCC_SECURE_LINK_FLAGS})
 
+ADD_CUSTOM_TARGET(link_target ALL COMMAND ${CMAKE_COMMAND} -E create_symlink libigfxcmrt${TARGET_MODIFIER}.so igfxcmrt${TARGET_MODIFIER}.so)
```
dvrogozh: This is useless command. This should be automatically handled by cmake with install() commands below (the one with NAMELINK_ONLY). For that user needs to execute "make install".

cmrtlib/linux/CMakeLists.txt
```diff
@@ -90,8 +91,26 @@ endif()
 set_target_properties( igfxcmrt PROPERTIES 
   FOLDER CM_RT 
   COMPILE_DEFINITIONS "${CMRT_DEFINES}" 
-  OUTPUT_NAME "igfxcmrt${TARGET_MODIFIER}"
+  OUTPUT_NAME "libigfxcmrt${TARGET_MODIFIER}"
   )
+
+# set the version, aligned with media driver
+if(NOT DEFINED MEDIA_VERSION OR MEDIA_VERSION STREQUAL "")
```
dvrogozh: This is already done here:
https://github.com/intel/media-driver/blob/f9b1c19c36ead3cca95295581f0fee8b8b99a137/CMakeLists.txt#L33
Remove.

cmrtlib/linux/CMakeLists.txt
HeFan2017: This CMakeFile can be used directly sometime, like by the build_linux.sh. This is what we need to generate the CM SDK package.

cmrtlib/linux/CMakeLists.txt
HeFan2017: This CMakeFile can be used directly sometime, like by the build_linux.sh. This is what we need to generate the CM SDK package.

cmrtlib/linux/CMakeLists.txt
dvrogozh: No, you don't. You don't actually have backward compatibility point in any of Linux distributions and don't pretend that you have. All what was before was custom patches over existing linux distributions, they don't count. You have nice opportunity to simplify the things and make them right. I strongly suggest to use it.

HeFan2017: For those applications who already depends on igfxcmrt64.so, obviously this will block them. They will find they can't find hte igfxcmrt64.so when they update their media driver

cmrtlib/linux/CMakeLists.txt
chivakker: yes, delete it.

cmrtlib/linux/CMakeLists.txt
chivakker: What is build_linux.sh? I cannot find it in the current repository.  Please avoid wrapping build targets in scripts.

cmrtlib/linux/CMakeLists.txt
chivakker: you let other applications know that they need to adjust by bumping the ABI/API version.  This is exactly what should be fixed for good.

cmrtlib/linux/CMakeLists.txt
chivakker: Avoid the "backwards" compatibility that you're trying to artificially create.  It will be a nightmare for distributions to know which ABI version they should use.

cmrtlib/linux/CMakeLists.txt
HeFan2017: build_linux.sh is in cmrtlib folder. But you are right, we can remove that script.
Back to this topic, With this checking, we can also make the cmrtlib alone without media_driver.

HeFan2017:
@chivakker @dvrogozh I uploaded one new version here, please help review. Thanks in advance

cmrtlib/linux/CMakeLists.txt
```diff
@@ -90,8 +77,25 @@ endif()
...
   )
-set_target_properties( igfxcmrt PROPERTIES PREFIX "")
+
+# set the version, aligned with media driver
+if(NOT DEFINED MEDIA_VERSION OR MEDIA_VERSION STREQUAL "")
+    set(MEDIA_VERSION "1.0.0")
+endif()
+set(CMRT_VERSION ${MEDIA_VERSION})
+string(REPLACE "." ";" CMRT_VERSION_LIST ${CMRT_VERSION})
+list(GET CMRT_VERSION_LIST 0 CMRT_VERSION_MAJOR)
```
dvrogozh: Well, this will not do. You can't actually delegate setting soversion on Linux outside of the build scripts. The critical point here is major version: thru it you communicate backward _incompatible_ changes. This means that you can't let it go wrong. Most distributions will not be aware that they need to setup it at all. If someone will change it - we will never guess what they will chose. Thus, the correct approach would be to simply hardcode versioning here like:
```
set(CMRT_VERSION_MAJOR 1)
set(CMRT_VERSION_MINOR 0)
set(CMRT_VERSION_PATCH 0)
set(CMRT_VERSION ${CMRT_VERSION_MAJOR}.${CMRT_VERSION_MINOR}.${CMRT_VERSION_PATCH}.)
```
And make sure that you
* Bump MAJOR version up on each backward incompatible uAPI change
* Bump MINOR version up on each backward compatible uAPI change
* Bump up PATCH version on each release bug fix.

cmrtlib/linux/CMakeLists.txt
```diff
@@ -90,8 +77,25 @@ endif()
 set_target_properties( igfxcmrt PROPERTIES 
   FOLDER CM_RT 
   COMPILE_DEFINITIONS "${CMRT_DEFINES}" 
-  OUTPUT_NAME "igfxcmrt${TARGET_MODIFIER}"
+  OUTPUT_NAME "igfxcmrt"
```
dvrogozh: isn't it the default? you can try to drop OUTPUT_NAME setting completely - it should work I think

cmrtlib/linux/CMakeLists.txt
HeFan2017: I did fix it in my last update. Can you review it in the latest file changes?

cmrtlib/linux/CMakeLists.txt
HeFan2017: I did fix it in my last update. Can you review it in the latest file changes?

cmrtlib/linux/CMakeLists.txt
HeFan2017: yes, i have dropped the old igfxcmrt64.so in last update. please review the latest files change

cmrtlib/linux/CMakeLists.txt
```diff
@@ -90,8 +77,22 @@ endif()
 set_target_properties( igfxcmrt PROPERTIES 
   FOLDER CM_RT 
   COMPILE_DEFINITIONS "${CMRT_DEFINES}" 
-  OUTPUT_NAME "igfxcmrt${TARGET_MODIFIER}"
   )
-set_target_properties( igfxcmrt PROPERTIES PREFIX "")
+
+# set the version
+set(CMRT_VERSION_MAJOR 1)
```
dvrogozh: You probably better use CM runtime version which you already have in cm_def.h for MAJOR and MINOR:
`cmrtlib/agnostic/share/cm_def.h:41:#define CURRENT_CM_VERSION  (CM_7_2)`
You can get it automatically parsing cm_def.h on the fly. Look here for the reference: https://github.com/Intel-Media-SDK/MediaSDK/blob/master/builder/FindFunctions.cmake#L83

This will actually simplify things a lot since customers will be able to deduct soversion from the API version and use dlopen() if needed.

HeFan2017: yes, it makes sense. Made the change in latest update. Please help review it. Thanks

dvrogozh:
You need to rebase, there is a conflict after recent merge of #394. By the way, do you realize that you have created merge commit for #394? Look into git history with `git log --graph`. Next time you would probably want to avoid that and use "rebase and merge" option on GitHub.

As for the rest, everything looks fine. And works as expected on my side.

cmrtlib/linux/CMakeLists.txt
```diff
@@ -90,8 +77,20 @@ endif()
...
+set(CMRT_VERSION ${CMRT_VERSION_MAJOR}.${CMRT_VERSION_MINOR}.${CMRT_VERSION_PATCH})
+
+set_target_properties(igfxcmrt PROPERTIES VERSION ${CMRT_VERSION})
+set_target_properties(igfxcmrt PROPERTIES SOVERSION ${CMRT_VERSION_MAJOR})
 target_link_libraries( igfxcmrt dl va rt ${GCC_SECURE_LINK_FLAGS})
 
+if(NOT DEFINED CMAKE_INSTALL_LIBDIR)
+    include(GNUInstallDirs)
+endif()
+
```
chivakker: `include(GNUinstallDirs)` should be used without conditional on the existence of CMAKE_INSTALL_LIBDIR

HeFan2017: updated

dvrogozh:
@chivakker : are you ok with this change now?

PR #414, opened 2018-11-12
bxt fix

ilapichev:
Hi guys. 
We had few issues regarding to MFX_ERR_NONE != MFX_ERR_GPU_HANG with corrupted streams.
Family: BXT
Platform: u18.04-64
This patch solves this problem.

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -31,6 +31,8 @@
 #include <sys/mman.h>
 #include <sys/ioctl.h>
 #include <unistd.h>
+#include <thread>
+#include <chrono>
 
 #ifndef ANDROID
 #include <X11/Xutil.h>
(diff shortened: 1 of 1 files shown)


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -3339,6 +3341,10 @@ static VAStatus DdiMedia_SyncSurface (
         // Just loop while gem_bo_wait times-out.
     }
 
+    PDDI_MEDIA_CONTEXT mediaDrvCtx = DdiMedia_GetMediaContext(ctx);
+    if(mediaDrvCtx->platform.eRenderCoreFamily <= IGFX_GEN9_CORE)
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
dvrogozh: This introduces 1ms delay for BDW, SKL and other Gen9 platforms. This is a _huge_ delay which will signinicantly affect performance. And looking into PR message you are trying to fix GPU hangs - in this change you just workaround issues, not fixing them.

This change is not something to be done. It has -2 from me.

ilapichev: Yes, this is workaround, not a fix. And this is really draft staff. After a conversation with A Larionov we decided to to create a pull request just to point a problem in a driver.

This workaround fixes MDP-51470, MDP-48211, MDP-51475 and MDP-50855, related to GPU hangs.

>> This change is not something to be done.
For sure, I agree with this.

PR #418, opened 2018-11-13
replace all uapi header files with upstream 

XinfengZhang:


media_driver/linux/common/os/i915/include/uapi/README
```diff
@@ -0,0 +1,15 @@
+This directory contains some copy of the installed kernel headers
```
dvrogozh: "contains **a** copy" - drop some

media_driver/linux/common/os/i915/include/uapi/README
```diff
@@ -0,0 +1,15 @@
...
+required by the iHD drivers to communicate with the kernel.
```
dvrogozh: "driver" - drop -s

media_driver/linux/common/os/i915/include/uapi/README
```diff
@@ -0,0 +1,15 @@
...
+Whenever either of those driver needs new definitions for new kernel
```
dvrogozh: "Whenever driver needs..." - drop "either of those"

media_driver/linux/common/os/i915/include/uapi/README
```diff
@@ -0,0 +1,15 @@
...
+APIs, these files should be updated.
+
+These files in master should only be updated once the changes have landed
+in the drm-tip repo https://anongit.freedesktop.org/git/drm-tip.git.
```
dvrogozh: Pay attention that https://github.com/mesa3d/mesa/blob/master/include/drm-uapi/README uses drm-next for this purpose. That's for the reason! drm-tip is a bad choice because it is subject to force updates. We will have difficulties to track back commitid from drm-tip - it may be missed because of the force updates... So, we really should sync against drm-next which is subject to no force updates. Once something landed in drm-next, this will land in the mainline Linux kernel at appropriate merge window.

Thus, please, change to use drm-next. And you may need to update header files a little bit since drm-tip can be slightly ahead.

dvrogozh:
Also, Github detect some merge conflict in media_driver/linux/common/os/libdrm/media_srcs.cmake. Please, resolve.

dvrogozh:
@XinfengZhang : any update?

media_driver/linux/common/cm/cm_hal_os.cpp
```diff
@@ -26,7 +26,7 @@
 #include "mos_os.h"
 #include "cm_hal.h"
 #include "cm_def_os.h"
-#include "i915_header.h"
+#include "i915_drm.h"
```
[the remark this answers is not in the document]
XinfengZhang: not test, but  build success

media_driver/linux/common/os/i915/include/uapi/README
XinfengZhang: thanks

media_driver/linux/common/cm/cm_hal_os.cpp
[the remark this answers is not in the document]
dvrogozh: Ok, that's enough. I actually asked because I oversaw rename of i915_header.h.

PR #421, opened 2018-11-19
Add readme in the cmrtlib folder

HeFan2017:


cmrtlib/README.md
```diff
@@ -0,0 +1,29 @@
+# Intel(R) C for Media Runtime
+
+
+## Introduction
+
+cmrtlib is a runtime library needed when user wants to execute their own GPU kernels on render engine. It calls iHD media driver to load the kernels and allocate the resources. It provides a set of APIs defined in agnostic/share/cm_rt.h for user to call directly from application.
+
+## License
+
+The Intel(R) Media Driver for VAAPI is distributed under the MIT license.
```
StonyShen: Need change this sentence accordingly.

cmrtlib/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+cmrtlib maintains the backward compatibility in both API level and ABI level
+
+## SDK package
+
+You can build your application with the SDK package we maintained in https://01.org/c-for-media-development-package/. In the SDK, we provided the CMake files to build your own application and kernels. You can also find the spec of APIs in the documents of the SDK package.
+
+## Notes
+
+This cmrtlib is a different project compared with https://github.com/intel/cmrt, although they have similar names and funtionalities. DO NOT replace one by the one.
```
dvrogozh: DO NOT replace one by **>>another<<**.

cmrtlib/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+cmrtlib is recommended to be built and installed together with media driver. Please refer to the readme in the root path.
+
+## Backward compatibility
+
+cmrtlib maintains the backward compatibility in both API level and ABI level
+
+## SDK package
+
+You can build your application with the SDK package we maintained in https://01.org/c-for-media-development-package/. In the SDK, we provided the CMake files to build your own application and kernels. You can also find the spec of APIs in the documents of the SDK package.
```
dvrogozh: This is not what we expect to have! Just above you said that header files are here, media-driver repository. Thus, the primary thing you should explain is: how to use these header files. That's the reason why you need to have cmrt.pc and *.h files installed within media-driver installation. Once done, you need to add this information in this readme. As for the "SDK package" section, please, make sure to formulate it in a way similar to: "**Additionally** you can consider usage of CM SDK package which is described here https://01.org/c-for-media-development-package. It will permit you to do this, this and this (probably compile EU kernels, review additional documentation on how to use C for media, etc.)."

cmrtlib/README.md
chivakker: This cmrtlib library is a separate effort from a similar library (https://github.com/intel/cmrt).  They may provide the same functionalities but this is not intended to replace the other. This cmrtlib works and is maintained to support the following platforms: (a link to driver's platforms maybe).

cmrtlib/README.md
chivakker: I'd prefer to make a short reference:

SDK package can be used with this cmrtlib.  Please refer to SDK package for instructions.

And if SDK package doesn't have instructions to use cmrtlib, file a bug there.

chivakker:
The sections require better wording, please be precise on how/what this library should be used for.  There's no need to be extra verbose.

PR #423, opened 2018-11-21
Fix return value of std:string.

wangyan42164:
The return type of CodechalDebugConfigMgr::GetMediaStateStr() is
std:string. So it should not be nullptr which will cause segment
fault when dump debugging information.


wangyan42164:
#278 has fixed it based on the same logic. So close it.

PR #426, opened 2018-11-23
Clean the unused code.

wangyan42164:


media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -3267,11 +3267,6 @@ do_exec2(struct mos_linux_bo *bo, int used, struct mos_linux_context *ctx,
         i915_execbuffer2_set_context_id(execbuf, ctx->ctx_id);
     execbuf.rsvd2 = 0;
 
-#ifdef ANDROID
-    i915_execbuffer2_set_tag(execbuf, tag);
```
dvrogozh: If we don't call i915_execbuffer2_set_tag() remove definition as well.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -3267,11 +3267,6 @@ do_exec2(struct mos_linux_bo *bo, int used, struct mos_linux_context *ctx,
...
-    aub_exec(bo, flags, used);
```
dvrogozh: Can aub_exec be also removed? And the whole aub path?

media_driver/linux/common/os/mos_os_specific.c
```diff
@@ -5780,53 +5780,6 @@ MOS_VDBOX_NODE_IND Mos_Specific_GetVdboxNodeId(
         return idx;
     }
 
-#if defined(MEDIA_EXT) && !defined(ANDROID)
```
dvrogozh: There is other code for VDBOX balancing under MEDIA_EXT - I suggest to remove it.

media_driver/linux/common/os/mos_context_specific.cpp
```diff
@@ -475,26 +475,6 @@ void OsContextSpecific::SetSliceCount(uint32_t *pSliceCount)
         }
     }
 
-    struct drm_i915_gem_context_param_sseu sseu = { .flags = I915_EXEC_RENDER };
```
dvrogozh: I suggest to remove all the code under MEDIA_EXT for SSEU. Both here in this function and everywhere across the driver.

media_driver/linux/common/os/i915/mos_bufmgr.c
dvrogozh: This is not addressed. I understand that you remove this in the next patch along with the whole i915_ext.h, but commits should be self-contained. You remove _tag in this commit, thus you need to remove declaration in _this_ commit

wangyan42164:
I have finished the patch set based on @dvrogozh. So this PR is unnecessary.

dvrogozh:
I don't see your changes in the Git history! Cleanup is not finished, for example this file still exists: https://github.com/intel/media-driver/blob/master/media_driver/linux/common/os/i915/include/uapi/i915_drm_ext.h

We are DISCUSSING cleanup in this PR, please, update it with your most recent changes, let's review and merge it.

PR #427, opened 2018-11-23
[decode] Add full open build support

yangnin:
With build option FULL_OPEN_BUILD, driver will build with
Full Open

cmake -DFULL_OPEN_BUILD="yes" <workspace>/media-driver


media_driver/media_top_cmake.cmake
```diff
@@ -123,7 +123,7 @@ endif(NOT DEFINED INCLUDED_LIBS OR "${INCLUDED_LIBS}" STREQUAL "")
 # post target attributes
 bs_set_post_target()
 
-if(MEDIA_RUN_TEST_SUITE)
+if(MEDIA_RUN_TEST_SUITE AND ${FULL_OPEN_BUILD} STREQUAL "no")
```
dvrogozh: Let's introduce a cmake native option rather than something self made: https://cmake.org/cmake/help/v3.0/command/option.html

This should be on a top level CMakeLists.txt I believe:
`option(FULL_OPEN_BUILD "Build fully open source driver (so not include binary kernels without sources)" OFF)`

After that all conditions will simply look like:
```
if (FULL_OPEN_BUILD)
endif()
```

And cmake command line:
`cmake -DFULL_OPEN_BUILD=ON ...`

media_driver/agnostic/gen10/codec/kernel/igcodecdummykrn_g10.c
```diff
@@ -0,0 +1,38 @@
...
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF                          
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.              
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY                
+ * CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,                
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE                   
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                              
+*/                                                                                     
+                                                                                       
+////////////////////////////////////////////////////////////////////////////////       
+// !!! WARNING - AUTO GENERATED FILE. DO NOT EDIT DIRECTLY. !!!
```
dvrogozh: is it?? I believe it is not:).

media_driver/agnostic/gen10/codec/kernel/igcodecdummykrn_g10.c
```diff
@@ -0,0 +1,38 @@
...
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE                   
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                              
+*/                                                                                     
+                                                                                       
+////////////////////////////////////////////////////////////////////////////////       
+// !!! WARNING - AUTO GENERATED FILE. DO NOT EDIT DIRECTLY. !!!                        
+// Generated by KernelBinToSource.exe tool                                             
+////////////////////////////////////////////////////////////////////////////////       
+
+#ifdef IGFX_GEN10_SUPPORTED
```
dvrogozh: Why you have ifdef if both branches are the same?

media_driver/agnostic/gen10/codec/kernel/media_srcs.cmake
```diff
@@ -18,9 +18,20 @@
 # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
 
-set(TMP_SOURCES_
-    ${CMAKE_CURRENT_LIST_DIR}/igcodeckrn_g10.c
-)
+
+if(${FULL_OPEN_BUILD} STREQUAL "yes")
```
dvrogozh: Just `if(FULL_OPEN_BUILD)` if you will use a normal cmake option.

media_driver/agnostic/gen10/vp/kernel/igvpdummykrn_g10.c
```diff
@@ -0,0 +1,32 @@
...
+*/                                                                                     
+                                                                                       
+////////////////////////////////////////////////////////////////////////////////       
+// !!! WARNING - AUTO GENERATED FILE. DO NOT EDIT DIRECTLY. !!!                        
+// Generated by KernelBinToSource.exe tool                                             
+////////////////////////////////////////////////////////////////////////////////       
+
+extern const unsigned int IGVPKRN_G10_SIZE = 216;
+extern const unsigned int IGVPKRN_G10[] = {
+    0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000,
```
dvrogozh: you define codec kernels with 0 size and VP kernels with non-zero. Why?

And the second question: you _define_ kernels even though they are trivial. Which behavior we should expect? Such kernel will fail to load? or we will get runtime error? /I will be able to check myself later/

media_driver/agnostic/gen11/codec/kernel/media_srcs.cmake
```diff
@@ -18,9 +18,19 @@
 # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
 
-set(TMP_SOURCES_
-    ${CMAKE_CURRENT_LIST_DIR}/igcodeckrn_g11.c
-)
+if(${FULL_OPEN_BUILD} STREQUAL "yes")
+
+    set(TMP_SOURCES_
+        ${CMAKE_CURRENT_LIST_DIR}/igcodecdummykrn_g11.c
```
dvrogozh: Ok, why we define so many identical _dummy_ files for each platform?! This does not make sense to me and complicates everything. For example, it is hard to review which kernels are dummy and which are not. Could we PLEASE have single file for the dummy kernels? And we can use `#define IGCODECKRN_G11_SIZE IGCODECKRN_G10_SIZE` in it or some c++ alias trick.

Code should be simple. Right now driver has so many duplication...  

IMHO, just _look_ in the size of this _trivial_ PR! It is huge while actually it can be made visible on just 2 screens.

media_driver/agnostic/gen10/vp/kernel/igvpdummykrn_g10.c
```diff
@@ -0,0 +1,32 @@
...
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE                   
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                              
+*/                                                                                     
+                                                                                       
+////////////////////////////////////////////////////////////////////////////////       
+// !!! WARNING - AUTO GENERATED FILE. DO NOT EDIT DIRECTLY. !!!                        
+// Generated by KernelBinToSource.exe tool                                             
+////////////////////////////////////////////////////////////////////////////////       
+
+extern const unsigned int IGVPKRN_G10_SIZE = 216;
```
dvrogozh: What is 216? Can it be 0? If not, please explain the size meaning.

media_driver/media_top_cmake.cmake
dvrogozh: This comment was not addressed. Please, create a normal cmake option. Also, I suggest to change a n option name. Please, use ENABLE_KERNELS for it.

media_driver/agnostic/gen9_kbl/codec/kernel/media_srcs.cmake
```diff
@@ -18,9 +18,13 @@
 # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
 
-set(TMP_SOURCES_
-    ${CMAKE_CURRENT_LIST_DIR}/igcodeckrn_g9_kbl.c
-)
+if(${FULL_OPEN_BUILD} STREQUAL "no")
+
+    set(TMP_SOURCES_
+        ${CMAKE_CURRENT_LIST_DIR}/igcodeckrn_g9_kbl.c
```
dvrogozh: What about HUC?

XinfengZhang: hi @dvrogozh , why we need to care about HUC? we have some huc relevant code, but it is just valid when there are Huc firmware. it just like: there are no MBEnc media kernel with full_open_source, but the relevant code still there, just does not take effect.

yangnin: For Q1: What will be the behavior when user will try to use a feature which requires the shader? Please, describe this in a commit message.
Answer: Yes, we could add these options support in the next steps discussion.

dvrogozh: @XinfengZhang : we need care about HUC shaders because this invalidates "free" media driver - you have a binary blob inside the driver and that's actually it. This will:
1. Raise questions whether you actually use it at runtime or not. We will _need_ to explain. This at least gives delays in the discussions.
2, Raise suspicious whether we accidentally may try to cheat the Linux kernel requirements to have open source user.

IMHO, it is just simpler to strip it our _explicitly_. Let's avoid being asked questions when we can avoid them altogether.

media_driver/cmake/linux/media_feature_flags_linux.cmake
```diff
@@ -28,6 +28,7 @@ bs_set_if_undefined(Encode_VDEnc_Supported "yes")
 # The opposite is full-feature which permits some features only have binaries.
 # The default build is full-feature build so set Full_Open_Source_Support as "no".
 bs_set_if_undefined(Full_Open_Source_Support "no")
+bs_set_if_undefined(FULL_OPEN_BUILD "no")
```
dvrogozh: What's the difference between Full_Open_Source_Support and FULL_OPEN_BUILD? If this PR will be merged, this will become a mess!!! We **must** agree on approach first and _refactor_ rather than introduce "fast" solution.

media_driver/agnostic/gen9_kbl/codec/kernel/media_srcs.cmake
dvrogozh: @yangnin : you did not answer on the question and it is critical one actually. I repeat: "What will be the behavior when user will try to use a feature which requires the shader? Please, describe this in a commit message."

media_driver/media_top_cmake.cmake
vivijim: First of all... is there a better word than "kernel"?
Every open source user will get really confused...

media_driver/cmake/linux/media_feature_flags_linux.cmake
vivijim: +1 on @dvrogozh comment.... only one clear option should be enough

vivijim:
+1 on this question as well: "What will be the behavior when user will try to use a feature which requires the shader? Please, describe this in a commit message."
This is very useful to understand how useful the full open source driver is... what are the limitations? Everything should be crystal clear somewhere

media_driver/agnostic/common/shared/dummykrn.c
```diff
@@ -0,0 +1,92 @@
...
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF                          
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.              
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY                
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,                
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE                   
+// SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                              
+*/                                                                                     
+                                                                                       
+////////////////////////////////////////////////////////////////////////////////       
+// !!! WARNING - AUTO GENERATED FILE. DO NOT EDIT DIRECTLY. !!!
```
dvrogozh: I don't think this file is auto generated. Right? If so, please, remove these 4 warning lines.

media_driver/agnostic/common/shared/dummykrn.c
```diff
@@ -0,0 +1,86 @@
...
+// THE SOFTWARE IS PROVIDED 'AS IS', WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+const unsigned int IGCDUMMYKRN_SIZE = 216;
```
dvrogozh: Why not =0?

media_driver/agnostic/gen10/codec/kernel/media_srcs.cmake
```diff
@@ -18,13 +18,16 @@
...
-set(TMP_HEADERS_
-    ${CMAKE_CURRENT_LIST_DIR}/igcodeckrn_g10.h
-)
+if(ENABLE_KERNELS)
+    set(TMP_SOURCES_
+        ${CMAKE_CURRENT_LIST_DIR}/igcodeckrn_g10.c
+    )
+
+    set(TMP_HEADERS_
+        ${CMAKE_CURRENT_LIST_DIR}/igcodeckrn_g10.h
```
dvrogozh: We need to pass ENABLE_KERNELS into C/C++ code to be able to wrap inclusion of .h files:
```
#ifdef ENABLE_KERNELS
#include "igcodeckrn_g10.h"
#endif
```

PR #430, opened 2018-11-26
[Decode] Fix partition lengths for MFD_VP8_BSD_OBJECT commands.

tlonnber:
The original code tries to fix the encoding of the partition length fields too
late as the command has already been written to the command stream. The command
encoding is the same for different gens so the fix can be  implemented in the
generic encoding of the command. This patch fixes some of the GPU hangs when
decoding VP8 streams.


XinfengZhang:
LGTM will test and merge it

rosetta-jpn:
Hi, I am hiroh@. How long will it take to merge this change?

sreerenjb:
LGTM.
Carl, can you merge this?
+ I would recommend to push this https://github.com/intel/media-driver/pull/535

PR #453, opened 2018-12-09
build: provide BUILD_KERNELS option to rebuild shaders (kernels) from sources

dvrogozh:
BUILD_KERNELS option will permit to rebuild shaders (kernels) from sources. This option assumes that -DFREE_KERNELS=ON is also specified, i.e. configure as follows:
`cmake -DBUILD_KERNELS=ON -DFREE_KERNELS=ON ..`

I mark this PR as RFC since there are few things to be clarified:
1. My implementation allows driver sources modifications. I.e. shaders and their headers will be regenerated, you will notice difference by git-diff. This can be desirable if someone will want to regenerate shaders and push PR.
2. Once I implemented BUILD_KERNELS I found that driver compilation fails. Few other patches in PR tries to address this to the best of my knowledge. This needs to be reviewed.

And there are other things to review and discuss around kernels.... So, here is a chance...


media_driver/agnostic/media_srcs.cmake
```diff
@@ -18,6 +18,79 @@
...
+    # Notes:
+    #   - <platform> = gen<X>_<kind>, for example, gen11_icllp
+    # TODO:
+    #   - As of now this function works for the single platform only (gen11_icllp) while it
+    #  actully should process all the platforms at once. This single processing
+    #  is needed to generate shared ${krn_header} with the unified kernels definition.
+    function(gen_kernel_from_asm name platform)
+        platform_to_genx(${platform} genx kind)
+
+        set(krn_dir ${CMAKE_SOURCE_DIR}/media_driver/agnostic/${platform}/vp/kernel)
```
yangnin: It is not a good idea to set new generated files under our repo. it is better to select cmak out put folder as new generated files destination folder

media_driver/agnostic/media_srcs.cmake
```diff
@@ -18,6 +18,79 @@
...
+          OUTPUT ${dst}
+          DEPENDS KrnToHex_IGA ${src}
+          COMMAND ${IGA} -p ${genx} -a ${src} -o ${tgt}.krn
+          COMMAND KrnToHex_IGA ${tgt}.krn
+          COMMAND ${CMAKE_COMMAND} -E remove ${tgt}.krn
+          COMMAND ${CMAKE_COMMAND} -E rename ${tgt}.hex ${dst})
+    endfunction()
+
+    function(platform_to_genx platform genx kind)
+        if(platform STREQUAL "gen11_icllp")
```
yangnin: Better not set sepecific platform code in cmake file.

dvrogozh:
Final rebase done. Please, review. Mind that Tools and driver cleanup _is_ required since generated files do _ not_ match files stored in the repository.

dvrogozh:
@Sherry-Lin, @XinfengZhang : any other comments? when this will be merged?

PR #456, opened 2018-12-11
[VP]Change create temp surface in GetImage

Jister:


Jister:
Fixed by another commit 2947bd664c44c7602e079dca71a1824ac7296cbe

PR #463, opened 2018-12-13
Add HME support for ICLLP AVC VDEnc

cchunbo:
The change will include Downscaling and HME source code and enable the usage for AVC VDEnc on ICLLP


media_driver/agnostic/gen11/codec/hal/codechal_kernel_header_g11.h
```diff
@@ -29,6 +29,7 @@
 
 #include "codechal_encoder_base.h"
 
+#ifndef _FULL_OPEN_SOURCE
```
dvrogozh: no. Please, move these ifdefs inside structs.

Why you drop part of the structures declarations?

media_driver/agnostic/gen11/codec/hal/codechal_kernel_header_g11.h
```diff
@@ -129,15 +162,16 @@ static MOS_STATUS GetCommonKernelHeaderAndSizeG11(
     case VDENC_ME:
         currKrnHeader = &kernelHeaderTable->hmeVdenc;
         break;
+    case ENC_SCALING_CONVERSION:
+        currKrnHeader = &kernelHeaderTable->dsConvertGenX0;
+        break;
+#ifndef _FULL_OPEN_SOURCE
     case VDENC_STREAMIN_HEVC:
         currKrnHeader = &kernelHeaderTable->hmeHevcVdenc;
```
dvrogozh: Why you don't expose HME kernel for HEVC?

media_driver/agnostic/gen11/codec/hal/codechal_kernel_header_g11.h
```diff
@@ -160,7 +195,11 @@ static MOS_STATUS GetCommonKernelHeaderAndSizeG11(
     *((PCODECHAL_KERNEL_HEADER)krnHeader) = *currKrnHeader;
 
     PCODECHAL_KERNEL_HEADER invalidEntry;
+#ifdef _FULL_OPEN_SOURCE
     invalidEntry = &(kernelHeaderTable->weightedPrediction) + 1;
+#else
+    invalidEntry = &(kernelHeaderTable->dsConvertGenX0) + 1;
```
dvrogozh: Oh my God. That's really ugly!!! How about:
```
struct HmeDsScoreboardKernelHeaderG11 {
    int nKernelCount;
    union
    {
        struct
        {
            CODECHAL_KERNEL_HEADER hmeDownscaleGenX0;
            CODECHAL_KERNEL_HEADER hmeDownscaleGenX1;
            CODECHAL_KERNEL_HEADER hmeDownscaleGenX2;
            CODECHAL_KERNEL_HEADER hmeDownscaleGenX3;
            CODECHAL_KERNEL_HEADER hmeP;
            CODECHAL_KERNEL_HEADER hmeB;
            CODECHAL_KERNEL_HEADER hmeVdenc;
            CODECHAL_KERNEL_HEADER hmeHevcVdenc;
            CODECHAL_KERNEL_HEADER dsConvertGenX0;
#ifndef _FULL_OPEN_SOURCE
            CODECHAL_KERNEL_HEADER hmeDetectionGenX0;
            CODECHAL_KERNEL_HEADER initSwScoreboard;
            CODECHAL_KERNEL_HEADER intraDistortion;
            CODECHAL_KERNEL_HEADER dynamicScaling;
            CODECHAL_KERNEL_HEADER weightedPrediction;
#endif
            CODECHAL_KERNEL_HEADER invalidEntry;
        };
    };
};

invalidEntry = &kernelHeaderTable->invalidEntry;

```

media_driver/agnostic/gen11/codec/hal/codechal_vdenc_avc_g11.cpp
```diff
@@ -674,11 +674,7 @@ CodechalVdencAvcStateG11::CodechalVdencAvcStateG11(
 
     CODECHAL_ENCODE_ASSERT(m_osInterface);
 
-#ifndef _FULL_OPEN_SOURCE
     m_kernelBase = (uint8_t*)IGCODECKRN_G11;
```
dvrogozh:
```
#if defined(ENABLE_KERNELS)
m_kernelBase = (uint8_t*)IGCODECKRN_G11;
#endif
```

media_driver/agnostic/gen11/codec/hal/codechal_vdenc_avc_g11.cpp
```diff
@@ -692,6 +688,7 @@ CodechalVdencAvcStateG11::CodechalVdencAvcStateG11(
     Mos_SetVirtualEngineSupported(m_osInterface, true);
 
     m_vdboxOneDefaultUsed = true;
+    m_openCommonKernel = true;
```
dvrogozh: I REALLY don't understand what you are doing with this flag!

media_driver/agnostic/gen11/codec/media_srcs.cmake
```diff
@@ -19,5 +19,9 @@
 # OTHER DEALINGS IN THE SOFTWARE.
 
 media_include_subdirectory(hal)
-media_include_subdirectory(kernel)
+if("${Full_Open_Source_Support}" STREQUAL "yes")
+  media_include_subdirectory(kernel_open)
```
dvrogozh: I don't like word "open" in this context. Can we, please, use "free" instead?

media_driver/agnostic/common/codec/hal/codechal_encode_tracked_buffer.cpp
```diff
@@ -106,9 +106,10 @@ void CodechalEncodeTrackedBuffer::Resize()
             }
             ReleaseMvData(i);
             ReleaseDsRecon(i);
-#ifndef _FULL_OPEN_SOURCE
-            ReleaseSurfaceDS(i);
-#endif
+
+            if (m_encoder->m_openCommonKernel)
```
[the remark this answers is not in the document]
cchunbo: sounds good to remove "common", but I think m_openKernel will be better than m_freeKernel. common means the kernel shared by all codecs

media_driver/agnostic/common/codec/hal/codechal_encoder_base.cpp
```diff
@@ -1209,26 +1209,29 @@ MOS_STATUS CodechalEncoderState::CheckResChangeAndCsc()
...
         ResizeOnResChange();
     }
-#ifndef _FULL_OPEN_SOURCE
-    // check recon surface's alignment meet HW requirement
-    CODECHAL_ENCODE_CHK_STATUS_RETURN(m_cscDsState->CheckReconSurfaceAlignment(&m_reconSurface));
 
-    if (!m_cscDsState->IsEnabled() ||
-        CodecHal_PictureIsField(m_currOriginalPic) ||
-        CodecHal_PictureIsInterlacedFrame(m_currOriginalPic))
+    if (m_openCommonKernel)
```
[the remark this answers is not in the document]
cchunbo: From programming perspective, should be, but actually we have this check earlier when create the encoder instance, if it's nullptr, it will failed at initialization phase.

media_driver/agnostic/common/codec/hal/codechal_encoder_base.cpp
```diff
@@ -4567,6 +4568,12 @@ CodechalEncoderState::CodechalEncoderState(
     MOS_ZeroMemory(&m_vdencMeKernelBindingTable, sizeof(m_vdencMeKernelBindingTable));
 
     MOS_ZeroMemory(&m_vdencStreaminKernelBindingTable, sizeof(m_vdencStreaminKernelBindingTable));
+
+#ifndef _FULL_OPEN_SOURCE
+    m_openCommonKernel = true;
```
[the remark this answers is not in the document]
cchunbo: sounds good

media_driver/agnostic/common/codec/hal/codechal_encoder_base.cpp
```diff
@@ -4567,6 +4568,12 @@ CodechalEncoderState::CodechalEncoderState(
...
+#else
+    m_openCommonKernel = false;
+#endif
```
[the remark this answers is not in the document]
cchunbo: yes, I need to refine based on your previous PR

media_driver/agnostic/gen11/codec/hal/codechal_kernel_header_g11.h
cchunbo: Current plan only enable for AVC VDEnc, for HEVC VDEnc not committed

media_driver/agnostic/gen11/codec/hal/codechal_kernel_header_g11.h
cchunbo: Looks good, but we cannot change the original kernel definition sequences, it will require current kernel binary/tools change, not only for Linux open source, also will impact Windows

media_driver/agnostic/gen11/codec/hal/codechal_vdenc_avc_g11.cpp
cchunbo: it's the same purpose as the "FREE_KERNEL" MACRO proposed, however, with this flag I can control by each platform

media_driver/agnostic/gen11/codec/media_srcs.cmake
cchunbo: Can you explain more details why prefer "free" to "open" ?

media_driver/agnostic/common/codec/hal/codechal_encoder_base.cpp
```diff
@@ -4567,6 +4568,12 @@ CodechalEncoderState::CodechalEncoderState(
     MOS_ZeroMemory(&m_vdencMeKernelBindingTable, sizeof(m_vdencMeKernelBindingTable));
 
     MOS_ZeroMemory(&m_vdencStreaminKernelBindingTable, sizeof(m_vdencStreaminKernelBindingTable));
+
+#ifndef _FULL_OPEN_SOURCE
```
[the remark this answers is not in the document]
cchunbo: The logic is that DS and HME are kernel tasks, only when the kernels enabled, the clauses will be executed, in this case, suppose use the m_freeKernel flag to indicate the kernels are enabled, then the logic will be if (m_freeKernels) { ... then kernel tasks execution}

media_driver/agnostic/gen11/codec/hal/codechal_encode_csc_ds_g11.cpp
```diff
@@ -29,9 +29,7 @@
 #include "codechal_encode_csc_ds_g11.h"
 #include "codechal_kernel_header_g11.h"
 #include "codeckrnheader.h"
-#ifndef _FULL_OPEN_SOURCE
 #include "igcodeckrn_g11.h"
```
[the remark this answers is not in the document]
cchunbo: Yes, I need to refine after #455 checked in.
However I don't think ENABLE_KERNEL can work as expected when we need to enable the *free*(*open*) kernels, it should be
#if defined(ENABLE_KERNELS) || defined(OPEN_KERNEL)
#include "igcodeckrn_g11.h"
#endif

media_driver/agnostic/gen11/codec/hal/codechal_encode_csc_ds_g11.cpp
```diff
@@ -743,11 +741,7 @@ CodechalEncodeCscDsG11::CodechalEncodeCscDsG11(CodechalEncoderState* encoder)
 {
     m_cscKernelUID = IDR_CODEC_HME_DS_SCOREBOARD_KERNEL;
     m_cscCurbeLength = sizeof(CscKernelCurbeData);
-#ifndef _FULL_OPEN_SOURCE
     m_kernelBase = (uint8_t*)IGCODECKRN_G11;
```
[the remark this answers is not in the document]
cchunbo: save explanation as above

media_driver/agnostic/gen11/codec/hal/codechal_kernel_header_g11.h
cchunbo: the reason is that we have different kernel structure, we only have DS and HME open sourced, the structure should align with the kernels enabled

media_driver/agnostic/common/codec/hal/codechal_encode_tracked_buffer.cpp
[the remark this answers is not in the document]
dvrogozh: You know, I am still confused. We will have 3 build types:
1. No kernels at all (ENABLE_KERNELS=OFF)
2. Only kernels w/ open sources are available (FREE_KERNELS=ON)
3. Full feature build when we use pre-built kernels containing all source, for some of them we will not publish sources (default build)

So, in my mind your flag m_encoder->m_openCommonKernel should be set as follows:
1. m_openCommonKernel=false for "no kernels at all (ENABLE_KERNELS=OFF)"
2. m_openCommonKernel=true for "only kernels w/ open sources are available (FREE_KERNELS=ON)"
3. m_openCommonKernel=true for "full feature build when we use pre-built kernels containing all source, for some of them we will not publish sources (default build)"

However, I am not sure that this is what you had in mind and actually implemented. Moreover, the above usage is confusing. Really, see yourself:
- In case of build 2) and 3) you should not differentiate the code which works with HME since feature presents in both open and non-open kernels
- In case of 1) you natively should expect different flag like if(m_kernelsEnabled) meaning whether any kernels are enabled at all

That's the origin of my confusion. Here is what I naturally expect to see in the code:
```
if (m_kernelsEnabled && !m_openKernels) {
  //program some kernel for which we don't publish sources
}

if (m_kernelsEnabled) {
  //program some kernel for which we publish sources: no difference for 2) and 3) since kernel is available
}

bool m_kernelsEnabled; // =true if kernels enabled in the build, =false if kernels are not enabled at all
bool m_openKernels; // =true if _only_ open kernel enabled in the build, =false if both open/closed kernels are enabled
```

Eventually you can consider to use ENABLE_KERNELS macro instead of bool variable, but idea still holds.

media_driver/agnostic/gen11/codec/hal/codechal_kernel_header_g11.h
dvrogozh: In that case we need to refactor for Windows as well. What exist in the code now is simply wrong and error prone. Actually, I don't quite understand what's the reason behind having pointer to the invalid entry?! Can you, please, explain more? In a way, what you have in a code now gives a question: do you have access violation hidden in the code? Why you store pointer to the address _after_ your structure if you can have sizeof(structure)?

media_driver/agnostic/gen11/codec/hal/codechal_vdenc_avc_g11.cpp
dvrogozh: I posted big comment above. Shortly: how you name the flag and how you use it in the code is _misleading_. "common" is misleading, "kernel" instead of "kernel>>>s<<<" is misleading. And having if (m_openCommonKernel) instead of if ( >>>!<<< m_openCommonKernel) is misleading.

IMHO, why you cover code which should be executed for both cases when you have open source kernels and when you have closed source kernels?! This is illogical. Expectation is that you will strip out the code which should _not_ be executed for one of the path, i.e. the code for closed source part.

media_driver/agnostic/common/codec/hal/codechal_encode_tracked_buffer.cpp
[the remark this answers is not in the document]
cchunbo: generally it's what I try to implement, only difference is that for free kernel, currently we only enabled ICLLP AVC VDEnc, for other codecs and platforms, it's not enabled, that's why I am using the variable rather than the MARCO to control it. Eventually if we have the free kernels for all platforms and codecs, we definitely can replace with the MARCO FREE_KERNELS

media_driver/agnostic/common/codec/hal/codechal_utilities.cpp
```diff
@@ -210,7 +210,15 @@ MOS_STATUS CodecHalGetKernelBinaryAndSize(
     uint8_t**  kernelBinary,
     uint32_t*  size)
 {
-#ifndef _FULL_OPEN_SOURCE
+#ifdef _FULL_OPEN_SOURCE
+    if (kernelBase == nullptr)
```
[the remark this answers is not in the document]
cchunbo: It's somehow tricky here, this function is very fundamental here, it won't have any side effect for FREE_KERNEL defined case if I return as successful here

media_driver/agnostic/common/codec/hal/codechal_encoder_base.cpp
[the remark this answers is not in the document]
chivakker: The name is confusing.  media kernels are always available.  the difference is that they are compiled or not.  Please change variable names to reflect that.

media_driver/agnostic/common/codec/hal/codechal_encoder_base.h
```diff
@@ -1682,6 +1682,8 @@ class CodechalEncoderState : public Codechal
 
     bool m_colorbitSupported;
 
+    bool m_freeKernels; //!< Used to indicate whether has free kernels, it will be always true if ENABLE_KERNELS defined
```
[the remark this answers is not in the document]
chivakker: Please properly rename.

media_driver/agnostic/gen11/codec/kernel_free/Source/downscale_convert_kernel_genx.cpp
```diff
@@ -0,0 +1,898 @@
+/*
+* Copyright (c) 2017, Intel Corporation
```
chivakker: This is 2018, check all files

media_driver/agnostic/gen11/codec/hal/codechal_encode_csc_ds_g11.cpp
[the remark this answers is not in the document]
dvrogozh: The problem in the current code is that you _will_ allow kernels to slip into the build if ENABLE_KERNELS=OFF and FREE_KERNELS=ON. And this is wrong. You need to use either:
```
#if defined(ENABLE_KERNELS) 
#include "igcodeckrn_g11.h"
#endif
```
or
```
#if defined(ENABLE_KERNELS) 
#if defined(_FULL_OPEN_SOURCE)
// do something
#else
#include "igcodeckrn_g11.h"
#endif
#endif
```

media_driver/agnostic/common/codec/hal/codechal_encoder_base.cpp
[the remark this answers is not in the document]
cchunbo: looks good to me to rename it

media_driver/agnostic/gen11/codec/hal/codechal_encode_csc_ds_g11.cpp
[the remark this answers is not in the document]
cchunbo: #if defined(ENABLE_KERNELS) 
#include "igcodeckrn_g11.h"
#endif

this one looks better

media_driver/agnostic/gen11/codec/kernel_free/Source/downscale_convert_kernel_genx.cpp
cchunbo: sure, thanks

media_driver/agnostic/gen11/codec/hal/codechal_vdenc_avc_g11.cpp
```diff
@@ -1075,7 +1076,7 @@ MOS_STATUS CodechalVdencAvcStateG11::ExecuteSliceLevel()
             &flushDwParams));
     }
 
-#if defined(ENABLE_KERNELS) && !defined(_FULL_OPEN_SOURCE)
+#if defined(ENABLE_KERNELS) || defined(_FULL_OPEN_SOURCE)
```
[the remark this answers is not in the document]
cchunbo: Should not revert, should be #if defined(ENABLE_KERNELS)

media_driver/agnostic/gen11/codec/hal/codechal_vdenc_avc_g11.cpp
```diff
@@ -676,7 +676,8 @@ CodechalVdencAvcStateG11::CodechalVdencAvcStateG11(
 
     CODECHAL_ENCODE_ASSERT(m_osInterface);
 
-#if defined(ENABLE_KERNELS) && !defined(_FULL_OPEN_SOURCE)
+#if defined(ENABLE_KERNELS) || defined(_FULL_OPEN_SOURCE)
+    m_freeKernels = true;
```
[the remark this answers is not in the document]
cchunbo: I have rename it to m_enableHmeKernel, and in the base class you can see:
`#if ENABLE_KERNELS
    #if _FULL_OPEN_SOURCE
        m_enableHmeKernel = false;
    #else
        m_enableHmeKernel = true;
    #endif
#endif`

When we only open free kernels, by default m_enableHmeKernel is disabled in the base, because we only enable it for ICL AVC VDEnc, that's why I add the assignment here to enable it for Gen11 AVC VDEnc

media_driver/agnostic/gen11/codec/hal/codechal_vdenc_avc_g11.cpp
```diff
@@ -31,7 +31,7 @@
 #include "mhw_vdbox_vdenc_g11_X.h"
 #include "mhw_vdbox_g11_X.h"
 #include "mos_util_user_interface.h"
-#if defined(ENABLE_KERNELS) && !defined(_FULL_OPEN_SOURCE)
```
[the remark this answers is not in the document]
cchunbo: As we discussed, will change to #if defined(ENABLE_KERNELS)

media_driver/agnostic/gen11/codec/hal/codechal_kernel_header_g11.h
```diff
@@ -36,19 +36,27 @@ struct HmeDsScoreboardKernelHeaderG11 {
         struct
         {
             CODECHAL_KERNEL_HEADER hmeDownscaleGenX0;
+#if defined(ENABLE_KERNELS) && !defined(_FULL_OPEN_SOURCE)
             CODECHAL_KERNEL_HEADER hmeDownscaleGenX1;
+#endif
             CODECHAL_KERNEL_HEADER hmeDownscaleGenX2;
+#if defined(ENABLE_KERNELS) && !defined(_FULL_OPEN_SOURCE)
```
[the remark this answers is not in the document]
cchunbo: yes, it could be gather together, but due to other projects impact, it will not be included in this change.  Would rather to refine from windows side.

media_driver/agnostic/common/codec/hal/codechal_encoder_base.cpp
```diff
@@ -4567,6 +4572,15 @@ CodechalEncoderState::CodechalEncoderState(
     MOS_ZeroMemory(&m_vdencMeKernelBindingTable, sizeof(m_vdencMeKernelBindingTable));
 
     MOS_ZeroMemory(&m_vdencStreaminKernelBindingTable, sizeof(m_vdencStreaminKernelBindingTable));
+
+
+#if ENABLE_KERNELS
+    #if _FULL_OPEN_SOURCE
+        m_enableHmeKernel = false;
```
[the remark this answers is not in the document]
cchunbo: Not exactly, actually FREE_KERNELS only enabled for ICLLP AVC VDEnc, for other cases, still no DS+HME kernel for _FULL_OPEN_SOURCE case

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
```diff
@@ -0,0 +1,44 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+# OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+
+if(ENABLE_KERNELS)
```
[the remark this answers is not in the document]
cchunbo: make sense , thanks

media_driver/agnostic/common/codec/hal/codechal_encode_tracked_buffer.cpp
```diff
@@ -106,9 +106,9 @@ void CodechalEncodeTrackedBuffer::Resize()
             }
             ReleaseMvData(i);
             ReleaseDsRecon(i);
-#ifndef _FULL_OPEN_SOURCE
-            ReleaseSurfaceDS(i);
-#endif
+            if (m_encoder->m_enableHmeKernel)
```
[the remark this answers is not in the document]
cchunbo: thanks, it sounds makes sense

media_driver/agnostic/common/codec/hal/codechal_encoder_base.cpp
```diff
@@ -1209,26 +1209,30 @@ MOS_STATUS CodechalEncoderState::CheckResChangeAndCsc()
...
         ResizeOnResChange();
     }
-#ifndef _FULL_OPEN_SOURCE
-    // check recon surface's alignment meet HW requirement
-    CODECHAL_ENCODE_CHK_STATUS_RETURN(m_cscDsState->CheckReconSurfaceAlignment(&m_reconSurface));
 
-    if (!m_cscDsState->IsEnabled() ||
-        CodecHal_PictureIsField(m_currOriginalPic) ||
-        CodecHal_PictureIsInterlacedFrame(m_currOriginalPic))
+    if (m_enableHmeKernel && m_cscDsState)
```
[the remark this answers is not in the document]
cchunbo: eventually I removed the m_enableHmeKernel flag, and use m_cscDsState instead, they have the same effect.

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
```diff
@@ -18,6 +18,35 @@
...
+
+        set(cm_krn_dir ${CMAKE_SOURCE_DIR}/media_driver/agnostic/${platform}/codec/kernel_free)
+        set(cm_krn ig${name}krn_g11)
+
+        # Generating source from the .krn file
+        add_custom_command(
+            OUTPUT ${cm_krn_dir}/${cm_krn}.c ${cm_krn_dir}/${cm_krn}.h
+            DEPENDS KernelBinToSource
+            COMMAND ${CMAKE_COMMAND} -E rename ${CMAKE_CURRENT_LIST_DIR}/commonkernel.krn ${CMAKE_CURRENT_LIST_DIR}/${cm_krn}.krn
+            COMMAND KernelBinToSource ARGS -i ${CMAKE_CURRENT_LIST_DIR}/${cm_krn}.krn -o ${cm_krn_dir} -v ${cm_krn} -index 14 -t 18
```
dvrogozh: What is 14? What is 18? This hardcoding wont't do, guys. It does not make sense and is not intuitive. Please, explain what you are trying to achieve?

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
```diff
@@ -18,6 +18,35 @@
...
+
+    function(gen_cm_kernel_from_source name platform)
+
+        set(cm_krn_dir ${CMAKE_SOURCE_DIR}/media_driver/agnostic/${platform}/codec/kernel_free)
+        set(cm_krn ig${name}krn_g11)
+
+        # Generating source from the .krn file
+        add_custom_command(
+            OUTPUT ${cm_krn_dir}/${cm_krn}.c ${cm_krn_dir}/${cm_krn}.h
+            DEPENDS KernelBinToSource
```
dvrogozh: You need to list kernel source files in the DEPENDS otherwise you may not actually regenerate anything..

media_driver/agnostic/gen11/codec/kernel_free/build.py
```diff
@@ -0,0 +1,199 @@
...
+        isExists=os.path.exists("Binary")
+        if not isExists:
+            os.mkdir("Binary")
+
+        shutil.copy('downscale_kernel_genx_0.dat', 'Binary/DS4x_Frame.krn')
+        shutil.copy('downscale_kernel_genx_1.dat', 'Binary/DS2x_Frame.krn')
+        shutil.copy('hme_kernel_genx_0.dat', 'Binary/HME_P.krn')
+        shutil.copy('hme_kernel_genx_1.dat', 'Binary/HME_B.krn')
+        shutil.copy('hme_kernel_genx_2.dat', 'Binary/HME_VDENC.krn')
+        shutil.copy('downscale_convert_kernel_genx_0.dat', 'Binary/DS_Convert.krn')
```
dvrogozh: Remove the hardcode. Provide parameters to the script to set which kernels should be processed.

media_driver/agnostic/gen11/codec/kernel_free/build.py
```diff
@@ -0,0 +1,199 @@
...
+        output = open('commonkernel.krn', 'wb')
+
+        bytes  = struct.pack('i', 6)
+        output.write(bytes)
+
+        count  = 64
+        bytes  = struct.pack('i', count)
+        output.write(bytes)
+
+        filepath = os.path.join('./Binary', 'DS4x_Frame.krn')
```
dvrogozh: This block from line 61 to line 67 is a common pattern and should be extracted to the separate function.

media_driver/agnostic/gen11/codec/kernel_free/build.py
```diff
@@ -0,0 +1,199 @@
+import os,shutil
```
dvrogozh: Why python version is supposed? python-2 or python-3?

CMakeLists.txt
```diff
@@ -66,6 +66,16 @@ if (BUILD_KERNELS)
         message(FATAL_ERROR "Failed to find IGA")
     endif()
 
+    find_program(CMC cmc)
+    if (NOT CMC)
+        message(FATAL_ERROR "Failed to find CMC")
+    endif()
+
+    find_program(PYTHON python)
```
dvrogozh: which python we actually require? 2 or 3? Optimally we would wish to use FindPythonN cmake module (https://cmake.org/cmake/help/v3.12/module/FindPython3.html). The problem is that we will need to bump cmake version up to 3.13 where they appear.

Tools/MediaDriverTools/KernelBinToSource/KernelBinToSource.cpp
```diff
@@ -445,10 +569,15 @@ void printUsage(const std::string &sProgram)
...
               << " [" << PARAM_V << " VarName]"
+              << " [" << PARAM_INDEX << " Kernel Index]"
+              << " [" << PARAM_TOTAL << " Kernel Total]"
               << std::endl
               << "    " << PARAM_I << " Path to Kernel binary input file (required)"             << std::endl
               << "    " << PARAM_O << " Path to Kernel binary output directory (optional)"       << std::endl
-              << "    " << PARAM_V << " Variable Name on the generated source file (optional)"   << std::endl;
+              << "    " << PARAM_V << " Variable Name on the generated source file (optional)"   << std::endl
+              << "    " << PARAM_INDEX << " Variab kernel Index  (optional)"                     << std::endl
+              << "    " << PARAM_TOTAL << " Variable kernel total count (optional)"              << std::endl;
```
dvrogozh: I don't understand what these variables do. Especially I don't understand what Index is for (total is I guess just total kernels count, but Index is a mystery to me).

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
```diff
@@ -18,6 +18,35 @@
...
+            COMMAND KernelBinToSource ARGS -i ${CMAKE_CURRENT_LIST_DIR}/${cm_krn}.krn -o ${cm_krn_dir} -v ${cm_krn} -index 14 -t 18
+            COMMAND ${CMAKE_COMMAND} -E remove ${CMAKE_CURRENT_LIST_DIR}/${cm_krn}.krn
+            COMMAND ${CMAKE_COMMAND} -E remove_directory ${CMAKE_CURRENT_LIST_DIR}/Binary
+            COMMENT "Generate source file from krn")
+
+    endfunction()
+
+    if(GEN11_ICLLP)
+
+        exec_program(${PYTHON} ${CMAKE_CURRENT_LIST_DIR}
```
dvrogozh: In this way you will run this command within configuration stage. I.e. you mix configuration and compilation stages. This will not do. Please, refactor in a way which will just configure targets and build them within make.

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
```diff
@@ -18,6 +18,35 @@
...
+                    ARGS build.py)
```
dvrogozh: Configuration fails on my side with:
```
running cd "/home/dvrogozh/git/github/media-driver/media_driver/agnostic/gen11/codec/kernel_free" && /usr/bin/python build.py  2>&1
./Source/downscale_convert_kernel_genx.cpp(23,10):  fatal error: 'cm/cm.h' file not found
#include <cm/cm.h>
         ^~~~~~~~~
1 error generated.
./Source/downscale_kernel_genx.cpp(23,10):  fatal error: 'cm/cm.h' file not found
#include <cm/cm.h>
         ^~~~~~~~~
1 error generated.
./Source/hme_kernel_genx.cpp(23,10):  fatal error: 'cm/cm.h' file not found
#include <cm/cm.h>
         ^~~~~~~~~
1 error generated.
```

I may need to update CM compiler though...

dvrogozh:
@cchunbo: I see that you have updated the PR, but I don't see actual changes. What did you change? or that was just a rebase? I believe that feedback I gave was not yet addressed...

media_driver/agnostic/common/codec/hal/codechal_vdenc_avc.cpp
```diff
@@ -1233,6 +1235,9 @@ MOS_STATUS CodechalVdencAvcState::Initialize(CodechalSetting * settings)
             &userFeatureData);
         m_staticFrameDetectionEnable = (userFeatureData.i32Data) ? true : false;
 
+#if (_FULL_OPEN_SOURCE)
```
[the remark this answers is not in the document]
cchunbo: looks better

media_driver/agnostic/gen11_icllp/media_interfaces/media_interfaces_g11_icllp.cpp
```diff
@@ -474,14 +474,21 @@ MOS_STATUS CodechalInterfacesG11Icllp::Initialize(
             return MOS_STATUS_INVALID_PARAMETER;
         }
 
+#if defined(ENABLE_KERNELS)
+
+    #if defined(_FULL_OPEN_SOURCE)
```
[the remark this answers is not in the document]
cchunbo: no, it's not correct, when we have _FULL_OPEN_SOURCE defined, only AVC VDEnc need to create the  DS kernel, in your suggestion, it's just the opposite direction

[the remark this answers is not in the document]
dvrogozh: I misprinted in the code sample, sorry. I meant:
```
#ifndef _FULL_OPEN_SOURCE
    if (info->Mode != CODECHAL_ENCODE_MODE_JPEG)
#else
    if (info->Mode == CODECHAL_ENCODE_MODE_AVC)
#endif
```
I suggested to have `#ifndef _FULL_OPEN_SOURCE` thru the driver code whenever possible.

Tools/MediaDriverTools/KernelBinToSource/KernelBinToSource.cpp
```diff
@@ -171,6 +174,119 @@ int32_t writeHeaderFile(
...
+            {
+                oSs << ", ";
+            }
+        }
+    }
+
+    oSs << std::endl << "};" << std::endl;
+
+    oSs << "#else" << std::endl
+        << sSizeName.c_str() << " = 216;" << std::endl
```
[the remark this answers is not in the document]
dvrogozh: Please, respond.

Tools/MediaDriverTools/KernelBinToSource/KernelBinToSource.cpp
dvrogozh: Please, respond.

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
dvrogozh: Please, respond.

Tools/MediaDriverTools/KernelBinToSource/KernelBinToSource.cpp
yangnin: Index is for kernel offset. this situation we use 14 as kernel index

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
yangnin: For media driver, it is not compilation. It is still in configration. if no kernel changes, file will not be chagned.

Tools/MediaDriverTools/KernelBinToSource/KernelBinToSource.cpp
dvrogozh: I still don't understand what is index needed for. Why it is 14? Please, explain.

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
dvrogozh: That's exactly the issue: you are running compilation commands within configuration stage. That's wrong.

dvrogozh: To avoid exec_program you should specify input/output dependencies for the add_custom_command:
* OUTPUT: kernel .h/.c files
* DEPENDENCIES: KernelBinToSource <cm .c files>

In this way make will know that to to produce ihd_drv_video.so it need to produce kernel .h/.c files first and to do that it should run your custom commands, i.e. generate KernlBinToSource application and use .cm files as input dependencies.

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
```diff
@@ -18,6 +18,33 @@
...
+        set(cm_krn_dir ${CMAKE_SOURCE_DIR}/media_driver/agnostic/${platform}/codec/kernel_free)
+        set(cm_krn ig${name}krn_g11)
+
+        # Generating source from the .krn file
+        add_custom_command(
+            OUTPUT ${cm_krn_dir}/${cm_krn}.c ${cm_krn_dir}/${cm_krn}.h
+            DEPENDS KernelBinToSource
+            COMMAND cp -r ${CMAKE_CURRENT_LIST_DIR}/Source .
+            COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_LIST_DIR}/build.py .
+            COMMAND ${PYTHON} ARGS build.py
```
dvrogozh: Please, avoid copying of the python script with:
`COMMAND ${PYTHON} ARGS ${CMAKE_CURRENT_LIST_DIR}/build.py`

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
```diff
@@ -18,6 +18,33 @@
...
+            COMMAND ${CMAKE_COMMAND} -E rename commonkernel.krn ${cm_krn}.krn
+            COMMAND KernelBinToSource ARGS -i ${cm_krn}.krn -o . -v ${cm_krn} -index 14 -t 17
```
dvrogozh: Add a comment before this line:
```
# -index 14 -t 17 are needed to match a description of the kernel in
# media_driver/agnostic/common/codec/kernel/codeckrnheader.h:
# 1. We generate a kernel with index 14 (IDR_CODEC_HME_DS_SCOREBOARD_KERNEL)
# 2. And the total number of kernels known to the media driver is 17 (IDR_CODEC_TOTAL_NUM_KERNELS)
```

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
```diff
@@ -18,6 +18,33 @@
...
+            COMMAND ${CMAKE_COMMAND} -E copy ${cm_krn}.c ${cm_krn}.h ${CMAKE_CURRENT_LIST_DIR}/
```
dvrogozh: You probably should not need this command. You can have just:
`COMMAND KernelBinToSource ARGS -i ${cm_krn}.krn -o ${CMAKE_CURRENT_LIST_DIR}/ -v ${cm_krn} -index 14 -t 17`

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
```diff
@@ -18,6 +18,33 @@
 # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
 
+if (BUILD_KERNELS)
+
+    function(gen_cm_kernel_from_source name platform)
```
dvrogozh: Please, move this function definition to https://github.com/intel/media-driver/blob/master/media_driver/agnostic/media_srcs.cmake

media_driver/agnostic/gen11/codec/kernel_free/media_srcs.cmake
```diff
@@ -18,6 +18,33 @@
...
+
+        set(cm_krn_dir ${CMAKE_SOURCE_DIR}/media_driver/agnostic/${platform}/codec/kernel_free)
+        set(cm_krn ig${name}krn_g11)
```
dvrogozh: Use instead:
```
platform_to_genx(${platform} genx kind)
set(cm_krn ig${name}krn_g${genx})
```

You may need to adjust `platform_to_genx` to recognize gen11 (it now knows only gen11_icllp).

media_driver/agnostic/gen11/codec/kernel_free/build.py
```diff
@@ -0,0 +1,199 @@
...
+class GenSource2Bin(object):
+    def __init__(self, component='codec'):
+        self.fromdir    = './Binary'
+
+    def iga_compile(self):
+        for filename in os.listdir('./Source'):
+            fp = os.path.join('./Source', filename)
+            fileext = os.path.splitext(filename)[1]
+            if os.path.isfile(fp) and 'cpp' in fileext:
+                command  = 'cmc /c /Qxcm -Qxcm_jit_target=gen11lp '
```
dvrogozh: -Qxcm_jit_target=gen11lp - why?! The kernel is inside gen11 folder. So, for which platform this kernel is compiled?

media_driver/agnostic/gen11/codec/kernel_free/build.py
```diff
@@ -0,0 +1,199 @@
...
+
+            if os.path.isfile(fp) and 'isa' in fileext:
+                os.remove(fp)
+
+    def join(self):
+        isExists=os.path.exists("Binary")
+        if not isExists:
+            os.mkdir("Binary")
+
+        shutil.copy('downscale_kernel_genx_0.dat', 'Binary/DS4x_Frame.krn')
```
dvrogozh: Does the order matters? If yes, how the driver knows about it?

dvrogozh:
Closing in favor of #489.

Issue #452, opened 2018-12-09
Duplicate VP shaders (kernels) in media driver


dvrogozh:
Looks like there are duplicated shaders (kernels) declared in media driver. At least these:

https://github.com/intel/media-driver/blob/bc71759059b6490d84c2f8547e69116f629527ee/media_driver/agnostic/common/vp/kernel/vpkrnheader.h#L593
https://github.com/intel/media-driver/blob/bc71759059b6490d84c2f8547e69116f629527ee/media_driver/agnostic/common/vp/kernel/vpkrnheader.h#L598

**Question:** which sources are actually published in https://github.com/intel/media-driver/blob/bc71759059b6490d84c2f8547e69116f629527ee/media_driver/agnostic/gen11_icllp/vp/kernel/Source/Save_444Scale16_ARGB_64Byte.asm?

https://github.com/intel/media-driver/blob/bc71759059b6490d84c2f8547e69116f629527ee/media_driver/agnostic/common/vp/kernel/vpkrnheader.h#L594
https://github.com/intel/media-driver/blob/bc71759059b6490d84c2f8547e69116f629527ee/media_driver/agnostic/common/vp/kernel/vpkrnheader.h#L610

**Question:** which sources are actually published in https://github.com/intel/media-driver/blob/bc71759059b6490d84c2f8547e69116f629527ee/media_driver/agnostic/gen11_icllp/vp/kernel/Source/Save_444Scale16_NV21.asm

https://github.com/intel/media-driver/blob/bc71759059b6490d84c2f8547e69116f629527ee/media_driver/agnostic/common/vp/kernel/vpkrnheader.h#L595
https://github.com/intel/media-driver/blob/bc71759059b6490d84c2f8547e69116f629527ee/media_driver/agnostic/common/vp/kernel/vpkrnheader.h#L625

**Question:** which sources are actually published in https://github.com/intel/media-driver/blob/bc71759059b6490d84c2f8547e69116f629527ee/media_driver/agnostic/gen11_icllp/vp/kernel/Source/Save_444Scale16_RGB_64Byte.asm

https://github.com/intel/media-driver/blob/bc71759059b6490d84c2f8547e69116f629527ee/media_driver/agnostic/common/vp/kernel/vpkrnheader.h#L596
https://github.com/intel/media-driver/blob/bc71759059b6490d84c2f8547e69116f629527ee/media_driver/agnostic/common/vp/kernel/vpkrnheader.h#L629

[...]

PR #467, opened 2018-12-14
[VP] Correct kernel

ljia5:
Correct kernel


Tools/MediaDriverTools/KrnToHex_IGA/KrnToHex_IGA.cpp
```diff
@@ -47,8 +47,8 @@ bool KrnToHex(char *file_krn, char *file_hex)
     char     line[40];
     FILE     *f_in, *f_out;
 
-    unsigned long   *pbuffer;
-    unsigned long    n_read;
+    unsigned int   *pbuffer;
```
dvrogozh: Can you, please add Fixes: #452 to the commit message of this patch:
```
[VP] Correct kernel

Fixes: #452 
```

This will automatically close the issue and associate commit with it.

jackie-guo: Sure, I think Lin will help add "Fixes: #452" later.

XinfengZhang:
commit message is too simple
please add more, for example, xxx to keep kernel order ...

PR #486, opened 2019-01-09
[VP9d] Implement continuous ring buffer logic for render target table

vshampor:
This also refactors the render target table content a bit - now it
also contains the VASurfaceID along with the surface pointer and the
usage flag.


vshampor:
This PR impacts *a lot* of files. While the devult tests are passing, I strongly recommend to run a *full* driver validation with this commit in place.

vshampor:
@XinfengZhang , what is the status on this one?

XinfengZhang:
LGTM

media_driver/linux/common/ddi/media_ddi_render_target_table.cpp
```diff
@@ -0,0 +1,362 @@
...
+        return VA_STATUS_SUCCESS;
+    }
+    else
+    {
+        using RTPair = std::pair<VASurfaceID, DDI_CODEC_RENDER_TARGET_INFO>;
+        using RTPairIterator = std::map<VASurfaceID, DDI_CODEC_RENDER_TARGET_INFO>::iterator;
+        std::vector<RTPairIterator> inactive_rt_vec;
+        inactive_rt_vec.reserve(m_va_to_rt_map.size());
+
+        for (RTPairIterator it = m_va_to_rt_map.begin(); it != m_va_to_rt_map.end(); it++)
```
dmitry-gurulev: We can find the smallest `FrameIdx` right in this loop, why we need additional `inactive_rt_vec` + `min_element`?

media_driver/linux/common/ddi/media_ddi_render_target_table.cpp
```diff
@@ -0,0 +1,362 @@
...
+//!
+VAStatus DDI_CODEC_RENDER_TARGET_TABLE::UnRegisterRTSurface(VASurfaceID id)
+{
+    if (!IsRegistered(id))
+    {
+        DDI_VERBOSEMESSAGE("UnRegisterRTSurface: render target was not registered in the RTtbl!");
+        return VA_STATUS_ERROR_INVALID_PARAMETER;
+    }
+
+    m_free_index_pool.push_front(m_va_to_rt_map[id].FrameIdx);
```
dmitry-gurulev: u push/pop elements from the same end of container (front) - what is intention to use `deque` instead of `vector` + `push_back/pop_back`

media_driver/linux/common/codec/ddi/media_ddi_decode_base.cpp
```diff
@@ -70,6 +70,15 @@ VAStatus DdiMediaDecode::BasicInit(
         return VA_STATUS_ERROR_ALLOCATION_FAILED;
     }
 
+    m_ddiDecodeCtx->pRTtbl = MOS_New(DDI_CODEC_RENDER_TARGET_TABLE);
+
+    if (m_ddiDecodeCtx->pRTtbl == nullptr)
+    {
+        MOS_Delete(m_ddiDecodeCtx->pRTtbl);
+        m_ddiDecodeCtx->pRTtbl = nullptr;
+        return VA_STATUS_ERROR_ALLOCATION_FAILED;
```
dmitry-gurulev: Why we need delete nullptr?

media_driver/linux/common/ddi/media_ddi_render_target_table.cpp
vshampor: A queue-like structure looks more natural here; besides, this makes it easier to implement other surface eviction policies, if need be.

media_driver/linux/common/codec/ddi/media_ddi_decode_base.cpp
vshampor: Copy-pasted for consistency. Lines 77 and 78 have no effect, really, will remove.

media_driver/linux/common/ddi/media_ddi_render_target_table.cpp
vshampor: Indeed, will re-write this section.

vshampor:
An improved version of this PR can be found at #636.

PR #487, opened 2019-01-09
build: change FREE_KERNELS option to ENABLE_NONFREE_KERNELS

dvrogozh:
It was noted that FREE_KERNELS may be confusing since with =OFF it
actually enables non-free kernels (wihtout published sources). So,
let's use ENABLE_NONFREE_KERNELS instead.

Mind that ENABLE_NONFREE_KERNELS default value is ON. So, driver
default behaviour did not change.

Change-Id: Ifae417c3ada32d065fa615c9d5cc17c41f3b93b5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ script:
       intelmediadriver/media-driver-docker
       cmake
       -DMEDIA_VERSION="0.1.0"
-      -DFREE_KERNELS=ON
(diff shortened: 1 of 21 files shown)


dvrogozh:
@XinfengZhang : just small README update (fixed typo).

dvrogozh:
typo corrections in option comments

PR #490, opened 2019-01-11
media_ddi_encode_avc: fix the if-else logic and code style

fulinjie:
Fix the if-else logic in DdiEncodeAvc::ParseMiscParamQuantization(),
otherwise driver will report ASSERTMESSAGE even if the situation is
valid.

Refine the code stye.

Signed-off-by: Linjie Fu <linjie.fu@intel.com>


Sherry-Lin:
@fulinjie could you please rebase this patch? Thanks.

fulinjie:
@Sherry-Lin updated, thanks.

PR #492, opened 2019-01-11
Add X11 detect support

yangnin:


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -43,7 +43,7 @@
 #include "media_libva_util.h"
 #include "media_libva_decoder.h"
 #include "media_libva_encoder.h"
-#ifndef ANDROID
+#if !defined(ANDROID) || defined(X11_FOUND)
```
azhar904: Shouldn't the line number 35 above, also have the X11_FOUND flag?

azhar904:
Also change is needed in media_driver/linux/common/ddi/media_libva_common.h

struct DDI_MEDIA_CONTEXT

yangnin:
same with 496, close it

PR #494, opened 2019-01-15
Add X11 detect support

azhar904:
This is an updated version of the original PR #492

### DIFF
--- a/media_driver/cmake/linux/media_compile_flags_linux.cmake
+++ b/media_driver/cmake/linux/media_compile_flags_linux.cmake
@@ -176,6 +176,10 @@ set(MEDIA_COMPILER_FLAGS_DEBUG
     -DINSTR_GTUNE_EXT
 )
 
+if(X11_FOUND)
+    add_definitions(-DX11_FOUND)
+endif()
+
 include(${MEDIA_EXT_CMAKE}/ext/linux/media_compile_flags_linux_ext.cmake OPTIONAL)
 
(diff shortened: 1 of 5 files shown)


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -32,7 +32,7 @@
 #include <sys/ioctl.h>
 #include <unistd.h>
 
-#ifndef ANDROID
+#if !defined(ANDROID) && defined(X11_FOUND)
```
dvrogozh: Shouldn't we use just:
`#ifdef X11_FOUND`
and assure that build system don't set this for Android path? Smaller number of defines and code, the better.

media_driver/cmake/linux/media_compile_flags_linux.cmake
```diff
@@ -176,6 +176,10 @@ set(MEDIA_COMPILER_FLAGS_DEBUG
     -DINSTR_GTUNE_EXT
 )
 
+if(X11_FOUND)
+    add_definitions(-DX11_FOUND)
```
dvrogozh: To me passing -DHAVE_X11=1 seems better since it is aligned with what people are accustomed to with autotools. In this case Android should explicitly set -DHAVE_X11=0 and checks below should look like:
`#if HAVE_X11`

dvrogozh:
What if someone will want to build the driver _without_ X11? Are we going to provide a build option to disable X11 support?

PR #503, opened 2019-01-24
refact vp kernel open source build

yangnin:
Signed-off-by: Yang, Ning <ning.yang@intel.com>


media_driver/agnostic/common/vp/kernel_free/media_srcs.cmake
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) 2017, Intel Corporation
```
dvrogozh: 2019

PR #516, opened 2019-02-01
Add I420 to supported Image formats

azhar904:


wangyan42164:
I tested it on my platform. It is OK. Thanks.

PR #521, opened 2019-02-11
[Decode] Prevent mt read/write conflict

maksenox:
Two concurrent DdiMedia_SyncSurface function calls read and
write the same data when the second frame uses the first frame
as a reference frame.

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -3289,6 +3289,8 @@ static VAStatus DdiMedia_SyncSurface (
     PDDI_DECODE_CONTEXT decCtx = (PDDI_DECODE_CONTEXT)surface->pDecCtx;
     if (decCtx && surface->curCtxType == DDI_MEDIA_CONTEXT_TYPE_DECODER)
     {
+        DdiMediaUtil_LockGuard guard(&mediaCtx->SurfaceMutex);
+
         Codechal *codecHal = decCtx->pCodecHal;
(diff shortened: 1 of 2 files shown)


maksenox:
Two concurrent DdiMedia_SyncSurface function calls read and
write the same data when the second frame uses the first frame
as a reference frame.
Enqueue frames order:
  Frame #0
  Frame #1 (uses #0 as a reference)
Threads release order:
  Thread #1 (waits for Frame #1)
  Thread #0 (waits for Frame #0)
Resolution is: lock with the mutex the problem source code
(read and update StatusReport), so only one thread writes
the data, the second thread is waiting.

The problem is the second thread (second frame) starts to update StatusReport and checks all its dependencies too.
Then the first thread (first frame) appears and start to update its own StatusReport at this point both threads do something with StatusReport of the first frame. If they do it sequentially nothing bad happens. Resolution lock all StatusReport checks with the mutex.

media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -3289,6 +3289,22 @@ static VAStatus DdiMedia_SyncSurface (
...
+        public:
+            MutexGuard(PMEDIA_MUTEX_T pMutex):m_pMutex(pMutex)
+            {
+                DdiMediaUtil_LockMutex(m_pMutex);
+            }
+            ~MutexGuard()
+            {
+                DdiMediaUtil_UnLockMutex(m_pMutex);
+            }
+        };
```
dvrogozh: Shouldn't that be somewhere in utilities?

maksenox: Fixed

PR #522, opened 2019-02-12
Don't create the GEM buffer manager for invalid device

xhaihao:
In a dual GPU system, it is possible to open a non-Intel GPU device node then try to use iHD driver.  See https://gitlab.freedesktop.org/gstreamer/gstreamer-vaapi/issues/135 for details.


uartie:
@XinfengZhang can we merge this patch?

XinfengZhang:
LGTM, will merge it after test.

uartie:
@XinfengZhang any update??

PR #526, opened 2019-02-15
Fix DeriveImage with UYVY fourcc

xhaihao:
This makes the gst pipeline below works:

gst-launch-1.0 videotestsrc num-buffers=1 ! msdkvpp ! \
video/x-raw\(memory:DMABuf\),format=UYVY ! filesink location=a.yuv

Signed-off-by: Haihao Xiang <haihao.xiang@intel.com>


xhaihao:
You need the latest msdk and https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/merge_requests/197 to verify the pipeline mentioned in https://github.com/intel/media-driver/pull/526#issue-253294212

PR #535, opened 2019-02-24
vp8_decode: Ensure the Partition0BitOffset in VP8_BSD_OBJECT

sreerenjb:
Partition0 FirstMBBitOffset from Frame Header allows HW
to jump to the location in the bitstream where per MB information starts in the Partition0


XinfengZhang:
@Jexu looks we dont use these 3 bits , and from spec , it should be 
Partition0 FirstMBBitOffset from Frame Header, could you help to check it?

Jexu:
Yeah, we never use this field before. @sreerenjb I may wonder what case that requires this bit and can you share your bitstream to us to take a look?

PR #539, opened 2019-02-27
Allow I420 for surface creation

azhar904:
Signed-off-by: Yu Kang Ku <yu.kang.ku@intel.com>


XinfengZhang:
we expect application create surface with RT_FORMAT combined with FOURCC, so for I420,  application should create surfaces with VA_RT_FORMAT_YUV420 & an Surface Attributes , type is VASurfaceAttribPixelFormat, value is VA_FOURCC_I420. but to make it more flexible,  we should accept somebody to create surface by FOURCC.  will merge it after test.

PR #541, opened 2019-03-01
Fix PutImage take into account copied rectangles

maksenox:
Use memcpy loop over image rows instead of creating new
temporary surface and run VPP
when we don't need to do the image resize it means
source and destination rectangle are the same.


XinfengZhang:
@maksenox, why not call VPP?

maksenox:
The change already merged.
[Decode] Fix sigabort in DdiMedia_PutImage  
Commit ID 245a566c82033d6870dcd0b11614405d298e2111

PR #544, opened 2019-03-01
Fix SurfaceFromID get rid of false alarm

maksenox:
Ignore special surfaceID value VA_INVALID_SURFACE.
GetSurfaceFromVASurfaceID may be called
for non existing reference frame,
e.g., for Intra-only frame with surfaceID = -1.
The caller validates the surface again,
fix non-existing reference,
and decoding proceeds without any problem.
But the driver reports about critical error constantly.


XinfengZhang:
LGTM

PR #546, opened 2019-03-04
Frame split

zxye:
Enable frame split on ICL.


media_driver/linux/common/os/hwinfo_linux.c
```diff
@@ -133,6 +133,21 @@ MOS_STATUS HWInfo_GetGfxInfo(int32_t           fd,
...
+        unsigned int nengine = MAX_ENGINE_INSTANCE_NUM;
+        struct i915_engine_class_instance uengines[MAX_ENGINE_INSTANCE_NUM];
+        if (mos_query_engines(fd,I915_ENGINE_CLASS_VIDEO,0,&nengine,uengines) == 0)
+        {
+            gtSystemInfo->VDBoxInfo.NumberOfVDBoxEnabled = nengine;
+        }
+        else
+        {
+            //enable 1 vdbox by default
+            gtSystemInfo->VDBoxInfo.NumberOfVDBoxEnabled = 1;
```
tursulin: Why this? It sounds more logical to trust the query - but if query failed then other places in the code seem to trigger failure rather faking data? It could be some code base quirk I do not understand of course...

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4934,3 +5073,106 @@ int mos_set_context_param_load_balance(struct mos_linux_context *ctx,
...
+{
+    int ret = 0;
+    struct {
+        struct i915_user_extension base;
+        __u16 engine_index;
+        __u16 num_siblings;
+        __u32 flags;
+        __u64 mbz64;
+        struct i915_engine_class_instance engines[1];
+    } __attribute__((packed)) balancer;
```
tursulin: You could also use the I915_DEFINE_CONTEXT_ENGINES_LOAD_BALANCE macro from i915_drm.h so you don't need to duplicate the structure definition:

I915_DEFINE_CONTEXT_ENGINES_LOAD_BALANCE(balancer, 1);

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4934,3 +5073,106 @@ int mos_set_context_param_load_balance(struct mos_linux_context *ctx,
...
+    struct engines {
+        uint64_t extension;
+        struct i915_engine_class_instance class_instance[2];
+    } engines;
+    engines.extension = (uintptr_t)(&balancer);
+    engines.class_instance[0].engine_class = I915_ENGINE_CLASS_INVALID;
+    engines.class_instance[0].engine_instance = I915_ENGINE_CLASS_INVALID_NONE;
+    engines.class_instance[1].engine_class = I915_ENGINE_CLASS_VIDEO;
+    engines.class_instance[1].engine_instance = 0;
+
```
tursulin: Like in VENG, if you won't ever submit to index 1 in the map you don't need to configure it.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4934,3 +5073,106 @@ int mos_set_context_param_load_balance(struct mos_linux_context *ctx,
...
+    } __attribute__((packed)) balancer;
+
+
+    memclear(balancer);
+    balancer.base.next_extension = 0;
+    balancer.base.name = I915_CONTEXT_ENGINES_EXT_LOAD_BALANCE;
+    balancer.num_siblings = 1;
+    balancer.engines[0].engine_class = I915_ENGINE_CLASS_VIDEO;
+    balancer.engines[0].engine_instance = 0;
+
```
tursulin: So this is hardcoded for VCS0 + VCS2 configs and not driven by the query?

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4934,3 +5073,106 @@ int mos_set_context_param_load_balance(struct mos_linux_context *ctx,
...
+    struct {
+        struct i915_user_extension base;
+        struct i915_engine_class_instance master;
+        __u16 virtual_index;
+        __u16 num_bonds;
+        __u64 flags;
+        __u64 mbz64[4];
+        struct i915_engine_class_instance engines[1];
+    } __attribute__((packed))  bond;
```
tursulin: And for this you could use the I915_DEFINE_CONTEXT_ENGINES_BOND macro.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4934,3 +5073,106 @@ int mos_set_context_param_load_balance(struct mos_linux_context *ctx,
...
+    balancer.base.name = I915_CONTEXT_ENGINES_EXT_LOAD_BALANCE;
+    balancer.num_siblings = 1;
+    balancer.engines[0].engine_class = I915_ENGINE_CLASS_VIDEO;
+    balancer.engines[0].engine_instance = sib_idx;
+
+    struct engines {
+        uint64_t extension;
+        struct i915_engine_class_instance engines[2];
+    } engines;
+    engines.extension = (uintptr_t)(&balancer);
```
tursulin: Equally I915_DEFINE_CONTEXT_PARAM_ENGINES if you want.

media_driver/linux/common/os/mos_gpucontext_specific.cpp
```diff
@@ -275,10 +318,14 @@ void GpuContextSpecific::Clear()
     MOS_SafeFreeMemory(m_attachedResources);
     MOS_SafeFreeMemory(m_writeModeList);
     MOS_SafeFreeMemory(m_createOptionEnhanced);
-    if (m_i915Context)
+
+    for (int i=0; i<MAX_ENGINE_INSTANCE_NUM; i++)
     {
```
tursulin: Hmm.. is number of created context really relating to max number of engine instances or it is actually one to three? (single pipe, master, slave)

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: I don't want to use VLA as some compiler doesn't support VLA. But redefining is not good. I will rewrite it with malloc/free, similar to the  mos_set_context_param_load_balance() function.

media_driver/linux/common/os/hwinfo_linux.c
zxye: At the beginning, wanted to keep the code workable even when the GT query patch is not available.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: yes, fixed.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: For ICL, this is enough. It will be refined to cover more engine class and platforms.

media_driver/linux/common/os/mos_gpucontext_specific.cpp
zxye: For ICL, 3 is enough. But same as mos_set_context_param_bond_slave, it needs to be refined to cover more slaves. It should be one single pipe, one master, and MAX_ENGINE_INSTANCE_NUM-1 slaves: for (int i=0; i<=MAX_ENGINE_INSTANCE_NUM; i++) {...}

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: mos_set_context_param_bond_master() is redundent. It can reuse the mos_set_context_param_load_balance() function.
mos_set_context_param_bond_slave() is rewritten and renamed as mos_set_context_param_bond(). It takes one master and n bond class_instance as input now.

PR #548, opened 2019-03-05
Add X11_FOUND flag

azhar904:
There was an instance where the X11_FOUND flag was missing
causing build failures on systems which do not support X(eg. chrome).
Fix this by adding the check.


XinfengZhang:
LGTM

PR #552, opened 2019-03-12
Unbreak build without X11

jbeich:
Followup to #494. At least 18.4.1 doesn't build if libX11 isn't installed.


jbeich:
Sorry, I forgot to check existing PRs. Duplicate of #548.

PR #557, opened 2019-03-13
Increase the number of render target surfaces

maksenox:
VP9 decoder bigsurfpool test uses 180 surfaces and
127 slots are not enough for the test.


XinfengZhang:
@maksenox , it changes the frameidx? how about the resources related with the frames? such as ppRefList?

maksenox:
Close the pull request as test workaround only for VP9 decoder.

PR #567, opened 2019-03-19
Use sw swizzling on Gen11

dmitryermilov:
Use sw swizzling since gtt map is very slow on gen11.
The swizzling code has been borrowed from mesa Intel Surface Layout library

Signed-off-by: Dmitry Ermilov <dmitry.ermilov@intel.com>

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -1459,7 +1459,7 @@ VAStatus DdiMedia__Initialize (
     mediaCtx->m_auxTableMgr = AuxTableMgr::CreateAuxTableMgr(mediaCtx->pDrmBufMgr,
         &mediaCtx->SkuTable, &mediaCtx->WaTable);
 
-    mediaCtx->m_useSwSwizzling = MEDIA_IS_SKU(&mediaCtx->SkuTable, FtrSimulationMode);
+    mediaCtx->m_useSwSwizzling = MEDIA_IS_SKU(&mediaCtx->SkuTable, FtrSimulationMode) || MEDIA_IS_SKU(&mediaCtx->SkuTable, FtrUseSwSwizzling);
     mediaCtx->m_tileYFlag      = MEDIA_IS_SKU(&mediaCtx->SkuTable, FtrTileY);
     mediaCtx->modularizedGpuCtxEnabled = true;
 
--- a/media_driver/linux/common/ddi/media_libva_util.cpp
+++ b/media_driver/linux/common/ddi/media_libva_util.cpp
@@ -38,6 +38,7 @@
 #include "media_libva_util.h"
 #include "mos_utilities.h"
 #include "mos_os.h"
+#include "mos_sw_swizzle.h"
 #include "hwinfo_linux.h"
 #include "media_ddi_decode_base.h"
 #include "media_ddi_encode_base.h"
@@ -881,9 +882,13 @@ void* DdiMediaUtil_LockSurface(DDI_MEDIA_SURFACE  *surface, uint32_t flag)
                     DDI_CHK_CONDITION((surface->TileType != I915_TILING_Y), "Unsupported tile type", nullptr);
                     DDI_CHK_CONDITION((surfSize <= 0 || surface->iPitch <= 0), "Invalid surface size or pitch", nullptr);
 
-                    Mos_SwizzleData((uint8_t*)surface->bo->virt, (uint8_t*)surface->pSystemShadow,
-                                    MOS_TILE_Y, MOS_TILE_LINEAR,
-                                    static_cast<int32_t>(surfSize / surface->iPitch), surface->iPitch, flags);
+                    mos_memcpy_tiled_to_linear(0, surface->iPitch,
+                                            0, surfSize / surface->iPitch,
+                                            (char*)surface->pSystemShadow, (char*)surface->bo->virt,
+                                            surface->iPitch, surface->iPitch,
+                                            false,
+                                            surface->TileType,
+                                            MOS_MEMCPY_STREAMING_LOAD);
                 }
             }
             else if (flag & MOS_LOCKFLAG_WRITEONLY)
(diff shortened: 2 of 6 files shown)


media_driver/agnostic/common/os/mos_utilities.c
```diff
@@ -6658,3 +6658,1008 @@ void Mos_SwizzleData(
         }
     }
 }
+
+//*********This is ISL (Intel Surface Layout) implementation for sw-based tiling/detiling*********
+
+#define INLINE_SSE41
```
dvrogozh: I thought on mesa this was in a separate file. Can this be done the same way here? It will be easier to compare against mesa, hence support.

FurongZhang:
Hello, 

Starting from SKL, VEBOX(Video processing Fixed function) supports this VEBOX_TILING_CONVERT which is a high speed by pass mode(8 pixels per clock) and supports tile conversion(TileY/TileX->Linear). In driver, we supported this mode from LKF. If this is a must-to-have and will benefit the pipeline performance, we could add that into driver as an improvement. But please note, this VEBOX_TILING_CONVERT only supports tiling mode and pitch change for one pass.

Thanks,
Furong

dmitryermilov:
Hi Furong,

what about Linear->TileY/TileX, is it support by VEBOX? Does it support REXT color formats?
If so, can you please make at least engineering version which will expose this in vaMapBuffer/vaUnmapBuffer so that we can estimate performance ?

FurongZhang:
Hi, what "REXT color formats" means?

dmitryermilov:
I meant range extensions: 10,12,16 bits formats

media_driver/agnostic/common/os/mos_sw_swizzle.c
```diff
@@ -0,0 +1,1029 @@
...
+//!
+
+#include <string.h>
+#include <immintrin.h>
+#include "mos_sw_swizzle.h"
+
+#define INLINE_SSE41
+#define USE_SSE41
+
+typedef void *(*isl_mem_copy_fn)(void *dest, const void *src, size_t n);
```
dvrogozh: Looks like Mesa has 3 files with memcpy:
* https://github.com/mesa3d/mesa/blob/master/src/intel/isl/isl_tiled_memcpy.c
* https://github.com/mesa3d/mesa/blob/master/src/intel/isl/isl_tiled_memcpy_normal.c
* https://github.com/mesa3d/mesa/blob/master/src/intel/isl/isl_tiled_memcpy_sse41.c

Which one you brought and why you did not brought others (if you did not)?

We need to make sure that this memcpy will work on all platforms we support. For example, check that we don't use sse4.2... And consider to have 2 paths: one for sse4.2, another for sse4.1...

media_driver/agnostic/common/os/mos_sw_swizzle.c
```diff
@@ -0,0 +1,1029 @@
+/*
+* Copyright (c) 2019, Intel Corporation
```
dvrogozh: If you copy from Mesa, you might check how you should credit them via license. Looks like you can copy the code, but copyright notice should be preserved. Here you miss "Copyright 2013 Google". You might also credit Intel with the correct years range since originally it is "Copyright 2012 Intel Corporation".

media_driver/agnostic/common/os/mos_sw_swizzle.c
```diff
@@ -0,0 +1,1029 @@
...
+*
+* Permission is hereby granted, free of charge, to any person obtaining a
+* copy of this software and associated documentation files (the "Software"),
+* to deal in the Software without restriction, including without limitation
+* the rights to use, copy, modify, merge, publish, distribute, sublicense,
+* and/or sell copies of the Software, and to permit persons to whom the
+* Software is furnished to do so, subject to the following conditions:
+*
+* The above copyright notice and this permission notice shall be included
+* in all copies or substantial portions of the Software.
```
dvrogozh: License looks similar between media-driver and mesa, but there are differences. For example, media driver miss "(including the next paragraph)" words. Please, check what you should do in this case.

media_driver/agnostic/common/os/mos_sw_swizzle.c
```diff
@@ -0,0 +1,1029 @@
...
+#    define ALWAYS_INLINE inline __attribute__((always_inline))
+#  elif defined(_MSC_VER)
+#    define ALWAYS_INLINE __forceinline
+#  else
+#    define ALWAYS_INLINE inline
+#  endif
+#endif
+
+/** Minimum of two values: */
+#define MIN2( A, B )   ( (A)<(B) ? (A) : (B) )
```
dvrogozh: media driver is mostly written in C++, but this file is C and you use some MOS_ASSERT stuff below. We potentially may step into problems if these macro defs will start using C++. Please, align with code maintainers whether you should use C or C++. 

If you will decide to use C++, consider use std::min/max and other primitives.

media_driver/agnostic/common/os/media_srcs.cmake
```diff
@@ -29,6 +29,7 @@ set(TMP_SOURCES_
     ${CMAKE_CURRENT_LIST_DIR}/mos_gpucontextmgr.cpp
     ${CMAKE_CURRENT_LIST_DIR}/mos_cmdbufmgr.cpp
     ${CMAKE_CURRENT_LIST_DIR}/mos_commandbuffer.cpp
+    ${CMAKE_CURRENT_LIST_DIR}/mos_sw_swizzle.c
```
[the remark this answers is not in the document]
dmitryermilov: I put it in agnostic part because it's pure sw code without linux/windows specific.

media_driver/agnostic/common/os/mos_sw_swizzle.c
dmitryermilov: isl_tiled_memcpy_normal.c and isl_tiled_memcpy_sse41.c are just wrappers. real implementation is in isl_tiled_memcpy.c

media_driver/agnostic/common/os/mos_sw_swizzle.h
```diff
@@ -0,0 +1,86 @@
...
+//!
+//! WARNING: These values differ from the hardware enum values, which are
+//! unstable across hardware generations.
+//!
+//! Note that legacy Y tiling is ISL_TILING_Y0 instead of ISL_TILING_Y, to
+//! clearly distinguish it from Yf and Ys.
+//!
+enum isl_tiling
+{
+   ISL_TILING_LINEAR = 0,
```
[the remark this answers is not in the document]
dmitryermilov: ok

media_driver/agnostic/common/os/mos_sw_swizzle.c
dmitryermilov: @XinfengZhang , please say what is preferable

media_driver/agnostic/common/os/mos_sw_swizzle.c
dmitryermilov: ok

XinfengZhang:
hi @dmitryermilov , what's the benefit comparing with  Mos_SwizzleData? perf?  do you want to replace all of Mos_SwizzleData?

dmitryermilov:
@XinfengZhang , it gives ~5X performance in comparison to default Mos_SwizzleData. Mos_SwizzleData doesn't have optimizations, it makes per-pixel copy.

>>do you want to replace all of Mos_SwizzleData?
Ideally, yes. It's better to have one function. But let's do it step-by-step.

media_driver/agnostic/common/os/mos_sw_swizzle.c
XinfengZhang: c++ is preferable

PR #584, opened 2019-04-02
ddi/media_libva: Distinguish VA memory type from external buffer flag

xhaihao:
And allocate tiling memory for VA_SURFACE_ATTRIB_MEM_TYPE_VA if required
from user application/library


xhaihao:
Any comment/concern on this PR?

XinfengZhang:
LGTM, marked as "verifying" will start test.

PR #575, opened 2019-03-25
Map VA_FOURCC_A2R10G10B10 to Media_Format_B10G10R10A2

xhaihao:

PR #585, opened 2019-04-02
Update libva build and add B10G10R10A2 format

Sherry-Lin:
Include https://github.com/intel/media-driver/pull/575
Add libva path when build driver


Sherry-Lin:
No build issue after set libva path

PR #586, opened 2019-04-02
Travis: add libva path in build

Sherry-Lin:
Add DLIBVA_DRIVERS_PATH and DLIBVA_INSTALL_PATH to make sure driver in
docker image use the correct libva lib.


yangnin:
OK for me

PR #589, opened 2019-04-09
cmake: check libva-x11 instead of X11

xhaihao:
This fixes the error below when X11 is not available

In file included from
/home/xhh/graphics/ex/vaapi/media-driver/media_driver/linux/common/ddi/media_libva.cpp:47:
/home/xhh/graphics/ex/vaapi/media-driver/media_driver/linux/common/ddi/media_libva_putsurface_linux.h:31:10:
fatal error: va/va_dricommon.h: No such file or directory
 #include <va/va_dricommon.h>

### DIFF
--- a/media_driver/media_top_cmake.cmake
+++ b/media_driver/media_top_cmake.cmake
@@ -21,7 +21,7 @@
 project( media )
 
 find_package(PkgConfig)
-find_package(X11)
+find_package(libva-x11)
 
 bs_set_if_undefined(LIB_NAME iHD_drv_video)
 


XinfengZhang:
is it possible that there are libva-x11 on a machine , but no X11?

xhaihao:
find_package(libva-x11) doesn't work due to lack of Findlibva-x11.cmake

xhaihao:
Closed this PR due to lack of Findlibva-x11.cmake

PR #592, opened 2019-04-11
reset BRC when the target bitrate or max bitrate changed

XinfengZhang:


rosetta-jpn:
Hi, may I ask you to check in this change?

Thanks!

XinfengZhang:
@rosetta-jpn sure, will check in it after validation

PR #555, opened 2019-03-13
Fix ayuv

xhaihao:
Some fixes for ayuv image/surface
* support derived ayuv image
* more RT formats for VAProfileNone
* AYUV and Y410 images have 32bits

PR #600, opened 2019-04-16
add Y210/AYUV/Y410 format support in vaDeriveImage.

feiwan1:
Signed-off-by: Wangfei <fei.w.wang@intel.com>

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -4227,6 +4227,7 @@ VAStatus DdiMedia_DeriveImage (
         break;
     case Media_Format_Y410:
     case Media_Format_AYUV:
+    case Media_Format_Y210:
         vaimg->format.bits_per_pixel    = 32;
         vaimg->data_size                = mediaSurface->iPitch * mediaSurface->iHeight;
         vaimg->num_planes               = 1;


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -4220,6 +4220,9 @@ VAStatus DdiMedia_DeriveImage (
         break;
     case Media_Format_X8R8G8B8:
     case Media_Format_X8B8G8R8:
+    case Media_Format_Y210:
```
FurongZhang: According to my understanding, Y210 is not 32 bits per pixel.

feiwan1: @FurongZhang , In my understanding, Y210 is ordered as YUYV..., each YUYV stands for 2 pixel, and each channel has 16bit. so the per_pixel = 4*16/2 =32.

wangyan-intel:
@feiwan1 Is it duplicated with https://github.com/intel/media-driver/pull/555? If yes, please close it.

feiwan1:
Thanks @wangyan-intel for your information. #555 only covered AYUV format in DeriveImage. So I will modify my patch to support Y210&Y410.

wangyan-intel:
#601 #602 #603 are closed. So I add verifying flag for starting merging process. Thanks.

PR #604, opened 2019-04-16
ignore error status value for jpeg decode

XinfengZhang:
no clear description of the meaning of the value for jpeg
just the description for AVC MPEG2 ... so, ignore it

### DIFF
--- a/media_driver/agnostic/common/codec/hal/codechal_decoder.cpp
+++ b/media_driver/agnostic/common/codec/hal/codechal_decoder.cpp
@@ -1603,14 +1603,14 @@ MOS_STATUS CodechalDecode::GetStatusReport(
                     else
                     {
                         // Check to see if decoding error occurs
-                        if ((m_decodeStatusBuf.m_decodeStatus[i].m_mmioErrorStatusReg &
-                             m_mfxInterface->GetMfxErrorFlagsMask()) != 0)
-                        {
-                            codecStatus[j].m_codecStatus = CODECHAL_STATUS_ERROR;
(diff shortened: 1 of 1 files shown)


media_driver/agnostic/common/codec/hal/codechal_decoder.cpp
```diff
@@ -1603,14 +1603,14 @@ MOS_STATUS CodechalDecode::GetStatusReport(
...
                         // Check to see if decoding error occurs
-                        if ((m_decodeStatusBuf.m_decodeStatus[i].m_mmioErrorStatusReg &
-                             m_mfxInterface->GetMfxErrorFlagsMask()) != 0)
-                        {
-                            codecStatus[j].m_codecStatus = CODECHAL_STATUS_ERROR;
-                        }
-                        //MB Count bit[15:0] is error concealment MB count for none JPEG decoder.
                         if (m_standard != CODECHAL_JPEG)
                         {
+                            if ((m_decodeStatusBuf.m_decodeStatus[i].m_mmioErrorStatusReg &
```
ss22ongithub: This check should apply only for AVC, VC1 and MPEG2 standards. This change checks VP8,VP9 standards  which may not be  required.

XinfengZhang: @ss22ongithub yes, you are correct, current change just avoid check for  JPEG. and keep old logic. it is just a bug fix. will change to only check AVC, VC1 MPEG2 later .

PR #623, opened 2019-05-15
Use sw swizzling on Gen11

dmitryermilov:
Use sw swizzling since gtt map is very slow on gen11.

Signed-off-by: Dmitry Ermilov <dmitry.ermilov@intel.com>

### DIFF
--- a/media_driver/linux/common/ddi/media_libva_util.cpp
+++ b/media_driver/linux/common/ddi/media_libva_util.cpp
@@ -841,6 +841,8 @@ VAStatus DdiMediaUtil_CreateBuffer(DDI_MEDIA_BUFFER *buffer, MOS_BUFMGR *bufmgr)
     return hr;
 }
 
+VAStatus SwizzleSurface(PDDI_MEDIA_CONTEXT mediaCtx, PGMM_RESOURCE_INFO pGmmResInfo, void *pLockedAddr, uint32_t TileType, uint8_t* pResourceBase, bool bUpload);
+
 // add thread protection for multiple thread?
 void* DdiMediaUtil_LockSurface(DDI_MEDIA_SURFACE  *surface, uint32_t flag)
 {
@@ -874,7 +876,7 @@ void* DdiMediaUtil_LockSurface(DDI_MEDIA_SURFACE  *surface, uint32_t flag)
             {
                 mos_bo_map(surface->bo, flag & MOS_LOCKFLAG_WRITEONLY);
             }
-            else if (surface->pMediaCtx->m_useSwSwizzling)
+            else if ((surface->pMediaCtx->m_useSwSwizzling) && !(flag & MOS_LOCKFLAG_NO_SWIZZLE))
             {
                 mos_bo_map(surface->bo, flag & MOS_LOCKFLAG_WRITEONLY);
                 if (surface->pSystemShadow == nullptr)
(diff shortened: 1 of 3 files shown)


media_driver/linux/common/ddi/media_libva_util.cpp
```diff
@@ -874,34 +876,36 @@ void* DdiMediaUtil_LockSurface(DDI_MEDIA_SURFACE  *surface, uint32_t flag)
...
+                                                   (uint8_t *)surface->pSystemShadow,
+                                                   false);
+                DDI_CHK_CONDITION((vaStatus != VA_STATUS_SUCCESS), "SwizzleSurface failed", nullptr);
+
             }
             else if (flag & MOS_LOCKFLAG_NO_SWIZZLE)
             {
                 mos_bo_map(surface->bo, flag & MOS_LOCKFLAG_READONLY);
             }
+            else if (flag & MOS_LOCKFLAG_WRITEONLY)
```
dvrogozh: did you have a particular reason to move 'if (flag & MOS_LOCKFLAG_WRITEONLY)' down?

media_driver/linux/common/ddi/media_libva_util.cpp
```diff
@@ -874,34 +876,36 @@ void* DdiMediaUtil_LockSurface(DDI_MEDIA_SURFACE  *surface, uint32_t flag)
...
-            }
-            else if (flag & MOS_LOCKFLAG_WRITEONLY)
-            {
-                mos_gem_bo_map_gtt(surface->bo);
+
+                uint64_t surfSize = surface->pGmmResourceInfo->GetSizeMainSurface();
+                DDI_CHK_CONDITION((surface->TileType != I915_TILING_Y), "Unsupported tile type", nullptr);
+                DDI_CHK_CONDITION((surfSize <= 0 || surface->iPitch <= 0), "Invalid surface size or pitch", nullptr);
+
+                VAStatus vaStatus = SwizzleSurface(surface->pMediaCtx,
```
dvrogozh: There are few other places where we use Mos_SwizzleData. Did you review them? Do we need them? Can we replace them with SwizzleSurface?

media_driver/linux/common/ddi/media_libva_util.cpp
dmitryermilov: To keep two blocks of sw swizzling related code closer to each ether

media_driver/linux/common/ddi/media_libva_util.cpp
dmitryermilov: Yes, I know. But I intentionally didn't touch them in this commit since: 
1) This change (as it's said in commit message) aims to enable SW swizzling on Gen11 , nothing more. 
2) Even current commit with minimal changes can causes (and did in the previous version) problems. With more changes in one commit, it will be harder to identify a RC.

media_driver/linux/common/ddi/media_libva_util.cpp
```diff
@@ -874,34 +876,36 @@ void* DdiMediaUtil_LockSurface(DDI_MEDIA_SURFACE  *surface, uint32_t flag)
             {
                 mos_bo_map(surface->bo, flag & MOS_LOCKFLAG_WRITEONLY);
             }
-            else if (surface->pMediaCtx->m_useSwSwizzling)
+            else if ((surface->pMediaCtx->m_useSwSwizzling) && !(flag & MOS_LOCKFLAG_NO_SWIZZLE))
```
Xiaogangli-intel: MOS_LOCKFLAG_NO_SWIZZLE flag is meaningful only when m_useSwizzling is true, right? So I think we should move the conditional statements about MOS_LOCKFLAG_NO_SWIZZLE into 'else if ((surface->pMediaCtx->m_useSwSwizzling)'.

dmitryermilov: > OS_LOCKFLAG_NO_SWIZZLE flag is meaningful only when m_useSwizzling is true, right? 
No, please see line #901

PR #626, opened 2019-05-20
[vp] fix surface sync issue

djie1:
fix the issue when there is implicit VPP process by vaGetImage(), vaSyncSurface() gets ERROR_OPERATION_FAILED.


XinfengZhang:
@FurongZhang  hi furong, please help to check the PR

PR #632, opened 2019-05-23
[Encode] Refactor encode quality factor

semelianova:
ICQ_quality_factor has been removed and replaced by more common quality_factor variable. Also removed prefix ICQ from other quality factor variables and constants. Added QualityFactor checking in QVBR case,
it was not in AVC.


feiwan1:
The patch looks good. If this patch could be merge, then we don't need pull #684 .
BTW, we need another patch to remove ICQ_quality_factor in libva.

semelianova:
> The patch looks good. If this patch could be merge, then we don't need pull #684 .
> BTW, we need another patch to remove ICQ_quality_factor in libva.

Yes, you are right.  ICQ_quality_factor should be replaced with  quality_factor in libva too. 

> this value will be deprecated in future, to use quality_factor instead of it.
[va.h](https://github.com/intel/libva/blob/06f023e84b407a2fe429191aa081f3cc81bb6445/va/va.h#L2020)

PR #633, opened 2019-05-23
Enable supporting p010 output on BXT

semelianova:
BXT support P010 output. Decoded hevce 10b output sreams from BXT and CFL
are bit2bit.


XinfengZhang:
LGTM

PR #636, opened 2019-06-04
Implement render target table to track SurfaceId and indices association

maksenox:
The pull request contains two commits the first is the pull request #486 and the second contains the improvements.
I believe the two commits should be squashed and commit the result.

Changes compared to the pull request #486
- get rid of surface state RT_STATE_INACTIVE ACTIVE_IN_LASTFRAME ACTIVE_IN_CURFRAME
- get rid of SetRTState and  ReleaseDPBRenderTargets

Media SDK usage example in async = 5

vaBeginPicture

// some surfaces participate in the decoding operation
// either as current, recon, reference or in loop filtering.
// all these surfaces must be marked by RegisterRTSurface call
// there is no surface state anymore.

vaEndPicture

vaBegin/vaEnd executed five times.
And now several threads wait for vaSyncSurface, all the time surfaces and their indices must be valid.
So we can't call ReleaseDPB function in vaEndPicture.

Whenever RegisterRTSurface is called the surface id is moved to the front of the useageOrder list.
recorded SurfaceId to index association, 
if it is known surface then reuse its index, 
if it is the new surface id then the first free index is used,
if there is no free index then remove the oldest SurfaceId and use its index.

E.g., VP9 decoder has 127 slots for indices.
One frame decoding requires eight reference frames and one current frame.
Every next frame may increase that number by one.
So in async=5 decoder requires ~13 frames.
When we exceed 127 slots it means 127 different surfaces were used and the decoder processed ~100 frames.
And we can reuse the index of the oldest used SurfaceId.


media_driver/linux/common/codec/ddi/media_ddi_decode_base.cpp
```diff
@@ -370,6 +379,9 @@ void DdiMediaDecode::DestroyContext(VADriverContextP ctx)
     MOS_FreeMemory(m_ddiDecodeCtx->DecodeParams.m_subsetParams);
     m_ddiDecodeCtx->DecodeParams.m_sliceParams = nullptr;
 
+    // ?! MOS_FreeMemory(m_ddiDecodeCtx) missing?
```
Losev-Ivan: @DenWolf could you, please, look at this

media_driver/linux/common/ddi/media_ddi_render_target_table.cpp
```diff
@@ -0,0 +1,293 @@
...
+    if (id == VA_INVALID_ID)
+    {
+        return VA_STATUS_ERROR_INVALID_PARAMETER;
+    }
+
+    if (!IsRegistered(id))
+    {
+        if (m_freeIndices.empty())
+        {
+            if (UnregisterRTSurface(m_usageOrder.back()) != VA_STATUS_SUCCESS)
```
Losev-Ivan: Suggest to collapse check for empty and unregister into one if condition

media_driver/linux/common/ddi/media_ddi_render_target_table.cpp
```diff
@@ -0,0 +1,293 @@
...
+//!
+//! \param    [in] id
+//!           VASurfaceID of the registered render target to be set as "current"
+//!
+//! \return   VAStatus
+//!           VA_STATUS_SUCCESS if successful, VA_STATUS_ERROR_INVALID_PARAMETER if
+//!           no such surface is registered
+VAStatus MediaDdiRenderTargetTable::SetCurrentRTSurface(VASurfaceID id)
+{
+    if (RegisterRTSurface(id) != VA_STATUS_SUCCESS)
```
Losev-Ivan: Suggest to use DDI_CHK_STATUS

media_driver/linux/common/ddi/media_ddi_render_target_table.cpp
```diff
@@ -0,0 +1,293 @@
...
+//!
+//! \param    [in] id
+//!           VASurfaceID of the registered render target to be set as containing the reconstructed frame
+//!
+//! \return   VAStatus
+//!           VA_STATUS_SUCCESS if successful, VA_STATUS_ERROR_INVALID_PARAMETER if
+//!           no such surface is registered
+VAStatus MediaDdiRenderTargetTable::SetCurrentReconTarget(VASurfaceID id)
+{
+    if (RegisterRTSurface(id) != VA_STATUS_SUCCESS)
```
Losev-Ivan: Suggest to use DDI_CHK_STATUS

media_driver/linux/common/ddi/media_ddi_render_target_table.h
```diff
@@ -0,0 +1,78 @@
...
+
+protected:
+    VASurfaceID m_currentRtSurface;
+    VASurfaceID m_currentReconTarget;
+    std::map<VASurfaceID, RTTableIdx> m_surfaceIdToIndexMap;
+    std::vector<RTTableIdx> m_freeIndices;
+    std::list<VASurfaceID> m_usageOrder;
+};
+
+#endif /* _MEDIA_DDI_RENDER_TARGET_TABLE_H_ */
```
Losev-Ivan: Please, add empty line

media_driver/linux/common/ddi/media_ddi_render_target_table.cpp
```diff
@@ -0,0 +1,293 @@
...
+    auto it = std::find_if(m_surfaceIdToIndexMap.begin(), m_surfaceIdToIndexMap.end(),
+        [=](const RTPair& pair){return pair.second == FrameIdx;});
+
+    if (it != m_surfaceIdToIndexMap.end())
+    {
+        id = it->first;
+    }
+
+    return id;
+}
```
Losev-Ivan: Please, add empty line

PR #641, opened 2019-06-11
Improvement: Adding scaling configuration according to processing mode.

FurongZhang:
The driver pipeline will be optimized to fix function SFC if power saving mode, media kernel path if performance mode.

### DIFF
--- a/media_driver/linux/common/vp/ddi/media_libva_vp.c
+++ b/media_driver/linux/common/vp/ddi/media_libva_vp.c
@@ -849,6 +849,7 @@ DdiVp_SetProcPipelineParams(
 
     VP_DDI_FUNCTION_ENTER;
     DDI_CHK_NULL(pVaDrvCtx, "Null pVaDrvCtx.", VA_STATUS_ERROR_INVALID_CONTEXT);
+    DDI_CHK_NULL(pPipelineParam, "Null pPipelineParam Buffer.", VA_STATUS_ERROR_INVALID_BUFFER);
 
     // initialize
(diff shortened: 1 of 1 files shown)


media_driver/linux/common/vp/ddi/media_libva_vp.c
```diff
@@ -1142,6 +1143,27 @@ DdiVp_SetProcPipelineParams(
...
+    switch (pPipelineParam->processing_mode)
+    {
+    case VAProcPowerSavingMode:
+        if (pVpHalSrcSurf->SurfType == SURF_IN_PRIMARY)
+        {
+            pVpHalSrcSurf->ScalingMode       = VPHAL_SCALING_AVS;
+            pVpHalSrcSurf->ScalingPreference = VPHAL_SCALING_PREFER_SFC;
+        }
+        break;
+    case VAProcPerformanceMode:
```
lizhong1008: I doubt AVS can show better performance.  Probably using VAProcDefaultMode is better here? Any other thought?

FurongZhang: Generally, AVS sampler has better performance for typical usage(such as scaling, CSC etc.). We have related performance data to prove that.

FurongZhang: And if the VAAPI user does not set processing mode, we used the default mode(value = 0) which makes it compatible with the previous one.

FurongZhang:
Longfei, can you merge it? Thanks  lot in advance!

PR #642, opened 2019-06-11
Enable supporting p010 output on SKL

semelianova:
Tested vpp with p010 output. Output streams are valid and they show good
metrics. All gen9 platforms support p010 output.


FurongZhang:
Looks good.

PR #646, opened 2019-06-13
Don't map gtt for linear surface.

FurongZhang:


media_driver/linux/common/os/mos_graphicsresource_specific.cpp
```diff
@@ -452,7 +452,14 @@ void* GraphicsResourceSpecific::Lock(OsContext* osContextPtr, LockParams& params
         {
             if (pOsContextSpecific->IsAtomSoc())
             {
-                mos_gem_bo_map_gtt(boPtr);
+                if (m_tileType != MOS_TILE_LINEAR ||params.m_uncached)
```
XinfengZhang: seems , it is different with your title description?

XinfengZhang:
@furong, now there are a full solution , right? is it duplicate one, or just need to update?

PR #661, opened 2019-06-25
Add A2RGB10 and A2BGR10 for the function of vaCreateImage.

FurongZhang:

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -3794,6 +3794,8 @@ VAStatus DdiMedia_CreateImage(
         case VA_FOURCC_I420:
         case VA_FOURCC_IYUV:
         case VA_FOURCC_UYVY:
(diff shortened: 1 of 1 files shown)


Jister:
Also need to add VA_FOURCC_A2R10G10B10/VA_FOURCC_A2B10G10R10in https://github.com/intel/media-driver/blob/f6eca7f75a845fdd7b894aa6794122284459c726/media_driver/linux/common/ddi/media_libva_caps.cpp#L3037

FurongZhang:
I have already added it in another PR. They filed two issues. So two different PRs.

Jister:
I see #651, what I mean is that we also need to add these two format in the function MediaLibvaCaps::ConvertFourccToGmmFmt, otherwise, DdiMedia_CreateImage will failed at the following line:
https://github.com/intel/media-driver/blob/f6eca7f75a845fdd7b894aa6794122284459c726/media_driver/linux/common/ddi/media_libva.cpp#L3823

Issue #281, opened 2018-08-17
[ffmpeg-vaapi][mjpeg enc]the quality of encoded file with option "global_quality" level 90 is lower than the level 50  


wangzj0601:
Test Env:
Platform: CFL/KBL/SKL
Arch: x86_64
Kernel version: CFL 4.15.0,    KBL  4.14.0-rc1 , SKL  4.15.0
Libva: 6ae717304c48e7710a76a87cef2a1096ff450af2
media_driver: 36b3a0a
libva-utils:  9a10ad663349732decd668a426cfc349b0010d1d
ffmpeg-vaapi: 5423fe29da6bbf28f5253800cace133ac8343a7e(master branch)

Test step:
1. ffmpeg -hwaccel vaapi -vaapi_device /dev/dri/renderD128 -y -s:v 1920x1080 -i /root/media_streams//encoderbitstreams//1080p_blue_sky.yuv -vf format=nv12,hwupload -c:v mjpeg_vaapi -global_quality 90 1080p_1920x1080_i420_gquality90_fps30.mp4
2. ffmpeg -i ./1080p_1920x1080_i420_gquality90_fps30.mp4 -pix_fmt yuv420p dump.yuv
3. metrics_calc_lite -i1 /root/media_streams//encoderbitstreams//1080p_blue_sky.yuv -i2 dump.yuv -w 1920 -h 1080 -nopfm psnr y
4. YSNR=' 12.41936', and play the encoded file, find that the quality is very low
5. ffmpeg -hwaccel vaapi -vaapi_device /dev/dri/renderD128 -y -s:v 1920x1080 -i /root/media_streams//encoderbitstreams//1080p_blue_sky.yuv -vf 'format=nv12,hwupload' -c:v mjpeg_vaapi -global_quality 50 1080p_1920x1080_i420_gquality50_fps30.mp4
7. YSNR=' 27.85109'

XinfengZhang:
we have a fix about the quality setting. https://github.com/intel/media-driver/pull/289/commits/e1345456a4a378ba19da80e47f77f2be5ad4e4d4

wangzj0601:
Does this commit has been merged into master branch? This issue still can be duplicated with the latest driver
commit d32d5ddc9daed9dd40a7fc22288a9ab3f1788d10
Author: xiong_(sean)__zhou <sean.zhou@intel.com>
Date:   Thu Sep 13 09:14:06 2018 +0800


global_quality | bitrate | psnr
[...]

PR #667, opened 2019-06-27
convert Qmatrix of Jpeg from Zigzag to Raster order

XinfengZhang:

### DIFF
--- a/media_driver/linux/common/codec/ddi/media_ddi_encode_jpeg.cpp
+++ b/media_driver/linux/common/codec/ddi/media_ddi_encode_jpeg.cpp
@@ -30,6 +30,18 @@
 #include "media_ddi_encode_const.h"
 #include "media_ddi_factory.h"
 
+static const uint32_t zigzag_order[64] =
+{
+    0,   1,  8, 16,  9,  2,  3, 10,
+    17, 24, 32, 25, 18, 11,  4,  5,
+    12, 19, 26, 33, 40, 48, 41, 34,
+    27, 20, 13,  6,  7, 14, 21, 28,
+    35, 42, 49, 56, 57, 50, 43, 36,
+    29, 22, 15, 23, 30, 37, 44, 51,
+    58, 59, 52, 45, 38, 31, 39, 46,
+    53, 60, 61, 54, 47, 55, 62, 63
+};
+
 extern template class MediaDdiFactoryNoArg<DdiEncodeBase>;
 static bool isEncodeJpegRegistered =
     MediaDdiFactoryNoArg<DdiEncodeBase>::RegisterCodec<DdiEncodeJpeg>(ENCODE_ID_JPEG);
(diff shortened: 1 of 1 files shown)


wangyan-intel:
LGTM.

XinfengZhang:
please help to verify #281 using this patch

XinfengZhang:
@dmitryermilov  could you help to check whether  MSDK also need to be changed?

dmitryermilov:
Thanks @XinfengZhang . I'm OOO this week. I'll forward to MSDK folks to analyze.

media_driver/linux/common/codec/ddi/media_ddi_encode_jpeg.cpp
```diff
@@ -703,7 +715,7 @@ VAStatus DdiEncodeJpeg::QualityScaleQmatrix()
                 quantValue = 255;
             }
 
-            quantMatrix->m_quantTable[qMatrixCount].m_qm[i] = (uint16_t)quantValue;
+            quantMatrix->m_quantTable[qMatrixCount].m_qm[zigzag_order[i]] = (uint16_t)quantValue;
```
Alexandr-Konovalov: Could you please clarify this a little? `m_qm` is modified in-place,
so, say, after iteration `i = 2` puts scaled value to `m_qm[8]`,  on iteration with `i = 8`
`quantValue` will be updated one, not original from an application. Is it intentional?

Alexandr-Konovalov:
@XinfengZhang  Media SDK is not affected, as for HW encode the default quantization matrix is from the driver, while a user-provided matrix is just passed by Media SDK to driver without modifications.

media_driver/linux/common/codec/ddi/media_ddi_encode_jpeg.cpp
XinfengZhang: hi @Alexandr-Konovalov , the matrix from application should be zigzag order, but HW need Raster order, I just convert it from ZigZag to Raster

XinfengZhang:
@Alexandr-Konovalov , how about application side? such as mfx_encoder etc..

Alexandr-Konovalov:
@XinfengZhang Those tools are not keeping matrices internally, so not affected.

media_driver/linux/common/codec/ddi/media_ddi_encode_jpeg.cpp
Alexandr-Konovalov: @XinfengZhang Thanks, I understand that. What I do not understand: is it OK to apply quality to the matrix in-place? Say, for `i = 8` the code puts to `quantMatrix->m_quantTable[qMatrixCount].m_qm[17]` quantValue that was scaled from output of iteration `i = 2`, not from value provided by an application.

dmitryermilov:
+ @alariono

media_driver/linux/common/codec/ddi/media_ddi_encode_jpeg.cpp
lakulako: You can't convert order inplace. Because matrix is read in one order and is written in another.

PR #681, opened 2019-07-12
Fix AYUV format mask issue.

FurongZhang:


XinfengZhang:
@FurongZhang  there are conflict, and I am confused why you remove the mask for AYUV?

PR #684, opened 2019-07-16
[Encode] use quality_factor under QVBR mode for h265 encode

feiwan1:
According the illustration of va.h, quality_factor will be used
under QVBR mode. ICQ_quality_factor is used only under ICQ mode.


XinfengZhang:
LGTM

PR #696, opened 2019-07-26
media_libva: Return P010 as a composite format

harishkrupo:
Exporting P010 buffers fails because we don't recognize it as a
composite object. Add DRM_FORMAT_P010 as a return value for
VA_FOURCC_P010.

Signed-off-by: Harish Krupo <harishkrupo@gmail.com>

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -5929,6 +5929,7 @@ static uint32_t DdiMedia_GetDrmFormatOfCompositeObject(uint32_t fourcc)
     case VA_FOURCC_Y800:
         return DRM_FORMAT_R8;
     case VA_FOURCC_P010:
+        return DRM_FORMAT_P010;
     case VA_FOURCC_I010:
         // These currently have no composite DRM format - they are usable
         // only as separate planes.


harishkrupo:
@FurongZhang This solves my problem of not being able to export P010 buffers. If this patch is incomplete, please do suggest the changes.

FurongZhang:
It is looking good.

media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -5929,6 +5929,7 @@ static uint32_t DdiMedia_GetDrmFormatOfCompositeObject(uint32_t fourcc)
     case VA_FOURCC_Y800:
         return DRM_FORMAT_R8;
     case VA_FOURCC_P010:
+	return DRM_FORMAT_P010;
```
Sherry-Lin: Could you please replace the Tab key with space? Thanks.

harishkrupo: Thanks, updated.

ValZapod:
Can this fix https://github.com/intel/intel-vaapi-driver/issues/398 if you will apply https://patchwork.ffmpeg.org/project/ffmpeg/patch/qv2PyTKpWVd6zJqsf3pwRO7gzci3X9OW3r2v3Kt9z1k@cp4-web-037.plabs.ch/?

PR #705, opened 2019-08-12
register VP8 decode reference frame

XinfengZhang:
the fix should be exist in ffmpeg
it is another solution. ffmpeg use the surface which is decoded by other
context to be reference. so need to register it


abhishek53k:
@XinfengZhang there is crash when playing vp8 in chromiumos, reverting this patch fix crash issue.
if we revert this patch what will be the impact ? from comments seems its specific to ffmpeg.

PR #708, opened 2019-08-13
check pitch from application for user ptr

XinfengZhang:

### DIFF
--- a/media_driver/linux/common/ddi/media_libva_util.cpp
+++ b/media_driver/linux/common/ddi/media_libva_util.cpp
@@ -354,6 +354,12 @@ VAStatus DdiMediaUtil_AllocateSurface(
                 gmmParams.BaseHeight = mediaSurface->pSurfDesc->uiSize / mediaSurface->pSurfDesc->uiPitches[0];
             }
         }
+        if(mediaSurface->pSurfDesc->uiVaMemType == VA_SURFACE_ATTRIB_MEM_TYPE_USER_PTR)
+        {
+            gmmParams.ExistingSysMemSize = mediaSurface->pSurfDesc->uiBuffserSize;
+            gmmParams.pExistingSysMem    = mediaSurface->pSurfDesc->ulBuffer;
+            gmmParams.Flags.Info.ExistingSysMem = true;
+        }
     }
     else
     {
@@ -405,6 +411,12 @@ VAStatus DdiMediaUtil_AllocateSurface(
     uint32_t    gmmPitch;
     uint32_t    gmmSize;
     uint32_t    gmmHeight;
+
+    gmmPitch    = (uint32_t)gmmResourceInfo->GetRenderPitch();
+    if( DdiMediaUtil_IsExternalSurface(mediaSurface) && ( mediaSurface->pSurfDesc->uiVaMemType == VA_SURFACE_ATTRIB_MEM_TYPE_USER_PTR ) && mediaSurface->pSurfDesc->uiPitches[0])
+    {
+        gmmResourceInfo->OverridePitch(mediaSurface->pSurfDesc->uiPitches[0]);
+    }
     gmmPitch    = (uint32_t)gmmResourceInfo->GetRenderPitch();
     gmmSize     = (uint32_t)gmmResourceInfo->GetSizeSurface();
     gmmHeight   = gmmResourceInfo->GetBaseHeight();


media_driver/linux/common/ddi/media_libva_util.cpp
```diff
@@ -405,6 +411,23 @@ VAStatus DdiMediaUtil_AllocateSurface(
...
+    {
+        uint32_t gmmHAlign;
+        gmmHAlign = gmmResourceInfo->GetHAlign();
+        if((gmmPitch < mediaSurface->pSurfDesc->uiPitches[0]) && (((mediaSurface->pSurfDesc->uiPitches[0]-gmmPitch)%gmmHAlign) == 0))
+        {
+            gmmResourceInfo->OverridePitch(mediaSurface->pSurfDesc->uiPitches[0]);
+        }
+        else
+        {
+           DDI_ASSERTMESSAGE("the pitch does not match the basic requirement.");
```
dvrogozh: This won't work for the case when user allocated pitch _matches_ the one from gmmlib:):). You need either use `<=` above and always call OverridePitch even if they match or explicitly avoid error with else `(gmmPitch != mediaSurface->pSurfDesc->uiPitches[0])`

media_driver/linux/common/ddi/media_libva_util.cpp
```diff
@@ -405,6 +411,23 @@ VAStatus DdiMediaUtil_AllocateSurface(
...
     uint32_t    gmmHeight;
+
+    gmmPitch    = (uint32_t)gmmResourceInfo->GetRenderPitch();
+    if( DdiMediaUtil_IsExternalSurface(mediaSurface) && ( mediaSurface->pSurfDesc->uiVaMemType == VA_SURFACE_ATTRIB_MEM_TYPE_USER_PTR ) )
+    {
+        uint32_t gmmHAlign;
+        gmmHAlign = gmmResourceInfo->GetHAlign();
+        if((gmmPitch < mediaSurface->pSurfDesc->uiPitches[0]) && (((mediaSurface->pSurfDesc->uiPitches[0]-gmmPitch)%gmmHAlign) == 0))
+        {
+            gmmResourceInfo->OverridePitch(mediaSurface->pSurfDesc->uiPitches[0]);
```
dvrogozh: I still don't understand why we need to call OverridePitch? If we notified gmmlib that memory is user allocated, from my perspective gmmlib should return user allocated pitch right away... So, I am not sure that this entire code block is actually needed.

media_driver/linux/common/ddi/media_libva_util.cpp
```diff
@@ -405,6 +411,23 @@ VAStatus DdiMediaUtil_AllocateSurface(
     uint32_t    gmmPitch;
     uint32_t    gmmSize;
     uint32_t    gmmHeight;
+
+    gmmPitch    = (uint32_t)gmmResourceInfo->GetRenderPitch();
+    if( DdiMediaUtil_IsExternalSurface(mediaSurface) && ( mediaSurface->pSurfDesc->uiVaMemType == VA_SURFACE_ATTRIB_MEM_TYPE_USER_PTR ) )
+    {
+        uint32_t gmmHAlign;
+        gmmHAlign = gmmResourceInfo->GetHAlign();
```
dvrogozh: I think this better be on the single line as:
int32_t gmmHAlign = gmmResourceInfo->GetHAlign();

media_driver/linux/common/ddi/media_libva_util.cpp
```diff
@@ -354,6 +354,12 @@ VAStatus DdiMediaUtil_AllocateSurface(
                 gmmParams.BaseHeight = mediaSurface->pSurfDesc->uiSize / mediaSurface->pSurfDesc->uiPitches[0];
             }
         }
+        if(mediaSurface->pSurfDesc->uiVaMemType == VA_SURFACE_ATTRIB_MEM_TYPE_USER_PTR)
+        {
+            gmmParams.ExistingSysMemSize = mediaSurface->pSurfDesc->uiBuffserSize;
+            gmmParams.pExistingSysMem    = mediaSurface->pSurfDesc->ulBuffer;
+            gmmParams.Flags.Info.ExistingSysMem = true;
```
dvrogozh: Surprisingly this regress some use cases for me (where I use LibVA VP to copy surface from userptr allocated to driver allocated). Specifically YUY2 and UYVY regressed. And some cases which I anticipated will be fixed (some of i420, yv12) were not fixed.

YUY2 regressed in the following way. Input surface is 176x144 and I allocated it as follows:
1. 16-alignment for width and 64-bytes alignment for pitch. I.e. width aligned to 192, stride=384
2. height was 4-aligned and remained 144

With the above allocation YUY2 worked just fine before the change. After the change we fail on a call to CreateResInfoObject(). Basically gmmlib for some reason wants height to be 154 while it is eventually not. I am trying to clarify with gmmlib folks where this requirement comes from.

I420 case actually follows the same YUY2 failure pattern. Before the change it just was copied incorrectly.

media_driver/linux/common/ddi/media_libva_util.cpp
XinfengZhang: 1. there are no interface to set the pitch from application except the "OverridePitch"

media_driver/linux/common/ddi/media_libva_util.cpp
XinfengZhang: yes,

media_driver/linux/common/ddi/media_libva_util.cpp
dvrogozh: In this case, how CreateResInfoObject() manages to validate anything? I can't imagine memory allocation validation knowing nothing about the pitch.

media_driver/linux/common/ddi/media_libva_util.cpp
XinfengZhang: sure

media_driver/linux/common/ddi/media_libva_util.cpp
XinfengZhang: yes, let me add engineer from https://github.com/intel/gmmlib , @drprajap, @kmvadsar

media_driver/linux/common/ddi/media_libva_util.cpp
XinfengZhang: @dvrogozh have you synced with gmmlib folks about it? why the alignment required 154 instead 144?

PR #717, opened 2019-08-29
Add VP9 dual pipe encoder support for KBL

sreerenjb:
The patch series is adding VP9 dual pipe encoder support for GEN9. It uses the existing media-kernel for gen9. Some of the parameter enable/disable is based on the legacy i965 driver implementation. Dynamic resolution change is not supported in this patch series.


rosetta-jpn:
Thanks for this feature. When is this feature expected to be checked in?

jbarnes993:
There's a Chromium bug for this too, but fwiw it would be either nice to have this or have the Intel media team support the i965 driver on KBL.  I think we'd prefer this CL though, since it seems to be fairly efficient relative to i965, at least for VP9.  Talk to hiroh for more details.

rosetta-jpn:
Sree, does this patch work for CML and WHL too, not only KBL?

sreerenjb:
Hiro, I haven't tested on CML & WHL. Ideally, similar code base should work for all since they share the same exact GEN9 graphics. But, since I limited some of the HW command extension in the patch only to kbl (eg: https://github.com/intel/media-driver/pull/717/commits/c00c1ec88de10c4e3e47f790f7dfe719ccfe6337), it many not work as it is on CML. We may move the kbl specific codebase to base classes in gen9 directory.

karoldob:
I've tested this patch on Ubuntu 21.04, merged into media-driver 21.1.1 branch,
Results:
on Intel gen10 - all works fine
on Intel gen11 - encoding process has no erros, but wideo is completely artifacted.

sreerenjb:
Gen11 has a fixed function VP9 encode and the iHD driver officially supports it. This patch is not intended for the full fixed function solution.

IK4MS:
Can confirm it appears to work fine on CML (10700K), at least with the few settings that I used.

ccorbacho:
Is there any chance of this being reviewed and merged? As it's been open for 6 years now, but is still useful for those of us with older hardware. @XinfengZhang ?

PR #724, opened 2019-09-06
BRC reset depending on external bits reset flag

semelianova:
Add possibilities to BRC reset in all cases regardless RateControlMethod
depending on external bits reset flag


XinfengZhang:
yes, if the external bit is set, we need to set BRC reset. but if the bitrate changed , we also need it, right?

PR #728, opened 2019-09-12
Avoid invalid write in media_libva_vp.c

xhaihao:
According to the documentation[1], user may provide an array
which can't hold all filter caps, so the driver must check the
the number of elements in the input array, otherwise it might
result in invalid write

Issue can be catched by the command below:

valgrind gst-launch-1.0 videotestsrc ! video/x-raw,width=720,height=480 !
vaapipostproc video-direction=90r ! vaapisink

==25222== Invalid write of size 4
==25222==    at 0x946197A:
DdiVp_QueryVideoProcFilterCaps(VADriverContext*, unsigned int, int,
void*, unsigned int*) (media_libva_vp.c:4100)
==25222==    by 0x93FF773:
DdiMedia_QueryVideoProcFilterCaps(VADriverContext*, unsigned int,
_VAProcFilterType, void*, unsigned int*) (media_libva.cpp:5646)
==25222==    by 0x6D19641: vaQueryVideoProcFilterCaps (va.c:2183)
==25222==    by 0x6C3A530: vpp_get_filter_caps_unlocked
(gstvaapifilter.c:238)
==25222==    by 0x6C3A530: vpp_get_filter_caps (gstvaapifilter.c:272)
==25222==    by 0x6C3A530: get_operations_ordered (gstvaapifilter.c:724)

[1] https://github.com/intel/libva/blob/d49b99d5d88f66e2e1ca451cf19251c7950c870d/va/va_vpp.h#L1408-L1426

### DIFF
--- a/media_driver/linux/common/vp/ddi/media_libva_vp.c
+++ b/media_driver/linux/common/vp/ddi/media_libva_vp.c
@@ -4072,15 +4072,15 @@ DdiVp_QueryVideoProcFilterCaps (
             {
                 VAProcFilterCap* baseCap     = (VAProcFilterCap*) filter_caps;
 
-                baseCap->range.min_value     = NOISEREDUCTION_MIN;
-                baseCap->range.max_value     = NOISEREDUCTION_MAX;
-                baseCap->range.default_value = NOISEREDUCTION_DEFAULT;
-                baseCap->range.step          = NOISEREDUCTION_STEP;
-
                 if (uQueryCapsNum < uExistCapsNum)
                 {
                     return VA_STATUS_ERROR_MAX_NUM_EXCEEDED;
                 }
+
+                baseCap->range.min_value     = NOISEREDUCTION_MIN;
+                baseCap->range.max_value     = NOISEREDUCTION_MAX;
+                baseCap->range.default_value = NOISEREDUCTION_DEFAULT;
+                baseCap->range.step          = NOISEREDUCTION_STEP;
             }
             break;
 
@@ -4094,13 +4094,13 @@ DdiVp_QueryVideoProcFilterCaps (
             {
                 VAProcFilterCapDeinterlacing* diCap = (VAProcFilterCapDeinterlacing*) filter_caps;
 
-                diCap[0].type                         = VAProcDeinterlacingBob;
-                diCap[1].type                         = VAProcDeinterlacingMotionAdaptive;
-
                 if (uQueryCapsNum < uExistCapsNum)
                 {
                     return VA_STATUS_ERROR_MAX_NUM_EXCEEDED;
                 }
+
+                diCap[0].type                         = VAProcDeinterlacingBob;
+                diCap[1].type                         = VAProcDeinterlacingMotionAdaptive;
             }
             break;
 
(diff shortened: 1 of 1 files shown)


media_driver/linux/common/vp/ddi/media_libva_vp.c
```diff
@@ -4208,6 +4213,12 @@ DdiVp_QueryVideoProcFilterCaps (
                 if (uQueryFlag == QUERY_CAPS_ATTRIBUTE)
                 {
                     VAProcFilterCapHighDynamicRange *HdrTmCap = (VAProcFilterCapHighDynamicRange *)filter_caps;
+
+                    if (uQueryCapsNum < uExistCapsNum)
+                    {
+                        return VA_STATUS_ERROR_MAX_NUM_EXCEEDED;
+                    }
+
```
dvrogozh: I believe there is one more thing which needs to be changed in this code. Basically, `num_filter_caps` gets assigned even in those cases when function will return an error. This is bad style especially when this can be avoided like here. So, I suggest to avoid assignment of num_filter_caps before we processed all error checks. And we actually can move assign out of the switch:
```
uint32_t uExistCapsNum = 0;  // please, add initialization
switch (type)
{
... // assign uExistCapsNum here
}

*num_filter_caps = uExistCapsNum;
```

media_driver/linux/common/vp/ddi/media_libva_vp.c
```diff
@@ -4167,6 +4167,11 @@ DdiVp_QueryVideoProcFilterCaps (
             {
                 VAProcFilterCap* baseCap     = (VAProcFilterCap *)filter_caps;
 
+                if (uQueryCapsNum < 1)
```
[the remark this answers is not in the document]
xhaihao: The original code didn't use `uExistCapsNum`, but I may change this code to use `uExistCapsNum`

media_driver/linux/common/vp/ddi/media_libva_vp.c
xhaihao: `num_filter_caps` is assigned before error checking in the original code, but I agree with you that we'd better to move the assign out of the switch.

xhaihao: According to https://github.com/intel/libva/blob/d49b99d5d88f66e2e1ca451cf19251c7950c870d/va/va_vpp.h#L1408-L1426, num_filter_caps must be adjusted when VA_STATUS_ERROR_MAX_NUM_EXCEEDED is returned, so `num_filter_caps` should get assigned before error checking.

media_driver/linux/common/vp/ddi/media_libva_vp.c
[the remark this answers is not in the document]
xhaihao: I updated the code, could you review it again when you get a chance?

xhaihao:
@dvrogozh any comment on the new version of patch?

media_driver/linux/common/vp/ddi/media_libva_vp.c
```diff
@@ -4179,6 +4185,9 @@ DdiVp_QueryVideoProcFilterCaps (
             {
                 VAProcFilterCapTotalColorCorrection* TccCap;
 
+                if (uQueryCapsNum < uExistCapsNum)
+                    return VA_STATUS_ERROR_MAX_NUM_EXCEEDED;
```
FurongZhang: Here, to make the coding style consistent, is it possible to use {}? Thanks a lot in advance!

xhaihao: Thanks for the comment, I updated the patch.

FurongZhang:
Thank you Haihao for the code changes. It is looking great!

PR #740, opened 2019-10-08
Fix Build fail under CentOS7.6 with its native gcc 4.8 compiler

### LINKED ISSUE #736: Build fail under CentOS7.6 with its native gcc 4.8 compiler
Two building issues found by now:
1. In constructor 'PerfUtility::PerfUtility()', error: array used as initializer. Consider to use char string[num] = {'\0'} instead of char string[num] = ""
2. unrecognized command line option "-Wno-delete-incomplete" [-Werror]. It will cause the MOS_Delete(m_bltInterface) fail and mhwBltInterface not defined.

shawnli2:
array used as initializer

fixes #736


jbeich:
Likely fixes #812 as well.

PR #777, opened 2019-11-15
Disable the scalability on TGL for Linux.

wangyan42164:
The scalability on TGL for Linux hasn't been finished yet.
If enabling, it will cause GPU hang.


wangyan-intel:
It is unnecessary now. Close it.

PR #783, opened 2019-11-20
add prediction direction caps for AVC VDEnc

XinfengZhang:
related libva change: https://github.com/intel/libva/pull/220


media_driver/linux/common/ddi/media_libva_caps.cpp
```diff
@@ -762,7 +762,13 @@ VAStatus MediaLibvaCaps::CreateEncAttributes(
         attrib.value = attribValMaxFrameSize.value;
         (*attribList)[attrib.type] = attrib.value;
     }
-
+    
+    if (IsAvcProfile(profile) && (entrypoint == VAEntrypointEncSliceLP))
```
fulinjie: This seems to add caps for AVC only, how about HEVC?

PR #792, opened 2019-11-29
Fix device info for GT1 BXT / GLK SKUs

felixbuenemann:
* UHD Graphics 500 / 600 are 12 EU GT1 SKUs
* UHD Graphics 505 / 605 are 18 EU GT1.5 SKUs

This also changes 18 EU SKUs to be treated as GT 1.5.

I've also added `0x0a84` as BXT 18 EU device based on [i965_pci_ids.h](https://gitlab.freedesktop.org/mesa/mesa/blob/master/include/pci_ids/i965_pci_ids.h) from mesa.


XinfengZhang:
BXT has 2x6 and 3x6 configure, so 2x6 is gt1, 3x6 is gt1.5, LGTM

PR #806, opened 2019-12-20
Start candidate searching after last active frame.

wangyan42164:
It will avoid selecting the 1st element at rtTbl always and
overwriting needed reference frame when rtTbl is full.


wangyan-intel:
@XinfengZhang Could you please review this PR? Thanks.

wangyan-intel:
After disccusion offline, https://github.com/intel/media-driver/pull/823 should be the better solution. So drop this PR.

PR #817, opened 2020-01-10
Calculate the right offset when vaCreateImage need width or height al…

wangyan42164:
…ignment.

Signed-off-by: Liang Li <liang.a.li@intel.com>


FurongZhang:
No concern from my side.

hanlong1:
Yan, could you please rebase this change and add "verifying" again

Issue #640, opened 2019-06-10
[Gst-vaapi][VPP][deinterlace] Bob and Motion-adaptive got the same result.


feiwan1:
When set the deinterlace mode as Bob and Motion-adaptive, the output are bit match:

$ gst-launch-1.0 -vf filesrc location=1080p_MotionTest.yuv num-buffers=20 ! rawvideoparse format=nv12 width=1920 height=1080 interlaced=true top-field-first=1 ! vaapipostproc format=nv12 width=1920 height=1080 deinterlace-mode=1 deinterlace-method=bob ! checksumsink2 dump-output=true dump-location=./de.bob.nv12

$ gst-launch-1.0 -vf filesrc location=1080p_MotionTest.yuv num-buffers=20 ! rawvideoparse format=nv12 width=1920 height=1080 interlaced=true top-field-first=1 ! vaapipostproc format=nv12 width=1920 height=1080 deinterlace-mode=1 deinterlace-method=motion-adaptive ! checksumsink2 dump-output=true dump-location=./de.ma.nv12

$md5sum de*.nv12
1ebd131328ce1475f98f623fd11b34ae  de.bob.nv12
1ebd131328ce1475f98f623fd11b34ae  de.ma.nv12

Expected result:
dump output nv12 file de.bob.nv12 should not bit match with de.ma.nv12.

FurongZhang:
Adding Jason to check this issue.

FurongZhang:
Jason, is it possible for you to assign this issue to yourself? Sorry, I did not know your account.

Jister:
@feiwan1 We have kernel solution and Vebox solution for BOB DI, and media driver use VEBOX to do it for better quality. For motion adaptive, app need to give reference frame(previous frame) to driver, but gst-vaapi didn't give reference, thus result is same to BOB. The issue need gst-vaapi's change to pass reference frame to driver.

feiwan1:
@Jister , looks media-driver performs different behavior with vaapi-drvier. When using vaapi-driver, it doesn't have this issue. @xhaihao , do you know how vaapi-driver deal with Motion-adaptive deinterlace? Is that make sense to provide an additional reference frame in middle-ware ?

Jister:
@feiwan1 By the way, can you help to compare the results between media-driver and vaapi-driver, what's the value of md5sum of vaapi-driver using the same input?

feiwan1:
@Jister , here is vaapi-driver's result:
[...]

PR #820, opened 2020-01-13
[VP] Add correct reference number for DI

### LINKED ISSUE #804: Regressed deinterlacing support as compared to i965-va-driver
I have tried out media-driver 19.3.x on a few systems with Broadwell and Skylake CPUs (Specifically, HD Graphics 5500 and Iris 540) (reproduced with Linux 5.3, 5.4, and 5.5-rc) and have found that the hardware deinterlacing support doesn't work as well compared to the i965-va-driver.  Here's what I have found:

1. If ENABLE_NONFREE_KERNELS=OFF is set, double-frame-rate deinterlacing doesn't work at all, resulting in 1080i30 content created from a 60FPS source playing jerkily at 30FPS instead of 60FPS as it did with the i965-va-driver.  (The previous driver supported double-frame-rate even when the non-free features were disabled.)  This is important for many news and sports broadcasts on 1080i ATSC channels.
2. If ENABLE_NONFREE_KERNELS=ON is set, double-frame-rate deinterlacing works only sometimes and with reduced quality compared to i965-va-driver.  In VLC, the frame rate is always doubled, but the video seems subjectively fuzzier than with i965-va-driver and horizontal lines seem to "shimmer" up and down.  In Kodi, it seems to alternate between no frame rate doubling (resulting in 30FPS) and the degraded frame rate doubling as I saw in VLC every few seconds at random intervals.

I apologize for not including any more information with this report, but there doesn't seem to be anything relevant in the logging output from either application and I'm not sure what else would be helpful.  I will certainly provide any requested information though.  Thanks!

Jister:
Fix #804 #640

### DIFF
--- a/media_driver/agnostic/common/hw/mhw_vebox.h
+++ b/media_driver/agnostic/common/hw/mhw_vebox.h
@@ -288,6 +288,8 @@ typedef struct _MHW_VEBOX_DNDI_PARAMS
     uint32_t  dwLPFWtLUT5;
     uint32_t  dwLPFWtLUT6;
     uint32_t  dwLPFWtLUT7;
+
+    bool      bMcdiEnable;
 } MHW_VEBOX_DNDI_PARAMS, *PMHW_VEBOX_DNDI_PARAMS;
 
 //!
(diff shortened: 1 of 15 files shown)


FurongZhang:
Thanks Jason for  this code changes. It looks good to me. We can merge into master branch.

smp79:
Any news for this? MADI/MCDI deinterlacing still does not work.

PR #821, opened 2020-01-13
Add VAProcColorStandardExplicit for input and output color standard q…

sunxinpeng:
…uery.


sunxinpeng:
@FurongZhang Hi, this PR is to make the VAProcColorStandardExplicit can be queried and used by application, which ensures that transfer characteristics can be set correctly in the driver when doing tone-mapping. Could you help to review and push it?

uartie:
@sunxinpeng please fix the compile error:
```
error: too many initializers for ‘VAProcColorStandardType [5]’
```

PR #845, opened 2020-02-14
Ignore the private VP9 status buffer

xhaihao:
Without a flag, user doesn't know VACodedBufferSegment.next pointing
to a private segment status, so we have to ignore this private
status buffer. Otherwise it will result in segmentation fault if user
takes VACodedBufferSegment.next as a VACodedBufferSegment pointer

Example:
gst-launch-1.0 videotestsrc !
video/x-raw,format=NV12,width=3840,height=2160 ! vaapivp9enc !
matroskamux ! filesink location=output.vp9

### DIFF
--- a/media_driver/linux/common/codec/ddi/media_ddi_encode_vp9.cpp
+++ b/media_driver/linux/common/codec/ddi/media_ddi_encode_vp9.cpp
@@ -1062,7 +1062,12 @@ VAStatus DdiEncodeVp9::ReportExtraStatus(
     codedBufStatus->next_frame_width = encodeStatusReport->NextFrameWidthMinus1 + 1;
     codedBufStatus->next_frame_height = encodeStatusReport->NextFrameHeightMinus1 + 1;
 
-    codedBufferSegment->next = codedBufStatus;
+    /*
+     * Ignore the private status buffer temporarily. According to the comment for VACodedBufferVP9Status in VA-API,
+     * driver must set codedBufferSegment->status to be VA_CODED_BUF_STATUS_CODEC_SPECIFIC, however
+     * VA_CODED_BUF_STATUS_CODEC_SPECIFIC is not defined in VA-API
+     */
+    // codedBufferSegment->next = codedBufStatus;
 
     return vaStatus;
 }


media_driver/linux/common/codec/ddi/media_ddi_encode_vp9.cpp
```diff
@@ -1062,7 +1062,12 @@ VAStatus DdiEncodeVp9::ReportExtraStatus(
     codedBufStatus->next_frame_width = encodeStatusReport->NextFrameWidthMinus1 + 1;
     codedBufStatus->next_frame_height = encodeStatusReport->NextFrameHeightMinus1 + 1;
 
-    codedBufferSegment->next = codedBufStatus;
+    /*
+     * Ignore the private status buffer temporarily. According to the comment for VACodedBufferVP9Status in VA-API,
+     * driver must set codedBufferSegment->status to be VA_CODED_BUF_STATUS_CODEC_SPECIFIC, however
+     * VA_CODED_BUF_STATUS_CODEC_SPECIFIC is not defined in VA-API
```
dvrogozh: This sounds like a bug in VAAPI spec. Is there libva bug for that? If not, can you, please, submit the one and link it here?

xhaihao: @dvrogozh It seems no one is interested in the private status buffer.

xhaihao: @XinfengZhang @dvrogozh I think it would be better to ignore this private buffer as no one is interested in this buffer. We may define VA_CODED_BUF_STATUS_CODEC_SPECIFIC in VAAPI spec and add this buffer back later if someone want to use this private buffer. How do you think?

VUNNYSODHI: Hi All,
Any update on merge status for this commit in media driver or we are looking for fix in gstreamer-vaapi?

sreerenjb:
Thank you Haihao for this!
I noticed this while implementing the gen9 vp9 encoder and put a hack in the iHD driver: https://github.com/intel/media-driver/pull/717/commits/681f7f2276060903ae912cf9c0b32309b59d65eb
Sorry I didn't file the bug :(

Even if iHD driver sets the VA_CODED_BUF_STATUS_CODEC_SPECIFIC flag in the current segment for indicating the next segment as status buffer, most of the existing middleware implementations will still segfault  unless there is a proper fix in the middleware.

For eg: this is the code in gstreamer
  for (segment = buf->segment_list; segment != NULL; segment = segment->next)
    size += segment->size;
It won't check for the status flag, and try to access the next segment as CodedData.

PR #848, opened 2020-02-17
Modify for fixing CIV two unlock KW issues on WW_08 2020

Kexiaox:


XinfengZhang:
@Kexiaox could you help to fix the build issue, seems you have not initialize tmpRet

PR #852, opened 2020-02-19
Fix the performance issue on APL.

FurongZhang:
Perf will drop from 25fps to 10fps for 1080p decode + vp(nv12tileY -> i420linear) if only mos_gem_bo_map_gtt(surface->bo) on AtomSOC(APL).


XinfengZhang:
LGTM

PR #856, opened 2020-02-24
media_libva_caps_g11: add max resolution support for vp8 encode

fulinjie:
Based on docs/media_features.md, the Max Res supported for ICL is 4K.

Same support in common media_libva_caps:
https://github.com/intel/media-driver/blob/master/media_driver/linux/common/ddi/media_libva_caps.cpp#L2640

Signed-off-by: Linjie Fu <linjie.fu@intel.com>


fulinjie:
ping

Issue #873, opened 2020-03-12
Mismatch of "intra_insert_size" for Rolling Intra Refresh feature


djie1:
In VAAPI, VAEncMiscParameterRIR::intra_insert_size is defined as "Indicates the number of columns or rows in MB", see https://github.com/intel/libva/blob/master/va/va.h#L2137,

but in iHD driver, it sets to "have an overlap of one MB row or column of Intra macroblocks",  see https://github.com/intel/media-driver/blob/master/media_driver/agnostic/gen9_skl/codec/hal/codechal_encode_avc_g9_skl.cpp#L2458 for example on G9 SKL.

So the result is that in the encoded stream with RIR enabled, the width of MB row or column of intra  MBs is 1 more than the specified "intra_insert_size" as specified in VAAPI.

djie1:
In the source file, there is comment says "Pass the same IntraRefreshUnit to the kernel w/o the adjustment by -1, so as to have an overlap of one MB row or column of Intra macroblocks across one P frame to another P frame, as needed by the RollingI algo", explains the reason for the assignment.
I don't understand why the "RollingI algo" requires overlap of one MB row of column. Even it's the case, it should be set by VAAPI application to VAEncMiscParameterRIR::intra_insertion_location to make the overlap, but not changing application's input inside the driver.

Submitted PR#875 for proposal of the fix.

Xiaogangli-intel:
Hi @djie1, I checked this issue/source code again, and synced with some engineers who familiar with this history, actually, this is not issue, pass IntraRefreshUnit to HW directly is what HW expected.\nNot sure if you still have concern about it, if yes, please feel free let us know. Thank you for your contribution.

Xiaogangli-intel:
Hi @djie1, I close this issue because it's not an issue as I descripted above. If you still have any concern about it, let me know please.

djie1:
> Hi @djie1, I checked this issue/source code again, and synced with some engineers who familiar with this history, actually, this is not issue, pass IntraRefreshUnit to HW directly is what HW expected. Not sure if you still have concern about it, if yes, please feel free let us know. Thank you for your contribution.

your reply didn't solve the issue I described.

PR #875, opened 2020-03-12
[Encode][RIR] correct the value of "IntraRefreshMBSizeMinusOne" to kernel or hardware.

djie1:
Correctly set the value of "IntraRefreshMBSizeMinusOne".

The incorrect assignment makes that in the encoded stream, size of the Intra refresh MBs is actual VAEncMiscParameterRIR::intra_insert_size + 1, which is one bug as it fails the documentation in VAAPI which says that "intra_insert_size" "Indicates the number of columns or rows in MB".
In the source file, there is comment says "Pass the same IntraRefreshUnit to the kernel w/o the adjustment by -1, so as to have an overlap of one MB row or column of Intra macroblocks across one P frame to another P frame, as needed by the RollingI algo", explains the reason for the assignment.
I don't understand why the "RollingI algo" requires overlap of one MB row of column. Even it's the case, it should be set to VAEncMiscParameterRIR::intra_insertion_location from VAAPI application level to make the overlap.

Change-Id: I0ebc387dfd92f58dec313fc53eca8dd6a6a20110

### DIFF
--- a/media_driver/agnostic/gen10/codec/hal/codechal_encode_avc_g10.cpp
+++ b/media_driver/agnostic/gen10/codec/hal/codechal_encode_avc_g10.cpp
@@ -4956,8 +4956,6 @@ MOS_STATUS CodechalEncodeAvcEncG10::SetCurbeAvcMbEnc(
         /* Multiple predictor should be completely disabled for the RollingI feature. This does not lead to much quality drop for P frames especially for TU as 1 */
         cmd.m_encCurbe.DW32.MultiPredL0Disable = CODECHAL_ENCODE_AVC_MULTIPRED_DISABLE;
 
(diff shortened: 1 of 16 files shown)


XinfengZhang:
@aleonenk , need your help to check it , it changes the behavior to avoid +1 to the Refresh Size for the overlap.

Xiaogangli-intel:
Hi @djie1, as I mentioned in #873, I  synced with some engineers who familiar with this history, actually, this is not an issue, pass IntraRefreshUnit to HW directly is what HW expected. So don't need to use IntraRefreshUnit - 1.
Thanks for your contribution.

Xiaogangli-intel:
I closed #873 because it's not an issue. So I will also close this PR. @djie1, thanks for your contribution, if you still have concern, let use know please, thank you.

Issue #876, opened 2020-03-12
Memory corruption caused by incorrect typecasting in KernelDll_BuildKernel_CmFc()


ykku16github:
When KernelDll_BuildKernel_CmFc() calls cm_fc_combine_kernels() (see [1]), one of the inputs is &dwEstimatedKernelSize.  It is a uint32_t pointer but is typecasted to a size_t pointer.

On platforms where size_t is a 64-bit integer, subsequent write operations using this size_t pointer results in 4 bytes of memory corruption.

[1] https://github.com/intel/media-driver/blob/master/media_driver/agnostic/common/vp/kdll/hal_kerneldll.c#L5045

FurongZhang:
I have commented the code changes #877. I think we had better not to use size_t for coding style(recommend to use uint32_t, uint64_t etc.).

FurongZhang:
There is code fix from @ykku16github , we are discussing the code changes.

FurongZhang:
The PR has been merged. @ykku16github I will close the issue. Please feel free to open it if you still see any issues.

PR #877, opened 2020-03-12
[VP] Fix memory corruption

ykku16github:
When KernelDll_BuildKernel_CmFc() calls cm_fc_combine_kernels()
(see [1]), one of the inputs is &dwEstimatedKernelSize.  It is a
uint32_t pointer but is typecasted to a size_t pointer.

On platforms where size_t is a 64-bit integer, subsequent write
operations using this size_t pointer results in 4 bytes of
memory corruption.

[1] https://github.com/intel/media-driver/blob/master/media_driver/agnostic/common/vp/kdll/hal_kerneldll.c#L5045

### DIFF
--- a/media_driver/agnostic/common/vp/kdll/hal_kerneldll.c
+++ b/media_driver/agnostic/common/vp/kdll/hal_kerneldll.c
@@ -4916,7 +4916,7 @@ bool KernelDll_BuildKernel_CmFc(Kdll_State *pState, Kdll_SearchState *pSearchSta
     int32_t           iOffset;
     uint32_t          dwResolveOffset[DL_MAX_EXPORT_COUNT];
     uint32_t          dwTotalKernelCount;
-    uint32_t          dwEstimatedKernelSize;
+    size_t            stEstimatedKernelSize;
     int32_t           iKUID;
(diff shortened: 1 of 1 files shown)


ykku16github:
This pull request addresses https://github.com/intel/media-driver/issues/876.

FurongZhang:
Thanks @ykku16github for the code changes. But generally, we use uint32_t, uint16_t etc. for coding style. Is it possible not to change to size_t? cm_fc_combine_kernels is a function which is originally from CMFC compiler, so it is using size_t instead of uint32_t or uint64_t. 

Thanks a lot in advance!

ykku16github:
@FurongZhang , thanks for reviewing.  The important thing is to make the types match between the 2 functions.

So, if we can't use size_t in KernelDll_BuildKernel_CmFc(), then that means we can't use size_t in cm_fc_combine_kernels().  Can the function prototype for cm_fc_combine_kernels() be changed to accept uint32_t?  Would this be your preference?

My preference is to use size_t.  Before I submitted this pull request, I had searched for size_t within the media driver.  I actually see many places where size_t is used.

FurongZhang:
@ykku16github OK.

PR #884, opened 2020-03-19
use gmmlib caculate the NV12 surface offset

XinfengZhang:
should apply similiar changes for other planar format in near future
it is quick fix for chromium encoding issue

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -4406,12 +4406,24 @@ VAStatus DdiMedia_DeriveImage (
     vaimg->height                   = mediaSurface->iRealHeight;
     vaimg->format.byte_order        = VA_LSB_FIRST;
 
+    GMM_RESOURCE_INFO          *gmmResourceInfo = mediaSurface->pGmmResourceInfo;
+    GMM_REQ_OFFSET_INFO reqInfo     = {0};
+    reqInfo.Plane                   = GMM_PLANE_U;
+    reqInfo.ReqRender               = 1;
+    gmmResourceInfo->GetOffset(reqInfo);
+    uint32_t offsetU                = reqInfo.Render.Offset;
+    MOS_ZeroMemory(&reqInfo, sizeof(GMM_REQ_OFFSET_INFO));
+    reqInfo.Plane                   = GMM_PLANE_V;
+    reqInfo.ReqRender               = 1;
+    gmmResourceInfo->GetOffset(reqInfo);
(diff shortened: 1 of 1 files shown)


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -4405,13 +4405,16 @@ VAStatus DdiMedia_DeriveImage (
     vaimg->width                    = mediaSurface->iWidth;
     vaimg->height                   = mediaSurface->iRealHeight;
     vaimg->format.byte_order        = VA_LSB_FIRST;
+    GMM_RESOURCE_INFO          *gmmResourceInfo = mediaSurface->pGmmResourceInfo;
+    vaimg->data_size                = (uint32_t)gmmResourceInfo->GetSizeSurface();
+    uint32_t UPlaneXOffset          = (uint32_t)gmmResourceInfo->GetPlanarXOffset(GMM_PLANE_U);
+    uint32_t UPlaneYOffset          = (uint32_t)gmmResourceInfo->GetPlanarYOffset(GMM_PLANE_U);
```
yellowdoge: Suggestion: make these two vars `const`, so that readers can assume they are essentially aliases.

XinfengZhang:
@andrescj-chromium they are different, #869 is changing the vertical alignment, this PR is changing the UV offset of deriveimage

media_driver/linux/common/ddi/media_libva.cpp
XinfengZhang: agree

media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -4564,21 +4563,27 @@ VAStatus DdiMedia_DeriveImage (
...
         vaimg->format.bits_per_pixel    = 12;
-        vaimg->data_size                = mediaSurface->iPitch * mediaSurface->iHeight * 3 / 2;
         vaimg->num_planes               = 2;
         vaimg->pitches[0]               = mediaSurface->iPitch;
         vaimg->pitches[1]               =
         vaimg->pitches[2]               = mediaSurface->iPitch;
         vaimg->offsets[0]               = 0;
-        vaimg->offsets[1]               = mediaSurface->iHeight * mediaSurface->iPitch;
-        vaimg->offsets[2]               = vaimg->offsets[1] + 1;
+        if(MEDIA_IS_WA(&mediaCtx->WaTable, FtrMemoryCompression))
```
Jister: Should use FtrE2ECompression here.

XinfengZhang: thanks, accept

Issue #882, opened 2020-03-18
[TGL] HEVC main-444-10 can not get correct  VAConfigAttribRTFormat attribute.


HeJunyan:
vaGetConfigAttributes failed to get the attribute for VAConfigAttribRTFormat on TGL, it should be VA_RT_FORMAT_YUV444_10.
I notice that the m_profileEntryTbl for Gen12 VAProfileHEVCMain444_10 is not inited correctly.

PR #888, opened 2020-03-23
Gen12: Set correct value for VAConfigAttribRTFormat for HEVC main-444 and main-444-10

xhaihao:
Without this setting, user application will fail to decode HEVC 8/10 bits 444 videos

Sample pipeline:

gst-launch-1.0 filesrc location=input.10bit.444.h265 ! h265parse ! vaapih265dec ! fakesink

Fixes https://github.com/intel/media-driver/issues/882


XinfengZhang:
LGTM, @AppleWang001 please help to check

wangyan-intel:
LGTM

PR #891, opened 2020-03-24
Version bump

onabiull:
Master code target 20.2, so let us reflect this in sources.


XinfengZhang:
@onabiull please pay attention to  commit message , "bump version to 20.2 pre" should be better. sorry , maybe late

PR #904, opened 2020-04-14
GEN12: HEVC: Don't enable real tile if no subset parameters on Linux

xhaihao:
Real tile decoding depends on subset parameters. It will result in GPU
hang if user doesn't provide subset parameters.

Example:
gst-launch-1.0 filesrc location=input.tile.h265 ! h265parse ! \
vaapih265dec ! fakesink

### DIFF
--- a/media_driver/linux/common/codec/ddi/media_ddi_decode_hevc.cpp
+++ b/media_driver/linux/common/codec/ddi/media_ddi_decode_hevc.cpp
@@ -433,6 +433,14 @@ VAStatus DdiDecodeHEVC::RenderPicture(
         }
         case VASubsetsParameterBufferType:
         {
+
+            if (m_ddiDecodeCtx->DecodeParams.m_subsetParams == nullptr) {
+                m_ddiDecodeCtx->DecodeParams.m_subsetParams = MOS_AllocAndZeroMemory(sizeof(CODEC_HEVC_SUBSET_PARAMS));
+
+                if (m_ddiDecodeCtx->DecodeParams.m_subsetParams == nullptr)
+                    break;
+            }
+
             MOS_SecureMemcpy(m_ddiDecodeCtx->DecodeParams.m_subsetParams, dataSize, data, dataSize);
             break;
         }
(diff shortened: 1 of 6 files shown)


xhaihao:
@Jexu, I updated the patch, could you help to review it again? Thanks!

media_driver/linux/common/codec/ddi/media_ddi_decode_hevc.cpp
```diff
@@ -433,6 +433,14 @@ VAStatus DdiDecodeHEVC::RenderPicture(
         }
         case VASubsetsParameterBufferType:
         {
+
+            if (m_ddiDecodeCtx->DecodeParams.m_subsetParams == nullptr) {
+                m_ddiDecodeCtx->DecodeParams.m_subsetParams = MOS_AllocAndZeroMemory(sizeof(CODEC_HEVC_SUBSET_PARAMS));
```
Jexu, line 438: Better not move init here. it will cause some issue. In hal level,  umd will copy m_subsetParams again to another array and use the array for long format decode. If the array is nullptr, it will return error. 
check function 'InitSliceTileParams' in codechal_decode_hevc_long_g12.cpp

xhaihao: `InitSliceTileParams` is called when `m_isSeparateTileDecoding || m_isRealTile` is true
1. for read tile decoding,  user has already provided a valid subset buffer by calling `vaRenderPicture`, so `InitSliceTileParams` should return MOS_STATUS_SUCCESS
2. If separate tile decoding,  may user not provide subset buffer?

xhaihao: @Jexu Thanks for the comment, I added code to handle the case of m_subsetParams=NULL, could you help to review it again?

Jexu: done！it is fine.

PR #911, opened 2020-04-19
media_libva: Fix memory leak

xhaihao:
This patchset fixed some memory leak in the driver.

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -4083,6 +4083,8 @@ VAStatus DdiMedia_CreateImage(
     vaimg->height               = height;
     vaimg->data_size            = size;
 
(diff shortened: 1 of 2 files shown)


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -6035,6 +6039,12 @@ VAStatus DdiMedia_ReleaseBufferHandle(
         buf->uiMemtype = 0;
     }
     DdiMediaUtil_UnLockMutex(&mediaCtx->BufferMutex);
+
+    if (!buf->uiExportcount && buf->bPostponedBufFree) {
```
FurongZhang, line 6045: It makes sense to add the code here. Thanks Haihao for the fix.

FurongZhang:
@xhaihao  I read the code, it looks good to me. Do you still need Carl's review, if no, I can merge into master.

PR #926, opened 2020-05-05
[CP] Unload CP lib only if it was loaded.

andrescj-chromium:
While debugging an issue, I noticed that we go inside the if in [1] which indicates an invalid condition. The problem is that DdiMedia_Terminate() unconditionally unloads CP lib, even when loading it fails. My proposed solution is to keep track of whether the CP lib was loaded on DdiMedia__Initialize() in order to decide if it needs to be unloaded on DdiMedia_Terminate().

[1] https://github.com/intel/media-driver/blob/7b4f9c93c8a0825ce1e8d9ab45d06f9e887a7150/media_driver/linux/common/cp/shared/cplib_utils.cpp#L136-L140


Xiaogangli-intel:
There should no issue with current way, but your change make the follow clearly. Thank you.

PR #933, opened 2020-05-14
fix usr ptr import error issue

dyang23:
Revert "check pitch from application for user ptr"
    
    This reverts commit ab3c3c6d5441f0800d34e0dbc5a5e7d2d7f144f1.
    There is a regression with this patch, the import from usr ptr
    failed.


onabiull:
I suggest to kill this PR as it stresses CI systems. Please contribute revert as a single commit on top of current master branch.

PR #939, opened 2020-05-25
[VP] Align cmake requirements

ljia5:
Align cmake requirements to extend Linux distributions coverage.


CMakeLists.txt
```diff
@@ -18,7 +18,7 @@
 # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
 
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 2.8)
```
dvrogozh, line 21: imho, that's not good idea to go down in requirements in top level cmake file. Can we go with cmake 3.5 (which looks to be current max) across the driver cmakes?

XinfengZhang: from my perspective, if we use latest cmake version, it require old user to promote the cmake version. and new version alwasy compatible with old one.   new version adoption should be determined by new feature.  have we use some special version which 3.5 support but 2.8 not?

PR #940, opened 2020-05-26
media_libva: avoid segmentation fault

xhaihao:
Without this fix, the media context is still used after freed.

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -1938,16 +1938,16 @@ static VAStatus DdiMedia_Terminate (
     mediaCtx->SkuTable.reset();
     mediaCtx->WaTable.reset();
 
-    // release media driver context, ctx creation is behind the mos_utilities_init
-    // If free earilier than MOS_utilities_close, memnja count error.
-    MOS_FreeMemory(mediaCtx);
-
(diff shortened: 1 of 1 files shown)


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -1938,12 +1938,14 @@ static VAStatus DdiMedia_Terminate (
...
 
+    bool cplib_was_loaded = mediaCtx->cpLibWasLoaded;
+
     // release media driver context, ctx creation is behind the mos_utilities_init
     // If free earilier than MOS_utilities_close, memnja count error.
     MOS_FreeMemory(mediaCtx);
 
     ctx->pDriverData = nullptr;
-    if (mediaCtx->cpLibWasLoaded)
+    if (cplib_was_loaded)
```
XinfengZhang, line 1948: why not move MOS_FreeMemory(mediaCtx) behind of cplib unload?

xhaihao: It doesn't change the code flow.  I may update the patch to unload cplib firstly if you prefer this way.  @XinfengZhang

xhaihao: @XinfengZhang  I updated the patch, could you please review it again?

PR #943, opened 2020-05-27
vaSyncSurface2 driver implementation

### LINKED ISSUE #932: Implementation of vaSyncSurface2
Please implement vaSyncSurface2 (https://github.com/intel/libva/pull/407/)

wyc55069407:
Add driver DDI side vaSyncSurface2 implementation with timeout.
Modify vaSyncSurface to share same status check function after bo_wait.

Fixes #932

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -3626,43 +3626,16 @@ static VAStatus DdiMedia_EndPicture (
     return vaStatus;
 }
 
-/*
- * This function blocks until all pending operations on the render target
- * have been completed.  Upon return it is safe to use the render target for a
- * different picture.
- */
-static VAStatus DdiMedia_SyncSurface (
-    VADriverContextP    ctx,
-    VASurfaceID         render_target
+static VAStatus DdiMedia_StatusCheck (
+    PDDI_MEDIA_CONTEXT mediaCtx,
+    DDI_MEDIA_SURFACE  *surface,
+    VASurfaceID surface_id
 )
 {
-    PERF_UTILITY_AUTO(__FUNCTION__, "ENCODE", "DDI");
-
     DDI_FUNCTION_ENTER();
 
-    DDI_CHK_NULL(ctx,    "nullptr ctx",    VA_STATUS_ERROR_INVALID_CONTEXT);
-
-    PDDI_MEDIA_CONTEXT mediaCtx = DdiMedia_GetMediaContext(ctx);
-    DDI_CHK_NULL(mediaCtx,               "nullptr mediaCtx",               VA_STATUS_ERROR_INVALID_CONTEXT);
-    DDI_CHK_NULL(mediaCtx->pSurfaceHeap, "nullptr mediaCtx->pSurfaceHeap", VA_STATUS_ERROR_INVALID_CONTEXT);
-
-    DDI_CHK_LESS((uint32_t)render_target, mediaCtx->pSurfaceHeap->uiAllocatedHeapElements, "Invalid render_target", VA_STATUS_ERROR_INVALID_SURFACE);
-
-    DDI_MEDIA_SURFACE  *surface = DdiMedia_GetSurfaceFromVASurfaceID(mediaCtx, render_target);
-    DDI_CHK_NULL(surface,    "nullptr surface",      VA_STATUS_ERROR_INVALID_CONTEXT);
-    if (surface->pCurrentFrameSemaphore)
-    {
-        DdiMediaUtil_WaitSemaphore(surface->pCurrentFrameSemaphore);
-        DdiMediaUtil_PostSemaphore(surface->pCurrentFrameSemaphore);
-    }
-
-    // check the bo here?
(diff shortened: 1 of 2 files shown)


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -3664,95 +3664,50 @@ static VAStatus DdiMedia_SyncSurface (
...
-            {
-                return VA_STATUS_ERROR_OPERATION_FAILED;
-            }
-        }
-        else
-        {
-            return VA_STATUS_ERROR_OPERATION_FAILED;
-        }
+    // zero is a expected return value when not hit timeout
+    int64_t timeoutBoWait = timeout_ns >= DDI_MAX_TIMEOUT ? DDI_INFINITE_TIMEOUT : (int64_t)timeout_ns;
```
dmitryermilov, line 3702: Yuchen,
I see you made it the way if value of uint64_t type timeout is  higher than max_vaue(int64_t) you do wait call with infinite. Is there a reason for it instead of calling mos_gem_bo_wait two times, first with max_vaue(int64_t), second with "orig_time_out - max_vaue(int64_t)". 
I don't insist on the 2nd solution, just want to understand objective reasons of following the proposed solution.

media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -3664,95 +3664,50 @@ static VAStatus DdiMedia_SyncSurface (
...
+    auto ret = mos_gem_bo_wait(surface->bo, timeoutBoWait);
```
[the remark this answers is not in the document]
wyc55069407: Yes, it will wait both parts. Because bo_wait will wait all the workloads already submitted to KMD. These workloads reference the bo (R/W)
When called after end picture, both part are submitted to KMD

media_driver/linux/common/ddi/media_libva.cpp
wyc55069407: because this is to align with i915 bo_wait behavior (it's int64_t and when <0, infinite wait)
when the value > max_value(int64_t), the bit 63 will be 1.  When uint64_t max bit is 1, int64_t <0. So actual range is 0-62 bit for timeout.Then bit63 set to 1 means infinite.

We can use option2. But we need firstly allign which value of uint64_t is indicating VA_TIMEOUT_INFINITE   (maybe max(uint64_t) ?)

dmitry-gurulev: >We can use option2. But we need firstly allign which value of uint64_t is indicating VA_TIMEOUT_INFINITE (maybe max(uint64_t) ?)

+1 for this option

wangyan-intel:
LGTM for code. But I have one question: Is it possible to cause dead lock in kernel when use both vaSyncSurface2 and vaSynBuffer? Or application should own buffer/surface scheduling. Thanks.

dmitryermilov:
Hi @wyc55069407 , https://github.com/intel/libva/pull/407 was just merged. Could you update the PR to use VA_TIMEOUT_INFINITE  and re-trigger CI build ?

media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -3753,6 +3726,121 @@ static VAStatus DdiMedia_SyncSurface (
...
+
+    DDI_MEDIA_SURFACE  *surface = DdiMedia_GetSurfaceFromVASurfaceID(mediaCtx, surface_id);
+    DDI_CHK_NULL(surface,    "nullptr surface",      VA_STATUS_ERROR_INVALID_CONTEXT);
+    if (surface->pCurrentFrameSemaphore)
+    {
+        DdiMediaUtil_WaitSemaphore(surface->pCurrentFrameSemaphore);
+        DdiMediaUtil_PostSemaphore(surface->pCurrentFrameSemaphore);
+    }
+    
+    if (timeout_ns == DDI_SYNC_INFINITE_TIMEOUT)
```
dmitryermilov, line 3804: Yuchen, can we use VA_TIMEOUT_INFINITE ( https://github.com/intel/libva/blob/master/va/va.h#L3730 ) and remove DDI_SYNC_INFINITE_TIMEOUT

wyc55069407: Sure, will change.

media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -3626,44 +3626,17 @@ static VAStatus DdiMedia_EndPicture (
...
-    // zero is a expected return value
-    uint32_t timeout_NS = 100000000;
-    while (0 != mos_gem_bo_wait(surface->bo, timeout_NS))
-    {
-        // Just loop while gem_bo_wait times-out.
-    }
-
+	
+    DDI_CHK_NULL(mediaCtx, "nullptr mediaCtx", VA_STATUS_ERROR_INVALID_CONTEXT);
+	DDI_CHK_NULL(surface,  "nullptr surface",  VA_STATUS_ERROR_INVALID_CONTEXT);
```
XinfengZhang, line 3638: align?

PR #944, opened 2020-05-27
[vaPutSurface] do not reuse previous vpp context

xuguangxin:
We usally use vaPutSurface in a renderer thread. It's different than vpp thread.
Reuse the vpp context will introduce a thread race condition issue.


XinfengZhang:
seems it could not avoid race condition, why not create temp vpp context and destory it after execution ? if you want to avoid create/destroy frequently , you should not store it in mediaCtx

xuguangxin:
> seems it could not avoid race condition, why not create temp vpp context and destory it after execution ? if you want to avoid create/destroy frequently , you should not store it in mediaCtx

you mean we should protect PutSurfaceContext ?

xuguangxin:
@XinfengZhang do you think need to protect entire vaPutSurface call with some lock?

Xiaogangli-intel:
@xuguangxin For race condition, we have PutSurfaceRenderMutex/PutSurfaceSwapBufferMutex to protect it. Is that still not enough, did you meet any issue currently?

xuguangxin:
@Xiaogangli-intel , yes, it will reuse the previous vpp context. image following usecase:
decode -> vpp-> renderer. render will use vaPutSurface in other thread. the render thread will race with vpp thread.

xuguangxin:
@XinfengZhang @Xiaogang-Li , protect the vp context creation. please help review it again.

thanks

xuguangxin:
@XinfengZhang @Xiaogang-Li , are we ready to merge this?
thanks

xuguangxin:
@XinfengZhang are we ready to merge this?
thanks

xuguangxin:
@XinfengZhang , any feedback for this?
thanks

xuguangxin:
@XinfengZhang @Xiaogang-Li any more suggestion on this?
thanks

xuguangxin:
@XinfengZhang any more comment for this?
thanks

media_driver/linux/common/ddi/media_libva_putsurface_linux.cpp
```diff
@@ -399,18 +399,19 @@ VAStatus DdiCodec_PutSurfaceLinuxHW(
...
+    if (mediaCtx->PutSurfaceContext == VA_INVALID_ID)
     {
-        return VA_STATUS_ERROR_INVALID_CONTEXT;
+        VAStatus vaStatus = DdiVp_CreateContext(ctx, 0, 0, 0, 0, 0, 0, &mediaCtx->PutSurfaceContext);
+        DdiMediaUtil_UnLockMutex(&mediaCtx->PutSurfaceRenderMutex);
+        DDI_CHK_RET(vaStatus, "Create VP Context failed");
+    } else {
+        DdiMediaUtil_UnLockMutex(&mediaCtx->PutSurfaceRenderMutex);
     }
+    vpCtx = (PDDI_VP_CONTEXT)DdiMedia_GetContextFromContextID(ctx, mediaCtx->PutSurfaceContext, &ctxType);
```
FurongZhang, line 411: Could we move these lines into mutex protection part considering here vphal is used? Thanks.
    DdiMediaUtil_LockMutex(&mediaCtx->PutSurfaceRenderMutex);
    eStatus = vpHal->Render(&renderParams);
    if (MOS_FAILED(eStatus))
    {
        DdiMediaUtil_UnLockMutex(&mediaCtx->PutSurfaceRenderMutex);
        mos_bo_unreference(drawable_bo);
        return VA_STATUS_ERROR_OPERATION_FAILED;
    }

    DdiMediaUtil_UnLockMutex(&mediaCtx->PutSurfaceRenderMutex);

xuguangxin: Hi Furong,
Thanks for the review. It may not needed. 
Above lock just protect us for context creation. Once it created, it only destroy at uninitialized  time. 
The DdiMedia_GetContextFromContextID will hold a lock, and later operation like "vpHal->Render(&renderParams);" is protected by lock too.
So the entire process is thread safe.

PR #945, opened 2020-05-27
vaSyncBuffer driver implementation

wyc55069407:
implementation vaSyncBuffer in driver side with timeout

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -3893,6 +3893,75 @@ static VAStatus DdiMedia_SyncSurface2 (
 }
 
 /*
+ * This function blocks until all pending operations on the buffer have been
+ * completed or exceed timeout.  Upon return it is safe to use the render target for a
+ * different picture.
+ */
+static VAStatus DdiMedia_SyncBuffer (
+    VADriverContextP    ctx,
+    VABufferID          buf_id,
+    uint64_t            timeout_ns
+)
+{
+    PERF_UTILITY_AUTO(__FUNCTION__, "ENCODE", "DDI");
+
+    DDI_FUNCTION_ENTER();
+
+    DDI_CHK_NULL(ctx,    "nullptr ctx",    VA_STATUS_ERROR_INVALID_CONTEXT);
+
+    PDDI_MEDIA_CONTEXT mediaCtx = DdiMedia_GetMediaContext(ctx);
+    DDI_CHK_NULL(mediaCtx,               "nullptr mediaCtx",               VA_STATUS_ERROR_INVALID_CONTEXT);
+    DDI_CHK_NULL(mediaCtx->pBufferHeap,  "nullptr mediaCtx->pBufferHeap",  VA_STATUS_ERROR_INVALID_CONTEXT);
+
+    DDI_CHK_LESS((uint32_t)buf_id, mediaCtx->pBufferHeap->uiAllocatedHeapElements, "Invalid buffer", VA_STATUS_ERROR_INVALID_BUFFER);
+
+    DDI_MEDIA_BUFFER  *buffer = DdiMedia_GetBufferFromVABufferID(mediaCtx, buf_id);
+    DDI_CHK_NULL(buffer,    "nullptr buffer",      VA_STATUS_ERROR_INVALID_CONTEXT);
+
+    if (timeout_ns == VA_TIMEOUT_INFINITE)
+    {
+        // zero is a expected return value when not hit timeout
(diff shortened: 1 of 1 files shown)


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -3755,6 +3755,44 @@ static VAStatus DdiMedia_SyncSurface (
...
+    DDI_CHK_NULL(mediaCtx->pBufferHeap,  "nullptr mediaCtx->pBufferHeap",  VA_STATUS_ERROR_INVALID_CONTEXT);
+
+    DDI_CHK_LESS((uint32_t)buf_id, mediaCtx->pBufferHeap->uiAllocatedHeapElements, "Invalid buffer", VA_STATUS_ERROR_INVALID_BUFFER);
+
+    DDI_MEDIA_BUFFER  *buffer = DdiMedia_GetBufferFromVABufferID(mediaCtx, buf_id);
+    DDI_CHK_NULL(buffer,    "nullptr buffer",      VA_STATUS_ERROR_INVALID_CONTEXT);
+
+    // zero is a expected return value when not hit timeout
+    int64_t timeoutBoWait = timeout_ns >= DDI_MAX_TIMEOUT ? DDI_INFINITE_TIMEOUT : (int64_t)timeout_ns;
+    auto ret = mos_gem_bo_wait(buffer->bo, timeoutBoWait);
```
dmitryermilov, line 3786: Same as in https://github.com/intel/media-driver/pull/943:
Does mos_gem_bo_wait wait for both ENC and PAK parts in case of encode synchronization?

media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -3755,6 +3755,44 @@ static VAStatus DdiMedia_SyncSurface (
...
+    DDI_CHK_NULL(mediaCtx,               "nullptr mediaCtx",               VA_STATUS_ERROR_INVALID_CONTEXT);
+    DDI_CHK_NULL(mediaCtx->pBufferHeap,  "nullptr mediaCtx->pBufferHeap",  VA_STATUS_ERROR_INVALID_CONTEXT);
+
+    DDI_CHK_LESS((uint32_t)buf_id, mediaCtx->pBufferHeap->uiAllocatedHeapElements, "Invalid buffer", VA_STATUS_ERROR_INVALID_BUFFER);
+
+    DDI_MEDIA_BUFFER  *buffer = DdiMedia_GetBufferFromVABufferID(mediaCtx, buf_id);
+    DDI_CHK_NULL(buffer,    "nullptr buffer",      VA_STATUS_ERROR_INVALID_CONTEXT);
+
+    // zero is a expected return value when not hit timeout
+    int64_t timeoutBoWait = timeout_ns >= DDI_MAX_TIMEOUT ? DDI_INFINITE_TIMEOUT : (int64_t)timeout_ns;
```
dmitryermilov, line 3785: Same as in https://github.com/intel/media-driver/pull/943:
Yuchen,
I see you made it the way if value of uint64_t type timeout is higher than max_vaue(int64_t) you do wait call with infinite. Is there a reason for it instead of calling mos_gem_bo_wait two times, first with max_vaue(int64_t), second with "orig_time_out - max_vaue(int64_t)".
I don't insist on the 2nd solution, just want to understand objective reasons of following the proposed solution.

media_driver/linux/common/ddi/media_libva.h
```diff
@@ -117,6 +117,11 @@
 // Enable unlimited output buffer, delete this build option (remove multiple output buffer) when it is verified
 #define ENABLE_ENC_UNLIMITED_OUTPUT
 
+// Max timeout for vaSyncSuraface2 and vaSyncBuffer
+#define DDI_MAX_TIMEOUT       (~(0x8000000000000000))
+// Negative value for infinite timeout
+#define DDI_INFINITE_TIMEOUT  (-1)
```
dmitryermilov, line 123: Same as in https://github.com/intel/media-driver/pull/943: 
propose to extend the comment that -1 is infinite for i915

media_driver/linux/common/ddi/media_libva.cpp
wyc55069407: Answered in Same as in #943

media_driver/linux/common/ddi/media_libva.cpp
wyc55069407: Answered in #943

wangyan-intel:
LGTM for code. But I have one question: Is it possible to cause dead lock in kernel when use both vaSyncSurface2 and vaSynBuffer? Or application should own buffer/surface scheduling. Thanks.

Issue #354, opened 2018-10-10
remove -fno-strict-aliasing from imposed building filags


chivakker:
media-driver imposes -fno-strict-aliasing, which in reality is hiding things like this:

/vphal_common.c:945:19: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
     dwInput   = *((uint32_t *) (&fInput));
                  ~^~~~~~~~~~~~~~~~~~~~~~~

Fix errors when this flag is removed from build.

XinfengZhang:
there are too many variable casting , for different purpose. such as: gpu memory access.

PR #946, opened 2020-05-27
[VP] Remove -fno-strict-aliasing

ljia5:
Remove -fno-strict-aliasing from imposed building filags


dvrogozh:
This PR effectively breaks driver build... What's the reason to remove this flag?

XinfengZhang:
it should because of #354

XinfengZhang:
@ljia5 there are build failure with your change, could you help to check it?

PR #950, opened 2020-05-29
Do not advertise support for nonexistent alpha channels in

weizhu-intel:
decode/encode

Change decode/encode in YUV 4:4:4 formats to only support surfaces with
no alpha channel, since there is no alpha support.

Signed-off-by: Mark Thompson <sw@jkqxz.net>


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -2279,20 +2294,20 @@ DdiMedia_CreateSurfaces2(
             expected_fourcc = VA_FOURCC_Y216;
             break;
         case VA_RT_FORMAT_YUV444:
-            expected_fourcc = VA_FOURCC_444P;
+            expected_fourcc = VA_FOURCC_XYUV;
```
fulinjie, line 2297: What if the decoder wants a VA_FOURCC_444P? 
IIRC, JPEG decoding for 444P seems to require VA_FOURCC_444P. (I could be wrong since it's been some time since my last check)

media_driver/linux/common/ddi/media_libva_caps.cpp
```diff
@@ -2668,23 +2668,23 @@ VAStatus MediaLibvaCaps::QuerySurfaceAttributes(
             attribs[i].type = VASurfaceAttribPixelFormat;
             attribs[i].value.type = VAGenericValueTypeInteger;
             attribs[i].flags = VA_SURFACE_ATTRIB_GETTABLE | VA_SURFACE_ATTRIB_SETTABLE;
-            attribs[i].value.value.i = VA_FOURCC_AYUV;
+            attribs[i].value.value.i = VA_FOURCC_XYUV;
```
fulinjie, line 2671: Directly abandoning AYUV in query would lead to issues in frameworks like gstreamer, since it's already implemented and working currently.

media_driver/linux/common/ddi/media_libva.cpp
weizhu-intel: Thanks linjie's input.
I will use VA_FOURCC_444P as default for yuv444. 
If you want to xyuv, you can use VA_FOURCC_XYUV directly.

media_driver/linux/common/ddi/media_libva_caps.cpp
weizhu-intel: Ok, I will add one more support option for XYUV.

weizhu-intel: Linjie,
What about  Y410 and Y416?  Any framework still using it?

fulinjie: Should be in the same condition IMHO, we'd better add new FORMAT support instead of removing the previous to avoid possible regression from application.

weizhu-intel: Linjie,
With latest version, all previous format is kept, and only new format added. pls help review again.

media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -463,10 +463,16 @@ int32_t DdiMedia_MediaFormatToOsFormat(DDI_MEDIA_FORMAT format)
             return VA_FOURCC_Y216;
         case Media_Format_AYUV:
             return VA_FOURCC_AYUV;
+        case Media_Format_XYUV:
+            return VA_FOURCC_XYUV;
         case Media_Format_Y410:
             return VA_FOURCC_Y410;
+        case Media_Format_XV30:
+            return VA_FOURCC_XV30;
```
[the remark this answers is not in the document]
weizhu-intel: FuRong,
Before our video driver advertise to support alpha channel, but actually not.
So now we remove all alpha  channel support.
XV30 is for Y410,  and XV48 is for Y416.

PR #951, opened 2020-06-01
Fix renderhal and MOS-specific ChromeOS compile errors.

MicheleL-Intel:
mos_os_specific.c: "DestroyIPC(perStreamParameters)" is undefined when IPC
                   is disabled after applying 0002-Disable-IPC-usage.patch

renderhal_gxx.cpp: Result of comparison of constant 8 in expression
                   "pRenderHalSurface->Rotation < 8" will always be true.
                   Replace the constant with a predefined enumerated type.

Signed-off-by: Michele Lim <michele.lim@intel.com>


MicheleL-Intel:
Need to resubmit to remote branch.

PR #957, opened 2020-06-03
Decoder:VP9:GEN{9,10,11,12} : Fix the GPU hang due to wrong SegmentIDStreamIn & UsePrevInFindMvReferences

sreerenjb:
We have been seeing random gpu hang on various devices while decoding 4k
VP9 streams. This patch is addressing a couple of issues which is expected
to fix the hangs:

a) The current code tries to reset SegmentIDStreamIn if there is
a resolution change on inter-frames. But due to the mishandling,
the code was eventually setting SegmentIDStreamIn to true for all
non-scaling use cases. The patch ensures that the reset happens
only for res-change on P-frames irrespective of other conditions.

b) The current implementation sets the SegmentIDStreamIn for any
decode where there is no Segmentation Enabled. We rather follow
strict constraints similar to the legacy i965 implementation.

c) Ensure the correct SegmentIDStreamIn for key-frames, intra-only
frames and error-resilient mode.

d) Reset UsePrevInFindMvReferences to zero if last picture has a different size,
Current picture is error-resilient mode, Last picture was intra_only or keyframe,
Last picture was not a displayed picture.

Thanks to Wang, Yan<yan.wang@intel.com> for the review
and  Li, Steven(steven.li@intel.com) for the suggestion
on keeping the scaling logic outside of the inter-frame code block.

ChromeOS bug id: https://partnerissuetracker.corp.google.com/issues/152554969

### DIFF
--- a/media_driver/agnostic/gen10/hw/vdbox/mhw_vdbox_hcp_g10_X.cpp
+++ b/media_driver/agnostic/gen10/hw/vdbox/mhw_vdbox_hcp_g10_X.cpp
@@ -2258,6 +2258,7 @@ MOS_STATUS MhwVdboxHcpInterfaceG10::AddHcpVp9PicStateCmd(
 
     uint32_t curFrameWidth               = vp9PicParams->FrameWidthMinus1 + 1;
     uint32_t curFrameHeight              = vp9PicParams->FrameHeightMinus1 + 1;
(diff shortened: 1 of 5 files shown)


wangyan-intel:
CI build may have some issue. So I add verifying flag at first.

PR #915, opened 2020-04-22
replace the reconstruct surface to be variant

XinfengZhang:
replace the reconstruct surface to be variant surface for HEVC REXT encoding
it could make the difference between recon/raw to be transparent to application
pay attention , the surface should be not read or write.

Signed-off-by: XinfengZhang <carl.zhang@intel.com>

PR #960, opened 2020-06-04
replace the reconstruct surface to be variant

XinfengZhang:
replace the reconstruct surface to be variant surface for HEVC REXT encoding
it could make the difference between recon/raw to be transparent to application
pay attention , the surface should be not read or write.

it is part of #915  to facilitate the test separately 


XinfengZhang:
@Xiaogangli-intel  it could not be merged until correlated application changes

PR #964, opened 2020-06-09
media_libva_caps_g12: create decoding attributes for SCC main 444 and main 10

xhaihao:
Otherwise user application cannot get the right attributes and will fail
to support SCC main 444 and main 10 decoding


XinfengZhang:
LGTM

PR #966, opened 2020-06-10
media_libva: Fix scc-main444-10 decoding

xhaihao:
VA-API has already defined VAProfileHEVCSccMain444_10 for scc-main444-10,
so we should use VAProfileHEVCSccMain444_10 instead of
VAProfileHEVCSccMain444, otherwise user doesn't know scc-main444-10 is
supported when querying the driver's capabilities


XinfengZhang:
LGTM

PR #969, opened 2020-06-10
[va utils] loop entire allocated heap elements to free resource

xuguangxin:
take DdiMedia_FreeContextHeap for example:
if we allocated 2 contexts, we released first context.The ctxNums is 1.
the loop will quit after elementId > 1, leaks the second context


xuguangxin:
@Xiaogang-Li , any comment on this?
thanks

xuguangxin:
@Xiaogang-Li , are we really to merge this?
thanks

PR #971, opened 2020-06-12
[ENCODE] change slice header prefix for AVC Vdenc

abhishek4kumar:
Currently in case of Vdenc  avc slice header prefix is 0x000001 (3bytes)
and in case of VME it is 0x00000001 (4 bytes) , so change slice header
prefix to 4 bytes for vdenc (same as i965 driver).


abhishek4kumar:
@XinfengZhang  Can you please review  PR , thanks

abhishek4kumar:
@XinfengZhang  can you please review ?

media_driver/agnostic/common/codec/hal/codechal_encode_avc_base.cpp
```diff
@@ -1763,7 +1763,7 @@ MOS_STATUS CodecHalAvcEncode_PackSliceHeader(
     // zero byte shall exist when the byte stream NAL unit syntax structure contains the first
     // NAL unit of an access unit in decoding order, as specified by subclause 7.4.1.2.3.
     // VDEnc Slice header packing handled by PAK does not need the 0 byte inserted
```
XinfengZhang, line 1765: it means you will remove this sentence , right

XinfengZhang:
@aleonenk , please have a look, it changes the VDEnc behavior, will insert one byte 0 into header.

media_driver/agnostic/common/codec/hal/codechal_encode_avc_base.cpp
abhishek4kumar: ok ,i will remove this line

abhishek4kumar:
> @aleonenk , please have a look, it changes the VDEnc behavior, will insert one byte 0 into header.

@aleonenk  can you please review patch ,thanks

thomasli21801:
Hi, just curious, what's the impact of different header?

guanjunz:
0x000001 or 0x00000001 should be fine to decode. From ecoder efficiency perspective, it's ok to keep current logic for VDENC. Did you find any exception?

PR #1000, opened 2020-07-13
Add 3-plane surface support in vaExportSurfaceHandle for non-MMC.

### LINKED ISSUE #995: vaExportSurfaceHandle is broken for YV12 and I420
And probably other three-plane formats. The same offset is returned for the U and V planes.

From a quick look at the code, it seems that `DdiMedia_ExportSurfaceHandle()` does not handle 3 planes at all.

Jister:
Fix #995


michaelolbrich:
Works as expected. Thanks!

PR #1003, opened 2020-07-15
[Encode] Removing 422 RTFormat support from VP9 VDEnc in TGL

jcescamilla:
This change will remove 422 RTFormats support from TGL to fix incorrect surface configuration.


sreerenjb:
@xhaihao @xuguangxin @uartie

sreerenjb:
CC @HeJunyan

uartie:
Yes, according to https://github.com/intel/media-driver/blob/master/docs/media_features.md#supported-encoding-input-format-and-max-resolution, only 420 and 444 is supported.

PR #1008, opened 2020-07-24
media: don't advertise AV1 decoding if the platform doesn't support it

xhaihao:
Otherwise user will be misled by the output of vainfo on Linux

Change-Id: I039fdea492db884a53d24ae8966c619c714e7a96


xhaihao:
@XinfengZhang Any comment for this PR?

XinfengZhang:
LGTM

ZefuLI:
LGTM

PR #1032, opened 2020-08-18
Update surface info of m_destSurface when film grain is enabled

wangyan42164:


wangyan42164:
@ZefuLI Could you please review it? Thanks.

PR #1048, opened 2020-09-16
Fix random corruption issue

### LINKED ISSUE #1028: [regression][ffmpeg-vaapi][vp8d] decode output corrupt
We are observing this issue only on gen 9.5 (KBL/CFL/WHL/CML ...).  However, some of our gen9.5 platforms don't exhibit this problem.  So maybe it's related to SKU or kernel that causes bad driver behavior?

```
$ ffmpeg -hwaccel vaapi -init_hw_device vaapi=hw:/dev/dri/renderD128 \
  -hwaccel_flags allow_profile_mismatch -filter_hw_device hw -v verbose \
  -i ./vp8_rev025_CoeffSkip_10_CapitolRecordsNight0.webm \
  -pix_fmt yuv420p -f rawvideo -vsync passthrough -autoscale 0 \
  -vframes 2 -y threads-default.yuv

$ md5sum threads-default.yuv
318bbd5defb50471c1fe5665f2a70c5a  threads-default.yuv
```

The second frame in threads-default.yuv output is corrupt:

threads-default-frame-2

However, if we add `-threads 1` to the ffmpeg pipeline, then we get the expected output:

```
$ ffmpeg -hwaccel vaapi -init_hw_device vaapi=hw:/dev/dri/renderD128 \
  -hwaccel_flags allow_profile_mismatch -filter_hw_device hw -v verbose \
  -threads 1 -i ./vp8_rev025_CoeffSkip_10_CapitolRecordsNight0.webm \
  -pix_fmt yuv420p -f rawvideo -vsync passthrough -autoscale 0 \
  -vframes 2 -y threads-1.yuv

$ md5sum threads-1.yuv
0da047495680366fc428d120368f35e4  threads-1.yuv
```
threads-1-frame-2

We have many other cases that produce this kind of corrupt behavior, too.

Xiaogangli-intel:
Set struct alignment for DDI_MEDIA_SURFACE and DDI_MEDIA_BUFFER which will fix random corruption issue. Fixed #1028

### DIFF
--- a/media_driver/linux/common/ddi/media_libva_common.h
+++ b/media_driver/linux/common/ddi/media_libva_common.h
@@ -267,6 +267,7 @@ typedef union _DDI_MEDIA_SURFACE_STATUS_REPORT
     } vpp;
 } DDI_MEDIA_SURFACE_STATUS_REPORT, *PDDI_MEDIA_SURFACE_STATUS_REPORT;
 
+#pragma pack(push, 2)
 struct _DDI_MEDIA_BUFFER;
 typedef struct _DDI_MEDIA_SURFACE
 {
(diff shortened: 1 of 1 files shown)


dvrogozh (changes requested):
This looks to be hack WA. How playing with alignments could fix such an issue? This basically means that you have some place in the driver which copies one struct to another via something like memcpy assuming that they have same fields - which is wrong behavior. Or there is place in the driver which somehow looks into struct padding area and makes some conclusions from that - this is wrong as well. In any case it seems that issue is not actually root caused.

@mgonchar, @onabiull, @dmitryermilov : FYI.

Issue #1037, opened 2020-08-28
vaDeriveImage()/vaGetImage() doesn't fill the last column and row in chroma plane if the dimension is odd


rosetta-jpn:
This is same issue of i965 driver filled in https://github.com/intel/intel-vaapi-driver/issues/516.
vaDeriveImage() in media driver has the same issue unlike i965 driver though.
I expect the same gstreamer command can reproduce this issue.
https://github.com/intel/intel-vaapi-driver/issues/516#issuecomment-672166170

sreerenjb:
Issue can be reproduced with slightly different command:
 gst-launch-1.0 videotestsrc num-buffers=90 !  video/x-raw, width=320, height=240 ! vaapipostproc width=321 height=241 ! xvimagesink

sreerenjb:
@XinfengZhang  Can you please find someone to look into this issue?

XinfengZhang:
@FurongZhang , there are VPP call from 320x240 to 321x241. could you help to to check it?
@sreerenjb, @rosetta-jpn  the command line shows it is a VPP issue, but not sure whether we are also discussing odd resolution decoding or encoding? we already have a similiar fix for VP9 encoding  https://github.com/intel/media-driver/commit/d6166b3538a89376771f2769d49d1ade8b177f7e.

rosetta-jpn:
I think decoding should be fine. I remember the video on screen looked fine when I played odd dimension video with iHD driver.

FurongZhang:
May I know why you need to do 320x240 to 321x241? It is only one pixel difference.

sreerenjb:
@FurongZhang one pixel difference is **big** :)
It is not just about 320x240 to 321x241, there are plenty of odd dimension videos out there and because of wrongly implemented vaGetImage (==vpp internally) in the driver, we see green lines in the output.

rosetta-jpn:
Any update on this bug?

FurongZhang:
Basically, we expect surface should be at least 2 aligned. What is your format? As I know, at least for NV12, we can not support odd resolution. What is the justification and impact from your side?

XinfengZhang:
@FurongZhang , I suggest to rounding up the alignment inside driver  ,such as 321x241 to 322 x 242, of course, application still treat it as 321x241 . suppose it will fix the issue .  @rosetta-jpn , does it resolve the issue?

sreerenjb:
[...]

PR #1064, opened 2020-10-14
media_libva_util: Fix odd resoultion handling of yuv420 formats

sreerenjb:
media_libva_util: Fix odd resolution handling of yuv420 formats
    
Ensure 2-byte width & height alignment for I420/YV12/IYUV surfaces
when there is no usage hint set for Encode or VPP write operations.
It seems the  odd resolution surface post-processing operation
(eg: nv12 to i420 conversion via vaGetImage API), causing a green
column on the right side of the video without this alignment.
    
Fixing #1037

### DIFF
--- a/media_driver/linux/common/ddi/media_libva_util.cpp
+++ b/media_driver/linux/common/ddi/media_libva_util.cpp
@@ -208,6 +208,8 @@ VAStatus DdiMediaUtil_AllocateSurface(
                 !(mediaSurface->surfaceUsageHint & VA_SURFACE_ATTRIB_USAGE_HINT_VPP_WRITE))
             {
                 tileformat = I915_TILING_NONE;
+                alignedWidth = MOS_ALIGN_CEIL(width, 2);
+                alignedHeight = MOS_ALIGN_CEIL(width, 2);
                 break;
             }
         case Media_Format_RGBP:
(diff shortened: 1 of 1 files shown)


media_driver/linux/common/ddi/media_libva_util.cpp
```diff
@@ -208,6 +208,7 @@ VAStatus DdiMediaUtil_AllocateSurface(
                 !(mediaSurface->surfaceUsageHint & VA_SURFACE_ATTRIB_USAGE_HINT_VPP_WRITE))
             {
                 tileformat = I915_TILING_NONE;
+                alignedWidth = MOS_ALIGN_CEIL(width, 8);
```
Jister, line 211: I think align to 2 is fine. By the way, height also need to be aligned to 2 here. Also, line 423 should be removed.
https://github.com/intel/media-driver/blob/fd9e510001b3151b8061a038fa4a7429fd9c804c/media_driver/linux/common/ddi/media_libva_util.cpp#L423

sreerenjb: Updated with some comment changes. Thanks.

sreerenjb:
@XinfengZhang Can you please merge this?

PR #1079, opened 2020-11-06
Break driver build into parts C/SSE2/SSE4

egorovdanil:
Fixes intel/libva#466


media_driver/cmake/linux/media_compile_flags_linux.cmake
```diff
@@ -38,29 +38,13 @@ set(MEDIA_COMPILER_FLAGS_COMMON
     -Werror=return-type
 
     # General optimization options
-    -march=${UFO_MARCH}
     -mpopcnt
```
sebastinas, line 41: popcnt is only available with SSE4.2, so please also drop that flag.

dvrogozh: yep. thank's for the catch. @egorovdanil : same idea - try to drop and check whether we use these instructions anywhere in the code.

@sebastinas : mind that this PR is only the first step. We likely need similar change in gmmlib.

egorovdanil: done

shengcon (commented):
This patch is almost identical to https://github.com/intel-innersource/drivers.gpu.unified/pull/2690.
Why are there 2 pull requests?

dvrogozh:
This PR required some internal follow up. We actually merged in the fix, it is available in master branch as https://github.com/intel/media-driver/commit/ff55f781c91509ad6b6d21cd357346656066740b. Fix available in releases >= `intel-media-21.1.0`.

Issue #1072, opened 2020-10-28
[regression][ffmpeg-vaapi] VPP scaling corrupts UV channel


uartie:
Since commit 450fcc06c0eb4a61a2f5f73b3e07e0e875a484d6 (#1064):

```
commit 450fcc06c0eb4a61a2f5f73b3e07e0e875a484d6
Author: Sreerenj Balachandran <sreerenj.balachandran@intel.com>
Date:   Tue Oct 13 18:38:00 2020 -0700

    media_libva_util: Fix odd resolution handling of yuv420 formats
    
    Ensure 2-byte width & height alignment for I420/YV12/IYUV surfaces
    when there is no usage hint set for Encode or VPP write operations.
    It seems the  odd resolution surface post-processing operation
    (eg: nv12 to i420 conversion via vaGetImage API), causing a green
    column on the right side of the video without this alignment.
    
    Fixing #1037
```

Scaling with ffmpeg-vaapi produces corrupted UV channel

```
ffmpeg -hwaccel vaapi -vaapi_device /dev/dri/renderD128 -v verbose \
  -f lavfi -i smptebars=size=1920x1080 \
  -vf 'format=yuv420p,hwupload,scale_vaapi=w=1280:h=720,hwdownload,format=yuv420p' \
  -pix_fmt yuv420p -f rawvideo -vsync passthrough -an \
  -vframes 1 -y ffmpeg-vaapi_1280x720_I420.yuv
```

Actual Result:
<p><img src="https://user-images.githubusercontent.com/677859/97484081-81f11480-1915-11eb-82e4-d106a0102af1.jpg" width="25%"/></p>

Expected Result:
<p><img src="https://user-images.githubusercontent.com/677859/97484091-861d3200-1915-11eb-9e84-36e2d715b950.jpg" width="25%"/></p>

cc: @sreerenjb

uartie:
related to #1071

uartie:
reverted in ca794cb70bdd5faf134324310b7054f240dd77f9 and fixes this issue.

Issue #1071, opened 2020-10-28
[regression][gstreamer] VPP rotation 90/270 YV12 corruption


uartie:
Since commit 450fcc06c0eb4a61a2f5f73b3e07e0e875a484d6 (#1064):

```
commit 450fcc06c0eb4a61a2f5f73b3e07e0e875a484d6
Author: Sreerenj Balachandran <sreerenj.balachandran@intel.com>
Date:   Tue Oct 13 18:38:00 2020 -0700

    media_libva_util: Fix odd resolution handling of yuv420 formats
    
    Ensure 2-byte width & height alignment for I420/YV12/IYUV surfaces
    when there is no usage hint set for Encode or VPP write operations.
    It seems the  odd resolution surface post-processing operation
    (eg: nv12 to i420 conversion via vaGetImage API), causing a green
    column on the right side of the video without this alignment.
    
    Fixing #1037
```

YV12 rotation at 90 or 270 degress with vaapipostproc and msdkvpp produces corrupted output.

```
gst-launch-1.0 -vf videotestsrc num-buffers=1 ! video/x-raw,format=YV12,width=1920,height=1080 \
  ! vaapipostproc video-direction=90r ! video/x-raw,format=YV12 \
  ! checksumsink2 file-checksum=false qos=false frame-checksum=false \
  plane-checksum=false dump-output=true dump-location=gst-vaapi-1080x1920_YV12.yuv

gst-launch-1.0 -vf videotestsrc num-buffers=1 ! video/x-raw,format=YV12,width=1920,height=1080 \
  ! msdkvpp video-direction=90r ! video/x-raw,format=YV12 \
  ! checksumsink2 file-checksum=false qos=false frame-checksum=false \
  plane-checksum=false dump-output=true dump-location=gst-msdk-1080x1920_YV12.yuv
```

Actual Result:
<p><img src="https://user-images.githubusercontent.com/677859/97481833-9a136480-1912-11eb-875d-928a9a9430f8.jpg" width="25%"/></p>

Expected Result:
<p><img src="https://user-images.githubusercontent.com/677859/97481887-ad263480-1912-11eb-8d8a-d66044b24157.jpg" width="25%"/></p>

cc: @sreerenjb

sreerenjb:
Ouch! we should revert immediately then.
@Jister @XinfengZhang 

Quickly tested with xvimagesink but didn't see any issue. Probably the conversion involved in the middle is hiding the actual issue.
[...]

PR #1088, opened 2020-11-19
Fix odd resolution handling of I420/YV12 formats

### LINKED ISSUE #1037: vaDeriveImage()/vaGetImage() doesn't fill the last column and row in chroma plane if the dimension is odd

Jister:
Fix #1037

### DIFF
--- a/media_driver/linux/common/ddi/media_libva_util.cpp
+++ b/media_driver/linux/common/ddi/media_libva_util.cpp
@@ -220,6 +220,8 @@ VAStatus DdiMediaUtil_AllocateSurface(
                 !(mediaSurface->surfaceUsageHint & VA_SURFACE_ATTRIB_USAGE_HINT_VPP_WRITE))
             {
                 tileformat = I915_TILING_NONE;
+                alignedWidth = MOS_ALIGN_CEIL(width, 2);
+                alignedHeight = MOS_ALIGN_CEIL(height, 2);
                 break;
             }
         case Media_Format_RGBP:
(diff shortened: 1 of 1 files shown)


FurongZhang:
I don't have any concern for this patch. Just need to run regression test to make sure no regression.

XinfengZhang:
LGTM @sreerenjb , how about this patch?

media_driver/linux/common/ddi/media_libva_util.cpp
```diff
@@ -208,6 +208,8 @@ VAStatus DdiMediaUtil_AllocateSurface(
                 !(mediaSurface->surfaceUsageHint & VA_SURFACE_ATTRIB_USAGE_HINT_VPP_WRITE))
             {
                 tileformat = I915_TILING_NONE;
+                alignedWidth = MOS_ALIGN_CEIL(width, 2);
+                alignedHeight = MOS_ALIGN_CEIL(width, 2);
```
uartie, line 212: should this be `MOS_ALIGN_CEIL(height, 2)`?

sreerenjb:
@XinfengZhang 
As @uartie mentioned, as long as it is not causing #1071 and #1072 , I'm okay with the patch.

media_driver/linux/common/ddi/media_libva_util.cpp
sreerenjb: It should be :)

Jister:
Verified locally that the case show no corruption with latest patch. Sorry for the typo in previous patch.

PR #1094, opened 2020-11-28
Encode - Display assert message before returning error

sushma08:
While checking igd logs for encode failure debug, normal message was
printed and then VA_STATUS_ERROR_ENCODING_ERROR was returned, so it was
misleading. Critical failure before returning error will be good.

Signed-off-by: Sushma Venkatesh Reddy <sushma08reddy@gmail.com>


XinfengZhang:
LGTM

PR #1098, opened 2020-12-03
MosUtils: free the key list when UserFeatureDumpFile returns error

xuguangxin:


xuguangxin:
@XinfengZhang any suggestion on this.
thanks

xuguangxin:
@XinfengZhang are we good to go?
This is a important issue for cfl, without this fix the unit test will failed on cfl. and we can't do make install. 
and "gst-launch-1.0 videotestsrc ! msdkvpp ! fakesink" will running out of memory on cfl.

XinfengZhang:
hi @xuguangxin , we are refine registry management code to avoid such issue, after refine, register will be classed by release type and call frequency, @Xiaogang-Li are working on it, so, if you want a quick fix, we could help to file an PR, but there will be fix in near future

xuguangxin:
but we still need free the memory after the  UserFeatureDumpFile return error, right?

thanks

XinfengZhang (approved):

xuguangxin:
this fixed by 196cfe6a7bdc31da0e3e07d420d9146487ca5ab1 too. 
It's no need merge this patch. 
thanks carl for help on on this

PR #1144, opened 2021-02-19
[Decode] Add 0YUV format for decode

wangqia6:
Support 0YUV format from decode side:
XYUV(8bit)


media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -607,6 +609,8 @@ DDI_MEDIA_FORMAT DdiMedia_OsFormatToMediaFormat(int32_t fourcc, int32_t rtformat
             return Media_Format_Y216;
         case VA_FOURCC_AYUV:
             return Media_Format_AYUV;
+        case VA_FOURCC_XYUV:
+            return VA_FOURCC_XYUV;
```
xhaihao, line 613: Here Media_Format_XYUV should be returned.

PR #1150, opened 2021-02-26
Wayne 2 26

weizhu-intel:
fix  issues from asan


XinfengZhang (changes requested):
you should not destroy buffers of other context in vaDestoryContext.

PR #1154, opened 2021-03-05
[Encode]: Fix incorrect MBQP map for HEVC VDENC.

ZhongXiaoxia:
Map the QP value for each macro block in one LCU instead of sharing the same value.

Signed-off-by: Zhong, Xiaoxia <xiaoxia.zhong@intel.com>


media_driver/agnostic/common/codec/hal/codechal_vdenc_hevc.cpp
```diff
@@ -796,14 +799,18 @@ MOS_STATUS CodechalVdencHevcState::SetupMbQpStreamIn(PMOS_RESOURCE streamIn)
...
-            int32_t ForceQp = pInputData[(h/2) * m_encodeParams.psMbQpDataSurface->dwPitch + (w/2)];
             //            (pData + (Offset + XOffset + YOffset))->DW7.QpEnable = 0xf;
             //            (pData + (Offset + XOffset + YOffset))->DW14.ForceQp_0 = ForceQp;
             //            (pData + (Offset + XOffset + YOffset))->DW14.ForceQp_1 = ForceQp;
             //            (pData + (Offset + XOffset + YOffset))->DW14.ForceQp_2 = ForceQp;
             //            (pData + (Offset + XOffset + YOffset))->DW14.ForceQp_3 = ForceQp;
             streaminDataParams.setQpRoiCtrl = true;
-            streaminDataParams.forceQp = (int8_t)ForceQp;
+
+            streaminDataParams.forceQp[0] = (int8_t) ( pInputData[(h * 2)     * m_encodeParams.psMbQpDataSurface->dwPitch + (w * 2)]);
```
thomasli21801, line 809: Hi Xiaoxia,
May I know what's the block size of input MbQpDataSurface?
Is the MB QP map in 16x16 block or 32x32 block or other size?

Thanks,
Thomas

thomasli21801: I'm asking because this change may be conflict with HW design and validation results. StreamIn surface 4 QP values need to be the same according to bspec.

ZhongXiaoxia: > Hi Xiaoxia,
> May I know what's the block size of input MbQpDataSurface?
> Is the MB QP map in 16x16 block or 32x32 block or other size?
> 
> Thanks,
> Thomas

Hi Thomas:
The block size is 16x16 for HEVC VDENC, which is defined in MSDK. As we tested here on SG1, it can work as expected. Each 16x16 block can be set to specified QP value. Is it possible the HW design limitation has already been removed or resolved for Gen12?

media_driver/agnostic/common/codec/hal/codechal_vdenc_hevc.cpp
```diff
@@ -796,14 +799,18 @@ MOS_STATUS CodechalVdencHevcState::SetupMbQpStreamIn(PMOS_RESOURCE streamIn)
...
+            streaminDataParams.forceQp[1] = (int8_t) ( pInputData[(h * 2)     * m_encodeParams.psMbQpDataSurface->dwPitch + (w * 2 + 1)]);
```
zxye, line 810: Hi Xiaoxia, reading pInputData four times in this loop will significantly increase the CPU utilization and impact the performance.

ZhongXiaoxia: > Hi Xiaoxia, reading pInputData four times in this loop will significantly increase the CPU utilization and impact the performance.

Yes, it will. But for 16x16 CU QP, it has to read four QP values for each 16x16 cu block from the QP map table. Do you have suggestion to improve it?

zxye: Copy the buffer to system memory before the loop can get lower CPU utilization and better performance.

ZhongXiaoxia: Thank you for your suggestion, Tony. I have updated the code as you said. Please check the change in below new patch.

zxye: Thanks for the update, Xiaoxia. The change looks good to me.

zxye (approved):

PR #1160, opened 2021-03-11
Stop using PWRITE

jekstrand:
We're attempting to remove PWRITE support from i915 for future platforms and this prepares the media driver for the move.

My attempt at media driver hacking looks correct to me but I've only compile-tested it.

### DIFF
--- a/media_driver/linux/common/codec/ddi/media_ddi_decode_base.cpp
+++ b/media_driver/linux/common/codec/ddi/media_ddi_decode_base.cpp
@@ -1096,13 +1096,10 @@ VAStatus DdiMediaDecode::CreateBuffer(
 
     if( true == buf->bCFlushReq )
     {
-        mos_bo_subdata(buf->bo, buf->uiOffset, size * numElements, data);
+        mos_bo_wait_rendering(buf->bo);
     }
-    else
-    {
-        status = MOS_SecureMemcpy((void *)(buf->pData + buf->uiOffset), size * numElements, data, size * numElements);
-        DDI_CHK_CONDITION((status != MOS_STATUS_SUCCESS), "DDI:Failed to copy buffer data!", VA_STATUS_ERROR_OPERATION_FAILED);
-    }
+    status = MOS_SecureMemcpy((void *)(buf->pData + buf->uiOffset), size * numElements, data, size * numElements);
+    DDI_CHK_CONDITION((status != MOS_STATUS_SUCCESS), "DDI:Failed to copy buffer data!", VA_STATUS_ERROR_OPERATION_FAILED);
     return va;
 
 CleanUpandReturn:
--- a/media_driver/linux/common/codec/ddi/media_ddi_decode_vp8.cpp
+++ b/media_driver/linux/common/codec/ddi/media_ddi_decode_vp8.cpp
@@ -264,21 +264,12 @@ VAStatus DdiDecodeVP8::ParseProbabilityData(
 {
     if (vp8ProbDataBuff->pData && probInputBuf)
     {
-        MOS_LINUX_BO *boDes = nullptr;
-        boDes               = vp8ProbDataBuff->bo;
-
-        mos_bo_wait_rendering(boDes);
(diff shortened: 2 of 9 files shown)


Jister (approved):

media_driver/linux/common/codec/ddi/media_ddi_decode_vp8.cpp
```diff
@@ -264,21 +264,12 @@ VAStatus DdiDecodeVP8::ParseProbabilityData(
...
-                     sizeof(CODECHAL_VP8_COEFFPROB_DATA),
-                     probInputBuf))
-        {
-            return VA_STATUS_SUCCESS;
-        }
-        else
-        {
-            return VA_STATUS_ERROR_INVALID_PARAMETER;
-        }
+        mos_bo_wait_rendering(vp8ProbDataBuff->bo);
```
XinfengZhang, line 267: if there are no a bo_map , vp8ProbDataBuff->pData should be NULL, should we add some protection or just add bo_map before memory copy?

media_driver/linux/common/codec/ddi/media_ddi_decode_base.cpp
```diff
@@ -1096,13 +1096,10 @@ VAStatus DdiMediaDecode::CreateBuffer(
...
     {
-        mos_bo_subdata(buf->bo, buf->uiOffset, size * numElements, data);
-    }
-    else
-    {
-        status = MOS_SecureMemcpy((void *)(buf->pData + buf->uiOffset), size * numElements, data, size * numElements);
-        DDI_CHK_CONDITION((status != MOS_STATUS_SUCCESS), "DDI:Failed to copy buffer data!", VA_STATUS_ERROR_OPERATION_FAILED);
+        mos_bo_wait_rendering(buf->bo);
     }
+    status = MOS_SecureMemcpy((void *)(buf->pData + buf->uiOffset), size * numElements, data, size * numElements);
```
XinfengZhang, line 1101: a bo_map before memory copy?

XinfengZhang (changes requested):

media_driver/linux/common/codec/ddi/media_ddi_decode_vp8.cpp
jekstrand: This is already guarded by
```
    if (vp8ProbDataBuff->pData && probInputBuf)
```
if you look a couple lines up in the file.  It should be fine.

media_driver/linux/common/codec/ddi/media_ddi_decode_base.cpp
jekstrand: I can't tell what the right thing is to do here.  If it really is as simple as dropping in a `mos_bo_map()` right after the `mos_bo_wait_rendering()`, I can do that.  That said, I think this one is actually ok too.  It's guarded by `if(true == buf->bCFlushReq)` and the one place that ever sets `bCFlushReq = true` does so right after setting `pData`.

That said, I'm happy to do whatever you want.  I'm coding entirely blind here; I've never touched the media driver before so I don't know any of your conventions.

XinfengZhang: CFlushReq should be used for the sku which has no LLC, but in this case, they used to distinguish whether the bistream buffer is swapped into system memory or gpu memory.  and yes, it is already mapped.

media_driver/linux/common/codec/ddi/media_ddi_decode_vp8.cpp
XinfengZhang: yes, it already been checked, thanks

XinfengZhang (approved):
thanks , I have no other concern.

media_driver/linux/common/codec/ddi/media_ddi_decode_base.cpp
zxye: looks good to me.

zxye (approved):
Looks fine.

PR #1203, opened 2021-06-05
cmrtlib/linux/CMakelists.txt: respect MEDIA_BUILD_FATAL_WARNINGS

ffontaine:
Respect `MEDIA_BUILD_FATAL_WARNINGS` to avoid the following build failure when the user provides `_FORTIFY_SOURCE`:

```
In file included from /home/buildroot/autobuild/instance-0/output-1/host/x86_64-buildroot-linux-gnu/sysroot/usr/include/dlfcn.h:22,
                 from /home/buildroot/autobuild/instance-0/output-1/build/intel-mediadriver-19.4.0r/cmrtlib/linux/../linux/share/cm_include.h:30,
                 from /home/buildroot/autobuild/instance-0/output-1/build/intel-mediadriver-19.4.0r/cmrtlib/agnostic/share/cm_printf_host.cpp:23:
/home/buildroot/autobuild/instance-0/output-1/host/x86_64-buildroot-linux-gnu/sysroot/usr/include/features.h:397:4: error: #warning _FORTIFY_SOURCE requires compiling with optimization (-O) [-Werror=cpp]
  397 | #  warning _FORTIFY_SOURCE requires compiling with optimization (-O)
      |    ^~~~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/52638d95312e464626d1c4047b3b26d4f57a1cd2

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/cmrtlib/linux/CMakeLists.txt
+++ b/cmrtlib/linux/CMakeLists.txt
@@ -32,12 +32,16 @@ else()
 endif()
 
 # Set up compile options that will be used for the Linux build
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPP_STANDARD_OPTION} -fPIC -fpermissive -fstack-protector-all -Werror")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPP_STANDARD_OPTION} -fPIC -fpermissive -fstack-protector-all")
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -fno-strict-aliasing -D_FORTIFY_SOURCE=2")
 set(CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS_DEBUG} -D_DEBUG -D__DEBUG -O0")
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${CPP_STANDARD_OPTION} -fPIC -fpermissive -fstack-protector-all -Werror")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${CPP_STANDARD_OPTION} -fPIC -fpermissive -fstack-protector-all")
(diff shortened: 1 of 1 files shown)


XinfengZhang:
from my understanding, we should resolve warning instead of remove the Werror option. could you help to explain why we could not fix the warning directly? 
@HeFan2017 could you help to review it.

cmrtlib/linux/CMakeLists.txt
```diff
@@ -32,12 +32,16 @@ else()
 endif()
 
 # Set up compile options that will be used for the Linux build
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPP_STANDARD_OPTION} -fPIC -fpermissive -fstack-protector-all -Werror")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPP_STANDARD_OPTION} -fPIC -fpermissive -fstack-protector-all")
```
FurongZhang, line 35: Why need to remove -Werror instead of fixing in the code.

dvrogozh (approved):
Warning should be fixed, no doubt.

However, I fully agree that having `-Werror` by default is a bad practice. Instead, in our CI we should add `-Werror` passing custom build flags, something like:
```
cmake -DCMAKE_C_FLAGS="-Werror" ...
```
This advise goes to majority of the build flags by the way. Imho, cmake setup should be minimal and if we have any additional requirements these should be handled in CI setup, not in the cmake setup.

As of now media-drviver does not follow that, but it has `MEDIA_BUILD_FATAL_WARNINGS` cmake variable to allow disabling of `-Werror`. By default, we force warnings as errors. That's good compromise. So, I believe this PR is a good one since it actually won't affect our CI builds, but allow community make experiments if needed (on their own risk). So, I approve the request.

ffontaine:
Thanks for your feedback, it should also be noted that we're carrying a patch in buildroot to drop hardening flags because some embedded toolchains don't support it: https://git.buildroot.net/buildroot/tree/package/intel-mediadriver/0001-Drop-hardening-related-flags.patch. It would be great if those kind of flags could also be disabled through a dedicated option even if stack-protection should be enabled by default for obvious security reasons.

XinfengZhang (approved):

PR #1212, opened 2021-06-29
[Decode][Gen12] Provide mem_types when querying surface attributes

zhangyuankun-star:
Provide mem_types VA_SURFACE_ATTRIB_MEM_TYPE_VA and VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2
for decode and encode when querying surface attributes on Gen12.

Signed-off-by: Yuankun Zhang <yuankunx.zhang@intel.com>

### DIFF
--- a/media_driver/linux/gen12/ddi/media_libva_caps_g12.cpp
+++ b/media_driver/linux/gen12/ddi/media_libva_caps_g12.cpp
@@ -1183,6 +1183,13 @@ VAStatus MediaLibvaCapsG12::AddEncSurfaceAttributes(
             attribList[numAttribs].value.value.i = m_encJpegMinHeight;
         }
         numAttribs++;
+
+        attribList[numAttribs].type = VASurfaceAttribMemoryType;
+        attribList[numAttribs].value.type = VAGenericValueTypeInteger;
+        attribList[numAttribs].flags = VA_SURFACE_ATTRIB_GETTABLE | VA_SURFACE_ATTRIB_SETTABLE;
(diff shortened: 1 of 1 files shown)


HeJunyan:
@XinfengZhang , please help. Or, we can not use VA and DMA in TGL.

XinfengZhang:
I think you just need VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME, please dont add other three items, because Decode does not support user_ptr or linear surface.

HeJunyan:
I think we need VA_SURFACE_ATTRIB_MEM_TYPE_VA and VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2

zhangyuankun-star:
@XinfengZhang , I think encode should add this part ,too. So I modified the encode part as other platform(gen11,common) does.
Did I do anything wrong? Can you review it again for me?  thanks.

XinfengZhang:
for decode: intel media driver just support tile surface as output surface, so it should be PRIME2, (VA is used to create linear surface by default )
for encode, there are different support for linear surface from gen12, for example: linear surface is supported for HEVC VDEnc path.

HeJunyan:
VASurfaceAttribMemoryType query means the supported surface type. VA_SURFACE_ATTRIB_MEM_TYPE_VA  means the generic VA surface, we do not care about its tiling, it is maintained by the VA driver.  VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2 means the surface that we can use vaExportSurfaceHandle to acquire its FD and tiling.

I think it is not related to liner or tiling here.

XinfengZhang (approved):

PR #1222, opened 2021-07-13
Added the offset back for Qp report on non-8 bit depth encode.

hongbolv:


FurongZhang (approved):

media_driver/agnostic/common/codec/hal/codechal_encode_hevc_base.cpp
```diff
@@ -2235,8 +2235,8 @@ MOS_STATUS CodechalEncodeHevcBase::GetStatusReport(
         encodeStatusReport->QpY = encodeStatusReport->AverageQp =
             (uint8_t)(((uint32_t)encodeStatus->QpStatusCount.hcpCumulativeQP)
```
walter-bai, line 2236: 1. From spec, for 10 bit case, QP is from -12 to 51, but only if each QP accumulated in hcpCumulativeQP is in the range [0, 51+12] can "- 6 * m_hevcSeqParams->bit_depth_luma_minus8" be correct. Since BSpec gives no clue, we need to confirm with HW about this.
2. (1 << log2LcuSize)) >> log2LcuSize) is always 1, but as we confirmed before, hcpCumulativeQP is the sum of QP value per 4x4-pixel-block.
3. since QP can be negative, do we need to redefine AverageQp to int16_t or something like that?

walter-bai (approved):

PR #1239, opened 2021-08-06
Add a function to query hardware configuration information from KMD.

zxye:


media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4431,8 +4431,194 @@ mos_gem_bo_is_exec_object_async(struct mos_linux_bo *bo)
...
+
+#define BIT(x) (1ul <<(x))
+enum {
+    INTEL_HWCONFIG_CACHE_TYPE_L3    = BIT(0),
+    INTEL_HWCONFIG_CACHE_TYPE_LLC   = BIT(1),
+    INTEL_HWCONFIG_CACHE_TYPE_EDRAM = BIT(2),
+};
+
+#ifndef DRM_I915_QUERY_HWCONFIG_TABLE
+#define DRM_I915_QUERY_HWCONFIG_TABLE   5
```
dvrogozh, line 4531: This  seems to be some part of i915 API. Is it? if yes, why it's defined here, but no in i915 uAPI?

zxye: This is just to make the build pass before the uAPI is merged. Will remove it when i915 uAPI is merged.

dvrogozh: I see. So, now I can set -1 for this PR :).

dvrogozh (changes requested):

media_driver/linux/common/os/i915/mos_bufmgr.c
johnharr-intel: There will be no UAPI header. The view from architecture is that it must be possible to provide updates to the table without any kernel side changes. Due to the very long lead time of getting a kernel change through a linux distro release cycle. Therefore, each UMD must provide their own local implementation of these definitions.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4431,8 +4431,194 @@ mos_gem_bo_is_exec_object_async(struct mos_linux_bo *bo)
...
+    ulength = query_item.length / sizeof(uint32_t);
+    i = 0;
+    assert(gfx_info);
+    while (i < ulength) {
+        /* Attribute ID starts with 1 */
+        assert(hw_info[i] > 0);
+
+        if (INTEL_HWCONFIG_MAX_SLICES_SUPPORTED == hw_info[i])
+        {
+            assert(hw_info[i+1] == 1);
```
XinfengZhang, line 4585: about these assert, the information size will never change in future for each configure?

johnharr-intel: The number of bytes used by the MAX_SLICES_SUPPORTED field is fixed and cannot ever change. The value contained within those bytes will (potentially) be different on every platform.

Also, why use a long series of 'if' statements instead of a 'switch' statement?

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: I separated the change into two patches. The first one will update the uapi header to define DRM_I915_QUERY_HWCONFIG_TABLE . We will use the header file generated from kernel source tree once the i915 patch is merged. The second patch is the implementation of the query function.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: Agree, switch is better. I have updated the patch. Thanks.

XinfengZhang (approved):

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4431,8 +4431,194 @@ mos_gem_bo_is_exec_object_async(struct mos_linux_bo *bo)
...
+#ifndef DRM_I915_QUERY_HWCONFIG_TABLE
+#define DRM_I915_QUERY_HWCONFIG_TABLE   5
+#endif
+
+static int mos_query_items(int fd, struct drm_i915_query_item *query_item)
+{
+    struct drm_i915_query query;
+    int ret;
+
+    assert(query_item);
```
FurongZhang, line 4539: could we use assert? if release or release internal version, usually crash instead of assert. Or i am wrong? thank you! Just for my education purpose.

media_driver/linux/common/os/i915/mos_bufmgr.c
```diff
@@ -4431,8 +4431,194 @@ mos_gem_bo_is_exec_object_async(struct mos_linux_bo *bo)
     return bo_gem->exec_async;
 }
 
+/**
+ * enum intel_hwconfig - Global definition of hwconfig table attributes
+ *
+ * Intel devices provide a KLV (Key/Length/Value) table containing
```
FurongZhang, line 4437: May I know the memory layout of Key, Length and Value? 

I am trying to understand the below logic. 

            assert(hw_info[i+1] == 1);
            gfx_info->SubSliceCount = hw_info[i+2];
            gfx_info->MaxSubSlicesSupported = hw_info[i+2];

Thanks a lot in advance!

zxye: Hi Furong, the layout is,
K, L, V,
K, L, V,
...
For now, all the K/L/V are one unit32_t. So the L (hw_info[i+1]) is always 1.

zxye: > 

Hi Furong, thanks for the review. The layout is,
K, L, V,
K, L, V,
...
For now, all the K/L/V are one unit32_t. So the L (hw_info[i+1]) is always 1.

media_driver/linux/common/os/i915/mos_bufmgr.c
zxye: Not sure if I understood your point. We are just following the convention of mos_bufmgr functions to use assert here to make sure query_item is a valid ptr.

Issue #1233, opened 2021-08-02
Dependency vs Prerequisite


daravi:
Might be a bit of a dumb question, but what exactly is the difference between a dependency and a prerequisite in README.md?

Sherry-Lin:
The "prerequisite" is the environment preparison for third party open source packages. The "dependency" focus on Intel media stack. To avoid confuse, I will update the it more clearly.

PR #1244, opened 2021-08-13
Readme update for building

Sherry-Lin:
To simplify the building step to combine prerequiresite and dependency together.

Fixes: #1233


README.md
```diff
@@ -23,15 +22,8 @@ apt install autoconf libtool libdrm-dev xorg xorg-dev openbox libx11-dev libgl1-
...
-
-* [Libva](https://github.com/intel/libva)
-* [GmmLib](https://github.com/intel/gmmlib)(Please check [comparability with GmmLib](https://github.com/intel/media-driver/wiki/Comparability-with-GmmLib))
-
-## Building
-
-1. Build and install libva master
-2. Build and install gmmlib master
+1. Build and install libva master from https://github.com/intel/libva
+2. Build and install gmmlib master from https://github.com/intel/gmmlib(Please check [comparability with GmmLib](https://github.com/intel/media-driver/wiki/Comparability-with-GmmLib))
```
dvrogozh, line 26: "comparability" ?! what is this:)?

You likely wanted to say "Compatibility" instead. And you need to add a space between `https://github.com/intel/gmmlib(Please`.

As for the https://github.com/intel/media-driver/wiki/Comparability-with-GmmLib, it needs to be refined. It refers some rough commits in gmmlib, but can we highlight in which gmm releases these commits present?

dvrogozh (changes requested):

README.md
Sherry-Lin: Thanks for correction. I will update in PR soon.
The incompatible gmmlib commits are at very beginning revision(first 10-15th commits) so all of them are before first tag intel-gmmlib-18.3.pre1. I'm thinking whether it's valuable to mention it or not as most of user will pull gmmlib tag version or latest one directly.

dvrogozh: I think it's not valuable then. If commits were somewhere in the middle of the history, between tagged releases, then this would be valuable. I think we need to exclude this link from readme. Wiki page can remain however for future need.

Sherry-Lin: Yes, agree. updated.

Sherry-Lin:
@dvrogozh any other comments for this PR? Is it good to be merge?

FurongZhang (approved):

Issue #1259, opened 2021-09-07
[regression] libva conformance failures since #1229


uartie:
## System information
- Impacted GPU platform:  <= Gen9 (CML, CFL, KBL, SKL,...)
- Host machine if it's descreted card: NA
- Good libva/libva-utils/gmmlib/media-driver version? 91cf6de96b87c30c2a4d340897ee967963812478
- Bad libva/libva-utils/gmmlib/media-driver version? 4ce77a648065205711f16dccd374ab83464aa911

### Describe the current behavior
Since PR #1229, libva conformance tests fail on <= gen9 platform (passes on gen11+)...

```
commit 4ce77a648065205711f16dccd374ab83464aa911
Author: Yuankun Zhang <yuankunx.zhang@intel.com>
Date:   Thu Jul 22 10:44:44 2021 +0800

    [VP] Fix loss of Y210 and Y410 format in vaQuerySurfaceAttributes.
    
    Fix loss of Y210 and Y410 format in vaQuerySurfaceAttributes.
    
    Signed-off-by: Yuankun Zhang <yuankunx.zhang@intel.com>
```

```
$ test_va_api
<snip>
==========] 17415 tests from 14 test cases ran. (33675 ms total)
[  PASSED  ] 17381 tests.
[  FAILED  ] 34 tests, listed below:
[  FAILED  ] CreateSurfaces/VAAPICreateSurfaces.CreateSurfacesWithConfigAttribs/136, where GetParam() = (-1:VAProfileNone, 10:VAEntrypointVideoProc, 16x16)
[  FAILED  ] CreateSurfaces/VAAPICreateSurfaces.CreateSurfacesWithConfigAttribs/137, where GetParam() = (-1:VAProfileNone, 10:VAEntrypointVideoProc, 32x32)
[  FAILED  ] CreateSurfaces/VAAPICreateSurfaces.CreateSurfacesWithConfigAttribs/138, where GetParam() = (-1:VAProfileNone, 10:VAEntrypointVideoProc, 176x120)
[  FAILED  ] CreateSurfaces/VAAPICreateSurfaces.CreateSurfacesWithConfigAttribs/139, where GetParam() = (-1:VAProfileNone, 10:VAEntrypointVideoProc, 352x240)
[  FAILED  ] CreateSurfaces/VAAPICreateSurfaces.CreateSurfacesWithConfigAttribs/140, where GetParam() = (-1:VAProfileNone, 10:VAEntrypointVideoProc, 352x288)
[...]

PR #1229, opened 2021-07-22
[VP] Fix loss of Y210 and Y410 format in vaQuerySurfaceAttributes.

zhangyuankun-star:
Fix loss of Y210 and Y410 format in vaQuerySurfaceAttributes.

Signed-off-by: Yuankun Zhang <yuankunx.zhang@intel.com>

PR #1260, opened 2021-09-08
[VP] Fix regression which caused by #1229

zhangyuankun-star:
Fix regression:libva conformance failures since #1229

Signed-off-by: Zhang Yuankun <yuankunx.zhang@intel.com>


uartie:
Fixes #1259

PR #1263, opened 2021-09-15
Add mem_types when querying surface attributes

zhangyuankun-star:
Add mem_types VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME for decode
and encode when querying surface attributes on Gen12.

Signed-off-by: Zhang Yuankun <yuankunx.zhang@intel.com>


FurongZhang:
Look good to me.

PR #1267, opened 2021-09-23
Handle odd dimensions for external surfaces

ykku16githubinner:
This is based on commit 55268ab458e20365f61d3c636e5eb2597e2d3a30 [1].

It extends the logic to external surfaces.

[1] https://github.com/intel/media-driver/commit/55268ab458e20365f61d3c636e5eb2597e2d3a30

### DIFF
--- a/media_driver/linux/common/ddi/media_libva_util.cpp
+++ b/media_driver/linux/common/ddi/media_libva_util.cpp
@@ -544,7 +544,17 @@ VAStatus DdiMediaUtil_AllocateSurface(
             MOS_ZeroMemory(&gmmCustomParams, sizeof(gmmCustomParams));
             gmmCustomParams.Type          = RESOURCE_2D;
             gmmCustomParams.Format        = mediaDrvCtx->m_caps->ConvertMediaFmtToGmmFmt(format);
-            gmmCustomParams.BaseWidth64   = width;
+            if ((format == Media_Format_YV12) || \
+                (format == Media_Format_I420) || \
+                (format == Media_Format_IYUV) || \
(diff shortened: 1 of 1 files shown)


ykku16githubinner:
@Jister , @XinfengZhang , please take a look.  Thanks in advance.

Chrome OS has use cases that use external surfaces with odd dimensions.  These use cases are still failing with https://github.com/intel/media-driver/commit/55268ab458e20365f61d3c636e5eb2597e2d3a30.

Testing shows that this proposed PR helps.

I have 2 follow-up questions:
1.  I currently don't see alignedWidth or alignedHeight used for external surfaces.  My proposed PR will now make use of alignedWidth for external surfaces.  How do we make sure we don't exceed the limits passed in by the external buffer descriptor?
2. I've made changes only for non-compressible surfaces.  I didn't make changes for compressible surfaces because I have no way to test them.  Will you be able to help with compressible surfaces?

ykku16githubinner:
OK, thanks!  Let me take a look and revise this PR.

ykku16githubinner:
@XinfengZhang , I just pushed version 2 of my changes.  Please take another look.

Note that instead of checking individual formats in an if-statement, I found a gmmlib API function that can tell us if a format is planar.  This function handles NV12 as well, which is what Chrome OS uses.

XinfengZhang:
how about RGBP or other planar format? seems they do not need align with 2

ykku16githubinner:
OK, thanks!

ykku16githubinner:
@XinfengZhang , @dvrogozh , just a follow up.  Did this PR pass internal testing?  Thanks in advance.

PR #1293, opened 2021-11-22
Remove the IPC in CreateAssociationNode

XinfengZhang:
Use random value to pick up a node.
Initialize a random seed in utilities initialization. Generate a random integer in query node.

Signed-off-by: Wang, Pingli <pingli.wang@intel.com>


dvrogozh:
This PR has conflicts. Can you, please, rebase?

XinfengZhang:
I rebased it, but after this patch , the behavior of multiple process will be changed. from "ping pong" to "random", it definitely will impact some test cases. 
from my understanding, these impact is unavoidable. but if there are argument of this behavior, maybe we could use a build option "ENABLE_IPC_FOR_LOAD_BALANCE" which is default false for some special customer.  
@pingliwa

jld:
I've modified Firefox's sandbox to allow this access (mozilla/gecko-dev@23ba2c71df1051d5cb338f340251e6acae94f307), so this shouldn't be necessary.  (The sandbox policy currently allows more of SysV IPC than is strictly necessary, but I think that can be fixed by having Firefox create the resources ahead of time so the IDs can be included in the policy.)  If this patch is going to cause a significant performance hit, I'd suggest not landing it.

XinfengZhang:
thanks @jld , we found that i915 already has the ping pong context switch basing on the device fd. so , for multiple process cases, it will not impact perf. but the code will be totally different, need to rebase and remove "random" selection. I will close it now. thanks

PR #819, opened 2020-01-11
FreeBSD support

jbeich:
Depends on https://github.com/intel/libva/pull/363 and https://github.com/intel/gmmlib/pull/68

<details><summary>Build logs</summary>

http://beefy9.nyi.freebsd.org/data/latest-per-pkg/libva-intel-media-driver/21.1.1/114amd64-default.log
http://beefy10.nyi.freebsd.org/data/latest-per-pkg/libva-intel-media-driver/21.1.1/114i386-default.log
http://beefy6.nyi.freebsd.org/data/latest-per-pkg/libva-intel-media-driver/21.1.1/122amd64-default.log
http://beefy5.nyi.freebsd.org/data/latest-per-pkg/libva-intel-media-driver/21.1.1/122i386-default.log
http://beefy18.nyi.freebsd.org/data/latest-per-pkg/libva-intel-media-driver/21.1.1/main-amd64-default.log
http://beefy17.nyi.freebsd.org/data/latest-per-pkg/libva-intel-media-driver/21.1.1/main-i386-default.log
https://sting.dragonflybsd.org/dports/logs/multimedia___libva-intel-media-driver.log
</details>

<details><summary>Tests output</summary>

```
[==========] Running 43 tests from 13 test cases.
[----------] Global test environment set-up.
[----------] 1 test from MediaCapsDdiTest
[ RUN      ] MediaCapsDdiTest.DecodeEncodeProfile
[       OK ] MediaCapsDdiTest.DecodeEncodeProfile (11 ms)
[----------] 1 test from MediaCapsDdiTest (11 ms total)

[----------] 2 tests from MediaDecodeDdiTest
[ RUN      ] MediaDecodeDdiTest.DecodeHEVCLong
[       OK ] MediaDecodeDdiTest.DecodeHEVCLong (6 ms)
[ RUN      ] MediaDecodeDdiTest.DecodeAVCLong
[       OK ] MediaDecodeDdiTest.DecodeAVCLong (10 ms)
[----------] 2 tests from MediaDecodeDdiTest (16 ms total)

[----------] 2 tests from MediaEncodeDdiTest
[ RUN      ] MediaEncodeDdiTest.EncodeHEVC_DualPipe
[       OK ] MediaEncodeDdiTest.EncodeHEVC_DualPipe (10 ms)
[ RUN      ] MediaEncodeDdiTest.EncodeAVC_DualPipe
[       OK ] MediaEncodeDdiTest.EncodeAVC_DualPipe (12 ms)
[...]

PR #1296, opened 2021-11-26
Replace -DLINUX/-DLINUX_ with __linux__

mattst88:


dvrogozh:
@mattst88 : while that's a correct thing to do per-se, can you, please, provide your side motivation why you are doing this change? is there any functional or build time issue standing behind?

mattst88:
> @mattst88 : while that's a correct thing to do per-se, can you, please, provide your side motivation why you are doing this change? is there any functional or build time issue standing behind?

I don't really have a side motivation; I just started helping to package this software in Gentoo Linux, glanced through the GitHub issues, and thought I'd see how easy or difficult it was to get contributions accepted from outside the company.

I guess if there's a side motivation, it's to work towards standardizing the build system to make it easier for packagers.

Tools/MediaDriverTools/GenDmyHex/CMakeLists.txt
```diff
@@ -22,6 +22,4 @@ cmake_minimum_required (VERSION 2.8)
 project(IntelGenDmyHexTool) 
 add_compile_options(-std=c++11)
 
-add_definitions(-DLINUX_)
```
dvrogozh, line 25: I think that all cmake level changes on Tools/MediaDriverTools/ level are fine. Don't remove them.

media_driver/cmake/linux/media_compile_flags_linux.cmake
```diff
@@ -56,8 +56,6 @@ set(MEDIA_COMPILER_FLAGS_COMMON
     -m${ARCH}
 
     # Global defines
-    -DLINUX=1
```
dvrogozh, line 59: Please, avoid dropping these defines.

media_softlet/cmake/linux/media_compile_flags_linux.cmake
```diff
@@ -56,8 +56,6 @@ set(MEDIA_COMPILER_FLAGS_COMMON
     -m${ARCH}
 
     # Global defines
-    -DLINUX=1
```
dvrogozh, line 59: And here. Please, avoid dropping these defines.

dvrogozh (changes requested):
@mattst88 : thank you, this aligns to my current understanding.

s/LINUX_/__linux__/ are all good. But changes where you remove self-defined macros are a little tricky because there is internal code which is not yet available in this public repo and it also uses these macros. And we have a sync in process to avoid code divergence. This sync eventually will be impacted. Thus, this change needs to be done in few stages:
1. Make all changes possible in this PR which won't impact internal flow
2. Someone internally need fix the internal code
3. After that cmake level change can be done 

Thus, may I ask you to remove some cmake level changes to make this PR easily mergeable? I'll mark them in inline comments.

dvrogozh:
@jbeich : can you, please, evaluate how FreeBSD will be impacted by this change?

media_softlet/cmake/linux/media_compile_flags_linux.cmake
mattst88: Thanks, done.

media_driver/cmake/linux/media_compile_flags_linux.cmake
mattst88: Thanks, done.

Tools/MediaDriverTools/GenDmyHex/CMakeLists.txt
mattst88: Thanks, I've kept all the CMake changes to Tools/MediaDriverTools.

jbeich:
At #819 DragonFly, FreeBSD, NetBSD, OpenBSD and, tentively, Solarish (OpenIndiana) all currently use `-DLINUX` (explicitly defined) in order to avoid patch churn. With `__linux__` (peredefined macro) to support all those Unix-like platform one would need to replace `__linux__` with `__unix__` (predefined on Linux and BSDs but **not** macOS) or explicitly check for `__DragonFly__`, `__FreeBSD__`, `__NetBSD__`, `__OpenBSD__` and `__sun`.

jbeich:
One negative side-effect of merging this PR would be rebase churn in #819. I usually don't rebase unless there's a patch conflict. The patches are used downstream on DragonFly, FreeBSD as cherry-picks and on OpenBSD as local copies... or were supposed to, I'm not sure if OpenBSD Ports has merged the media-driver package.
 https://github.com/freebsd/freebsd-ports/blob/main/multimedia/libva-intel-media-driver/Makefile#L7-L16

mattst88:
I would be happy to rebase/redo this on top of your patch series, if Intel would accept yours. It's approaching its second birthday however...

mattst88:
Another two weeks have passed.

Could someone explain what the actual process is for accepting PRs from Github? I see tags like 'verifying' which indicate to me that something is happening behind the scenes.

Is there someone upstream who is interested and able to merge external contributions?

dvrogozh:
PR should pass review and get approval. Then someone on Intel side should trigger internal ci for the PR, if it will pass it will be merged. "verifying" label is to mark PRs which got picked for this internal ci.

In case of this PR we got https://github.com/intel/media-driver/pull/1296#issuecomment-982084427 which highlights that we can't take this patch as is without taking care of freebsd. I understand that we don't support freebsd right now w/o additional patches, but I don't want to make things worse than they are.

mattst88:
Obsoleted by the first patch in https://github.com/intel/media-driver/pull/1785

Issue #1065, opened 2020-10-14
Build fail with -DGEN8=OFF or -DGEN10=OFF


eero-t:
When building media driver sources from latest release tag "intel-media-20.3.0":
`cmake -G Ninja  -DGEN8=OFF -DGEN9=ON -DGEN10=OFF -DGEN11=ON -DGEN12=ON .  &&  ninja`

The build fails with:
```
In file included from ../media_driver/agnostic/common/cm/cm_command_buffer.cpp:44:
../media_driver/agnostic/gen12/cm/cm_hal_g12.h:31:10: fatal error: mhw_render_hwcmd_g8_X.h: No such file or directory
   31 | #include "mhw_render_hwcmd_g8_X.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

eero-t:
If I change cmake to use "-DGEN8=ON", then the build fails to:
```
../media_driver/agnostic/gen11/codec/hal/codechal_vdenc_vp9_g11.cpp:41:10: fatal error: igcodeckrn_g10.h: No such file or directory
   41 | #include "igcodeckrn_g10.h"
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```

Aren't you doing automated build-testing before releases to check that the GEN options in CMake still work?

dvrogozh:
> Aren't you doing automated build-testing before releases to check that the GEN options in CMake still work?

No, we don't. These are advanced options used in rare cases. So far there were no requests asking for such build configurations to be permanently supported. We focus on the driver build which targets all platforms supported by the driver. Supported build options are described in the top level readme file: https://github.com/intel/media-driver#build-options.

eero-t:
Thanks!

If those options are deprecated legacy ones that're going away, that's fine.  Some note in CMake file about their status would be nice though.

However, if media driver wants to avoid adding accidental dependencies between GENs, I would suggest adding tests for building with GENs disabled, as that kind of dependency mess can be very hard to clean up afterwards.

It needs only build testing, and these two alternate configs would cover it pretty well, so it wouldn't take too much testing time either:
* `-DGEN8=OFF -DGEN9=ON  -DGEN10=OFF -DGEN11=ON  -DGEN12=OFF`
* `-DGEN8=ON  -DGEN9=OFF -DGEN10=ON  -DGEN11=OFF -DGEN12=ON`

mattst88:
[...]

Issue #813, opened 2020-01-01
Multi-GEN disable build failure


th0ma7:
While trying to compile a version specific to Gen9 I deactivated all other uneeded GENx such as
```
ifeq ($(findstring $(ARCH), 'apollolake'),$(ARCH))
CMAKE_ARGS += -DGEN8=OFF
CMAKE_ARGS += -DGEN9=ON
CMAKE_ARGS += -DGEN10=OFF
CMAKE_ARGS += -DGEN11=OFF
CMAKE_ARGS += -DGEN12=OFF
endif
```

Ended-up getting the following build error:
```
/home/th0ma7/git-ffmpeg-QS/spksrc/cross/intel-media-driver/work-apollolake-6.1/media-driver-intel-media-19.4.0r/media_driver/agnostic/common/cm/cm_execution_adv.cpp:46:24: fatal error: cm_hal_g11.h: No such file or directory
 #include "cm_hal_g11.h"
                        ^
compilation terminated.
make[4]: *** [media_driver/CMakeFiles/iHD_drv_video_OBJ.dir/build.make:531: media_driver/CMakeFiles/iHD_drv_video_OBJ.dir/agnostic/common/cm/cm_execution_adv.cpp.o] Error 1
make[4]: Leaving directory '/home/th0ma7/git-ffmpeg-QS/spksrc/cross/intel-media-driver/work-apollolake-6.1/media-driver-intel-media-19.4.0r/build'
make[3]: *** [CMakeFiles/Makefile2:168: media_driver/CMakeFiles/iHD_drv_video_OBJ.dir/all] Error 2
make[3]: Leaving directory '/home/th0ma7/git-ffmpeg-QS/spksrc/cross/intel-media-driver/work-apollolake-6.1/media-driver-intel-media-19.4.0r/build'
make[2]: *** [Makefile:152: all] Error 2
make[2]: Leaving directory '/home/th0ma7/git-ffmpeg-QS/spksrc/cross/intel-media-driver/work-apollolake-6.1/media-driver-intel-media-19.4.0r/build'
make[1]: *** [Makefile:74: intel-media_custom_compile] Error 2
make[1]: Leaving directory '/home/th0ma7/git-ffmpeg-QS/spksrc/cross/intel-media-driver'
make: [../../mk/spksrc.cross-cc.mk:138: arch-apollolake] Error 2 (ignored)
```

th0ma7:
[...]

Issue #451, opened 2018-12-08
Can't build Gen11 only driver w/ disabled VDENC


dvrogozh:
Note: HEVC vdenc is disabled comparing to #999.

```
cmake \
  -DCMAKE_INSTALL_PREFIX=/usr \
  -DCMAKE_INSTALL_LIBDIR=/usr/lib64 \
  -DGEN8_Supported="no" \
  -DGEN8_BDW_Supported="no" \
  -DGEN9_Supported="no" \
  -DGEN9_BXT_Supported="no" \
  -DGEN9_SKL_Supported="no" \
  -DGEN9_CFL_Supported="no" \
  -DGEN9_GLK_Supported="no" \
  -DGEN9_KBL_Supported="no" \
  -DGEN10_Supported="no" \
  -DGEN10_CNL_Supported="no" \
  -DHEVC_Encode_VDEnc_Supported="no" \
  -DFull_Open_Source_Support="yes" ..

CMakeFiles/iHD_drv_video_OBJ.dir/agnostic/common/codec/hal/codechal_encoder_base.cpp.o: In function `CodechalEncodeTrackedBufferHevc* MOS_NewUtil<CodechalEncodeTrackedBufferHevc, CodechalEncoderState*>(CodechalEncoderState*&&)':
/home/dvrogozh/git/github/media-driver/media_driver/agnostic/common/os/mos_utilities.h:799: undefined reference to `CodechalEncodeTrackedBufferHevc::CodechalEncodeTrackedBufferHevc(CodechalEncoderState*)'
CMakeFiles/iHD_drv_video_OBJ.dir/agnostic/common/codec/hal/codechal_huc_cmd_initializer.cpp.o:(.data.rel.ro._ZTV22CodechalCmdInitializer[_ZTV22CodechalCmdInitializer]+0x30): undefined reference to `CodechalCmdInitializer::CmdInitializerSetDmem(bool)'
CMakeFiles/iHD_drv_video_OBJ.dir/agnostic/gen11/codec/hal/codechal_huc_cmd_initializer_g11.cpp.o:(.data.rel.ro._ZTV25CodechalCmdInitializerG11[_ZTV25CodechalCmdInitializerG11]+0x30): undefined reference to `CodechalCmdInitializerG11::CmdInitializerSetDmem(bool)'
[...]

PR #1304, opened 2021-12-02
Allow disabling per-Gen hardware support; add CI testing 

mattst88:
Commit messages contain explanations and data.

tl;dr: This allows the `-DGEN*=OFF` options to work by ignoring them in cases where we know the build is broken. That allows us to use them to disable the large per-Gen media kernels, which are 4.36 MiB for Gen12, 3.77 MiB for Gen11, 3.86 MiB for Gen10, 4.85 MiB for Gen9, 2.47 MiB for Gen8.

CI builds are added so we don't regress. Two follow-on patches disable Gen10 by default (since it never shipped), and (re)disable the Gen8 driver code now that PR 1302 has landed.

cc: @edwarddavidbaker in case you're interested.


mattst88:
*tap* *tap*

Is this thing on?

.github/workflows/ubuntu.yml
```diff
@@ -336,3 +336,328 @@ jobs:
...
+        cmake -DGEN9=OFF \
+          -DCMAKE_INSTALL_PREFIX=/usr \
+          -DCMAKE_INSTALL_LIBDIR=/usr/lib/x86_64-linux-gnu \
+          -DCMAKE_C_FLAGS_RELEASE="$_CFLAGS" \
+          -DCMAKE_CXX_FLAGS_RELEASE="$_CFLAGS" \
+          ..
+        make VERBOSE=1 -j$(nproc)
+        sudo make install
+
+  gcc-10-gen10-on:
```
XinfengZhang, line 470: suggest to remove this one, we will disable gen10 by default.

.github/workflows/ubuntu.yml
```diff
@@ -336,3 +336,328 @@ jobs:
...
+        cmake -DGEN10=ON \
+          -DCMAKE_INSTALL_PREFIX=/usr \
+          -DCMAKE_INSTALL_LIBDIR=/usr/lib/x86_64-linux-gnu \
+          -DCMAKE_C_FLAGS_RELEASE="$_CFLAGS" \
+          -DCMAKE_CXX_FLAGS_RELEASE="$_CFLAGS" \
+          ..
+        make VERBOSE=1 -j$(nproc)
+        sudo make install
+
+  gcc-10-gen11-off:
```
XinfengZhang, line 535: also, suppose need correlated changes if Gen10 is disabled.

media_driver/cmake/linux/media_gen_flags_linux.cmake
```diff
@@ -18,6 +18,8 @@
 # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
 
+option(BROKEN "Accept potentially broken builds when disabling generation-specific options" OFF)
```
XinfengZhang, line 21: BROKEN=ON, then could enable per-gen build. what's the motivation about this build option, why GENXX = OFF is not enough?

.github/workflows/ubuntu.yml
mattst88: I reversed the condition from -DGEN10=OFF to ON in this CI config in the patch that disables Gen10 by default.

What is it that you would rather I do?

.github/workflows/ubuntu.yml
mattst88: I don't understand what you mean. What change to the Gen11 CI config should I make when disabling Gen10 by default?

Enable Gen10 in the Gen11 CI config? Why?

media_driver/cmake/linux/media_gen_flags_linux.cmake
mattst88: Currently the per-Gen build options are broken, except for Gen8.

But the per-Gen build options control two separate things: 1. The per-Gen media kernel binaries, and 2. The per-Gen driver code.

The per-Gen driver code cannot be disabled arbitrarily now.

The per-Gen media kernel binaries can be disabled arbitrarily, and they are large and disabling them provides large file size reductions.

The BROKEN option controls whether the per-Gen driver code is affected by the per-Gen config options, allowing us to disable the per-Gen media kernel binaries.

dvrogozh: I understand the desire to have options to reduce driver footprint, but it's not good idea to workaround something which should actually be fixed. Issue with gen10 switch off is a clear issue and needs proper resolution.

I think we should not introduce BROKEN option. Instead, I suggest to start introducing new build configurations (w/ disabled gens) which currently works to make sure we don't regress. For example, -DGEN8=OFF got fixed, let's have this path covered in github actions ci. And add other configs which currently build ok.

mattst88: I have some work in progress patches to fix Gen9, but I don't expect to be able to fix disabling the driver code for Gen10, Gen11, or Gen12.

Is someone from Intel going to sign up for that?

We can of course ship my patch as is in ChromeOS to be able to reduce the binary size, but I feel like I've provided a solid and incremental path forward that gets 90% of the size reduction, strictly improves the situation, and doesn't force anyone to drop what they're doing to untangle the currently broken per-Gen driver code disablement.

Alternatively, I'm happy to either
 -  rename the `BROKEN` option if the name is the problem
 - or I can remove `BROKEN` entirely and make the per-Gen driver code that is currently required non-optional, leaving the `GEN*` options to only disable the media kernel binaries. I create the `BROKEN` option to make the incremental work of fixing the remaining Gen options simpler.

mattst88:
Anyone want to continue the discussions?

.github/workflows/ubuntu.yml
XinfengZhang: I mean we may do not need this catalog "gcc-10-gen10-on" or "gcc-10-gen10-off" in near future. of course ,we may keep it during transition period

.github/workflows/ubuntu.yml
mattst88: Okay. I'm resolving this conversation -- unless I've misunderstood.

.github/workflows/ubuntu.yml
mattst88: Right, I totally agree.

media_driver/cmake/linux/media_gen_flags_linux.cmake
mattst88: > * rename the BROKEN option if the name is the problem

Having received no feedback for a month, I've speculatively made this change, changing the option name to `-DONLY_MEDIA_KERNELS` and inverting the value (so that you have to set it to `OFF` to produce a broken build with `-DGEN*=OFF`).

media_driver/cmake/linux/media_gen_flags_linux.cmake
```diff
@@ -18,6 +18,9 @@
 # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
 
+option(ONLY_MEDIA_KERNELS "Make per-Gen options disable only media-kernels (not driver code) if driver code is known to be required for a successful build"
```
XinfengZhang, line 21: want to clarify these:
1. it is a build option during per-gen options is not clean. related driver code still could not excluded when the per-gen build option is "OFF"
2. from previous conditions "if(ONLY_MEDIA_KERNELS OR GEN12_TGLLP);xxx"; if the option is "ON", it will always build related media driver.  your expectation should be , if ONLY_MEDIA_KERNELS is ON, but GEN12_TGGLP is OFF, then the media kernel files of TGLLP should not be compiled , but the code of TGLLP will still be compiled, right?
from this perspective, the ONLY_MEDIA_KERNELS is looks confusing, it looks not a good name to be a switch.  it is something like "allow source code compiled even disable the platform",  "ALLOW_REDUNDANT_CODE" is also not good enough, I still have no idea with a better name.  or remove the new definitions, but add comments to describe that current status, GENxx=OFF maybe just disable the kernel files.  both methods looks not so graceful. 
which configuration you cared mostly?  if you just care 1 or 2 platform disabling, maybe we could focus on that , resolve it one by one?
3. your changes with github actions does not include any reference of MEDIA_KERNELS, so, why it will benefit your project?

mattst88: >     1. it is a build option during per-gen options is not clean. related driver code still could not excluded when the per-gen build option is "OFF"

It's true that it's not a clean option. My plan/hope is for the `ONLY_MEDIA_KERNELS` option to be removed once each Gen's driver code can be safely disabled.

>     2. from previous conditions "if(ONLY_MEDIA_KERNELS OR GEN12_TGLLP);xxx"; if the option is "ON", it will always build related media driver.  your expectation should be , if ONLY_MEDIA_KERNELS is ON, but GEN12_TGGLP is OFF, then the media kernel files of TGLLP should not be compiled , but the code of TGLLP will still be compiled, right?

That's exactly right.

>        from this perspective, the ONLY_MEDIA_KERNELS is looks confusing, it looks not a good name to be a switch.  it is something like "allow source code compiled even disable the platform",  "ALLOW_REDUNDANT_CODE" is also not good enough, I still have no idea with a better name.  or remove the new definitions, but add comments to describe that current status, GENxx=OFF maybe just disable the kernel files.  both methods looks not so graceful.
>        which configuration you cared mostly?  if you just care 1 or 2 platform disabling, maybe we could focus on that , resolve it one by one?

I agree, it's not graceful and it's somewhat confusing. I'd kind of suggest that this isn't unexpected given that almost none of the `-DGEN` options currently work if set to `OFF` :)

>     3. your changes with github actions does not include any reference of MEDIA_KERNELS, so, why it will benefit your project?

It would benefit anyone that wishes to reduce the disk usage of the project by disabling media kernels (and code, in the case of Gen8).

The intention behind the `ONLY_MEDIA_KERNELS` option (which is `ON` by default) is to provide a baseline of support for disabling the `GEN` options. Currently most of them, when disabled, cause the build to fail. My strategy is to

1. avoid those build failures with the `ONLY_MEDIA_KERNELS` mechanism (while still providing a simple way to reproduce the build failures for purposes of fixing them)
2. allow significant disk usage savings by disabling the media kernels even in cases where the rest of that Gen's driver support code cannot (yet) be disabled

To that end, the GitHub Actions are there to ensure that each of the `-DGEN=` options does not produce a broken build. In the case of Gen 8, this means disabling the media kernels *and* driver code. For other Gens it means only disabling the media kernels -- until those issues are resolved which will enable us to disable the Gen-specific driver code as well.

I could invert the default of `ONLY_MEDIA_KERNELS` from `ON` to `OFF` so that a build configuration that only specified `-DGEN9=OFF` produces a build failure, and I could add `-DONLY_MEDIA_KERNELS=ON` to the GitHub Action for Gen9, but I don't know why that would be better in any way.


Could I get some clarity on what the concerns are? 

* Is the goal of building the project for a single generation in order to reduce the disk usage from 36 MiB to 18 MiB not considered sufficiently valuable?

* Is the strategy of achieving 90% of the goal by allowing disabling media kernels independently of resolving long standing issues not a good one? (Issue #451 was filed in 2018; similar issues #813 and #1065 were filed later).

I'm unsure whether we're in agreement on the goals, the strategy, both, or neither.

Like if it's just confusion about how the option should be named or how it should behave then, great, we can surely find something better. If there's disagreement about the value in reducing binary size, then we probably need to take a step back and have a different discussion. In between those, if there's disagreement on the strategy of allowing disabling media kernels independently of driver code, then we probably need to figure out a new path.

XinfengZhang: thanks @mattst88,  I just noticed you did not add "ONLY_MEDIA_KERNEL" for gen8, because gen8 is clean. yes, we should not change the value of "ONLY_MEDIA_KERNEL" in  configuration Github Actions.
I have no concern with the commit itself , except the naming. better to "ALLOW_CODE_BESIDES_KERNEL" ?  , "ALLOW_REDUNDANT_CODE" or .... , just like what I said, I have no good idea about it. anyway, I agree the value and the purpose. thanks for your contribution

XinfengZhang (approved):

media_driver/cmake/linux/media_gen_flags_linux.cmake
mattst88: Okay, thank you very much!

I've changed the name to `ENABLE_REQUIRED_GEN_CODE` which I think is better than `ONLY_MEDIA_KERNEL`. I've also added a comment above the option documenting that we plan to remove it when all Gen options can disable the driver code.

XinfengZhang:
2 build failure.

mattst88:
Okay, rebased onto master, updated for some platform additions, and fixed the two build failures.

Looks like I need someone to approve the workflow to run the CI again.

mattst88:
> Okay, rebased onto master, updated for some platform additions, and fixed the two build failures.

More fixes. Should be good to go now.

mattst88:
Thank you. I'll open further PRs that allow disabling per-Gen driver code as I have time. Hopefully those will not take nearly as long to merge.

Issue #1300, opened 2021-11-29
media-driver: hitting double free running with gst-vaapi decode pipeline with multiple decode


SiewHoon:
## System information
- CPU information(`cat /proc/cpuinfo | grep "model name" | uniq`):
model name      : Genuine Intel(R) CPU 0000 @ 1.70GHz
- GPU information(`lspci -nn | grep -E 'VGA|isplay`):
 00:02.0 VGA compatible controller [0300]: Intel Corporation Device [8086:9a78] (rev 01)
- Display server if rendering to display(X or wayland): X
- OS Ubuntu: 20.04

## Issue behavior
Every time running same command. Hitting double free.

Errror:
Pipeline is PREROLLED ...
Setting pipeline to PLAYING ...
Redistribute latency...
New clock: GstSystemClock
malloc_consolidate(): invalid chunk size
./ifpd1.sh: line 18: 215113 Aborted                 (core dumped) gst-launch-1.0 filesrc location=remote_1080_8bit_3mbps_noB.265 ! ....

Command pipeline to reproduce the issue:

**shell scripting :**
#!/bin/bash
export DISPLAY=:0
WIDTH=960
HEIGHT=720

gst-launch-1.0 \
    filesrc location=remote_1080_8bit_3mbps_noB.265 ! h265parse ! vaapih265dec ! vaapipostproc width=960 height=720 ! vaapisink \
    filesrc location=remote_1080_8bit_3mbps_noB.265 ! h265parse ! vaapih265dec ! vaapipostproc width=960 height=720 ! vaapisink \
    filesrc location=remote_1080_8bit_3mbps_noB.265 ! h265parse ! vaapih265dec ! vaapipostproc width=960 height=720 ! vaapisink \
    filesrc location=remote_1080_8bit_3mbps_noB.265 ! h265parse ! vaapih265dec ! vaapipostproc width=960 height=720 ! vaapisink \
    filesrc location=remote_1080_8bit_3mbps_noB.265 ! h265parse ! vaapih265dec ! vaapipostproc width=960 height=720 ! vaapisink \
    filesrc location=remote_1080_8bit_3mbps_noB.265 ! h265parse ! vaapih265dec ! vaapipostproc width=960 height=720 ! vaapisink \
    filesrc location=remote_1080_8bit_3mbps_noB.265 ! h265parse ! vaapih265dec ! vaapipostproc width=960 height=720 ! vaapisink \
[...]

PR #1305, opened 2021-12-02
[Media Common]Fix the user setting memory free

pingliwa:
This commit fix the #1300

Signed-off-by: Wang, Pingli <pingli.wang@intel.com>

### DIFF
--- a/media_softlet/linux/common/os/mos_utilities_specific_next.cpp
+++ b/media_softlet/linux/common/os/mos_utilities_specific_next.cpp
@@ -70,6 +70,7 @@ double MosUtilities::MosGetTime()
 //!
 const char *const MosUtilitiesSpecificNext::m_mosTracePath  = "/sys/kernel/debug/tracing/trace_marker_raw";
 int32_t           MosUtilitiesSpecificNext::m_mosTraceFd    = -1;
+MosMutex          MosUtilitiesSpecificNext::m_userSettingMutex;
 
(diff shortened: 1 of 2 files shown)


XinfengZhang (approved):
this mutex maybe removed in future after registry refined.

media_softlet/linux/common/os/mos_utilities_specific_next.cpp
```diff
@@ -612,6 +613,7 @@ MOS_STATUS MosUtilitiesSpecificNext::UserFeatureSet(MOS_PUF_KEYLIST *pKeyList, M
     MOS_AtomicIncrement(&MosUtilities::m_mosMemAllocFakeCounter);  //ulValueBuf does not count it, because it is freed after the MEMNJA final report.
     MOS_OS_NORMALMESSAGE("ulValueBuf %p for key %s", ulValueBuf, NewKey.pValueArray[0].pcValueName);
 
+    m_userSettingMutex.Lock();
```
XinfengZhang, line 616: how about multiple-process case?

pingliwa:
Hi,
Different process loads different .so. The mutex is different.  This function only writes value to cpu memory not write to files.

Best Wishes,
Pingli

hanlong1:
@pingliwa could you please rebase this PR?

pingliwa:
> @pingliwa could you please rebase this PR?

Done

PR #1321, opened 2022-01-01
Report HEVC features and block sizes for gen11

fhvwy:
Like gen12, gen11 also has feature support which doesn't match the
original gen9 implementation, so we need to expose these attributes
there as well.

I do think all encoders should probably expose these attributes so that the user doesn't need to carry default settings based on the gen9 i965 implementation.  I'm not sure exactly what the capabilities of gen9 are, though, so I've only done gen11 here.

### DIFF
--- a/media_driver/linux/gen11/ddi/media_libva_caps_g11.cpp
+++ b/media_driver/linux/gen11/ddi/media_libva_caps_g11.cpp
@@ -410,6 +410,68 @@ VAStatus MediaLibvaCapsG11::GetPlatformSpecificAttrib(VAProfile profile,
             }
             break;
         }
+#if VA_CHECK_VERSION(1, 12, 0)
+        case VAConfigAttribEncHEVCFeatures:
+        {
+            if ((entrypoint == VAEntrypointEncSlice || entrypoint == VAEntrypointEncSliceLP) && IsHevcProfile(profile))
+            {
+                VAConfigAttribValEncHEVCFeatures hevcFeatures = {0};
+                hevcFeatures.bits.separate_colour_planes = VA_FEATURE_NOT_SUPPORTED;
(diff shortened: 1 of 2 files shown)


dvrogozh:
Sounds reasonable, but someone else supporting hevc on our side needs to look. @XinfengZhang : can you, please, review and help add other people?

media_driver/linux/gen11/ddi/media_libva_caps_g11.cpp
```diff
@@ -410,6 +410,68 @@ VAStatus MediaLibvaCapsG11::GetPlatformSpecificAttrib(VAProfile profile,
             }
             break;
         }
+#if VA_CHECK_VERSION(1, 12, 0)
+        case VAConfigAttribEncHEVCFeatures:
+        {
```
XinfengZhang, line 415: VAConfigAttribEncHEVCBlockSizes was separated for VAEntrypointEncSliceLP & VAEntrypointEncSlice. 
but  VAConfigAttribEncHEVCFeatures not, in my impression, they also have different features. @Tianhaol  could you help to check it?

XinfengZhang: @thomasli21801

Tianhaol: Yes, for VAConfigAttribEncHEVCFeatures, implement it as one if VAEntrypointEncSliceLP and  VAEntrypointEncSlice have the same values for each field.

thomasli21801: Block size is related w/ log2_max_coding_tree_block_size_minus3 below for different implementations. Others LGTM.

XinfengZhang (approved):

PR #1327, opened 2022-01-19
ADL-N Support : Added the cmake options GEN12_ADLN 

ss22ongithub:
Added missing cmake option GEN12_ADLN for ADL-N SoCs

### DIFF
--- a/media_driver/cmake/linux/media_gen_flags_linux.cmake
+++ b/media_driver/cmake/linux/media_gen_flags_linux.cmake
@@ -81,6 +81,10 @@ cmake_dependent_option(GEN12_ADLP
     "Enabled ADLP support (Gen12)" ON
     "GEN12_TGLLP" OFF)
 
+cmake_dependent_option(GEN12_ADLN
+    "Enabled ADLN support (Gen12)" ON
+    "GEN12_TGLLP" OFF)
+
 cmake_dependent_option(Xe_M
     "Enabled support for Xehp_sdv+ platforms" ON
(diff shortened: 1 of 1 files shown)


media_driver/cmake/linux/media_gen_flags_linux.cmake
```diff
@@ -81,6 +81,10 @@ cmake_dependent_option(GEN12_ADLP
     "Enabled ADLP support (Gen12)" ON
     "GEN12_TGLLP" OFF)
 
+cmake_dependent_option(GEN12_ADLN
```
XinfengZhang, line 84: I think , maybe the pre-condition should not be GEN12, it should be GEN12_TGLLP, because ADLN are using lots of TGLLP code

XinfengZhang (changes requested):
I suppose, there should be another patch which will refer IGFX_GEN12_ADLN_SUPPORTED in the code

media_driver/cmake/linux/media_gen_flags_linux.cmake
ss22ongithub: This commit was already referring to  IGFX_GEN12_ADLN_SUPPORTED
 https://github.com/intel/media-driver/commit/348e98e90d240deecd3040f3846a27e9bb6c3ac1
 
The media_driver/media_interface/media_srcs.cmake ,  was using GEN12_ADLN 

let me

XinfengZhang (approved):

Issue #690, opened 2019-07-23
[ffmpeg][vaapi][ARGB/BGRA] mismatch of ARGB/BRGA


fulinjie:
cmdline:
ffmpeg -hwaccel vaapi -vaapi_device /dev/dri/renderD128 -hwaccel_output_format vaapi -i Nature.h264 -vf scale_vaapi=format=rgb32,hwdownload,format=rgb32 -f rawvideo /dev/null

https://github.com/FFmpeg/FFmpeg/blob/3883c9d1479d3e653bb66c9960fdd2ebd6ba9594/libavutil/hwcontext_vaapi.c#L607
if (expected_format->fourcc == test_image.format.fourcc) {

expected_format->fourcc is different from  test_image.format.fourcc.

There is mismatch between ffmpeg and media driver for ARGB/BGRA, thus vaDeriveImage can't be used, and it causes the performance drop.

wangyan-intel:
@fulinjie Could you share your video clip for debugging? Thanks.

fulinjie:
@wangyan-intel Random clips should be ok, since the cmdline simply decoded a clips and do color space convert.

wangyan-intel:
I tried it by one MP4 h264 clip but cannot find this issue. the commit of iHD driver is c2aae3e8. Could you please double check it? Thanks.

fulinjie:
Hi Yan, update to latest version of iHD driver, and this issue still exists.
This happens when vaapi_frames_init is called for VPP (csc, called the second time.)

vaCreateImage is used instead of vaDeriveImage, thus makes the performance drop.

wangyan42164:
@fulinjie Could you please try #692? It works fine on my side. Thanks.

fulinjie:
Hi @wangyan42164, how is this going?

wangyan-intel:
@XinfengZhang As I know, @Jister is working on this?

MicroYY:
We're trying @wangyan42164 's fix.  As it's an old PR and contains confilct, @wangyan42164 could you pls rebase the patch?

MicroYY:
@LhGu What's the status of your PR?

LhGu:
> @LhGu What's the status of your PR?

It is under review.

LhGu:
There is a limitation in the driver stack. It only supports the hex order is B-G-R-A, and our driver treats the hex order as VA_FOURCC_ARGB format.  I already have two PRs（#1334， #1335 ）to work around it.  
[...]

PR #1334, opened 2022-01-24
Split BGRA/ARGB media_format for the surface

LhGu:
Fix: #690
Signed-off-by: lihao.gu@intel.com

### DIFF
--- a/media_driver/linux/common/codec/ddi/media_ddi_encode_avc.h
+++ b/media_driver/linux/common/codec/ddi/media_ddi_encode_avc.h
@@ -527,6 +527,7 @@ private:
     inline bool NeedDisplayFormatSwizzle(DDI_MEDIA_SURFACE *rawSurface)
     {
         if (Media_Format_A8R8G8B8 == rawSurface->format ||
+            Media_Format_B8G8R8A8 == rawSurface->format ||
             Media_Format_B10G10R10A2 == rawSurface->format)
         {
             return true;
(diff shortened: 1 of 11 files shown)


LhGu:
In the past,  both VA_FOURCC_BGRA and VA_FOURCC_ARGB are related to Media_Format_A8R8G8B8,  and further is related to  Format_A8R8G8B8 in MOS. Maybe media-driver only supports one format Format_A8R8G8B8 now.  
In the little endian mode, A8R8G8B8 may mean BGRA in the memory, and the output data is in the B G R A order in the hex file.

LhGu:
When our surface gets  VA_FOURCC format by Media_Format, Media_Format_A8R8G8B8 only is mapping to VA_FOURCC_ARGB. This could cause that ffmpeg checks the VA_FOURCC format failed, because original VA_FOURCC is VA_FOURCC_BGRA. 
Maybe we only need to make Media_Format_A8R8G8B8  to be mapping to VA_FOURCC_BGRA, because our output  data is  BGRA order in the hex file. 
However,  our original mapping relation that  Media_Format_A8R8G8B8 is mapping to VA_FOURCC_ARGB may be correct for some other usages.  I add a Media_Format_B8G8R8A8  format  and keep the MOS_FORMAT unchanged to resolve this VA_FOURCC mismatch issue. This not affect original logic flow .

LhGu:
Close it,  and  I have another simple fix, because only BGRA  can work.

LhGu:
I have verified some cases, and I found that Media_Format_A8R8G8B8 mapping to VA_FOURCC_ARGB has some workaround usages for MediaSDK. If we change this mapping relation, some cases will be failed. So I reopened this PR to fix this issue.

PR #1335, opened 2022-01-24
Change VA_FOURCC format in the mapping relation to support BGRA

LhGu:
Fix: #690
Signed-off-by: Gu, Lihao <lihao.gu@intel.com>

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -404,7 +404,7 @@ int32_t DdiMedia_MediaFormatToOsFormat(DDI_MEDIA_FORMAT format)
         case Media_Format_R8G8B8A8:
             return VA_FOURCC_RGBA;
         case Media_Format_A8R8G8B8:
-            return VA_FOURCC_ARGB;
+            return VA_FOURCC_BGRA;
         case Media_Format_B10G10R10A2:
             return VA_FOURCC_A2R10G10B10;
         case Media_Format_R10G10B10X2:


LhGu:
In the past, both VA_FOURCC_BGRA and VA_FOURCC_ARGB are related to Media_Format_A8R8G8B8, and further is related to Format_A8R8G8B8 in MOS. Maybe media-driver only supports one format Format_A8R8G8B8 now.
In the little endian mode, A8R8G8B8 may mean BGRA in the memory, and I have checked the output data's order is B G R A in the hex file.

LhGu:
When our surface gets VA_FOURCC format by Media_Format, Media_Format_A8R8G8B8 only is mapping to VA_FOURCC_ARGB. This could cause that ffmpeg checks the VA_FOURCC format failed, because original VA_FOURCC is VA_FOURCC_BGRA.
We only need to make Media_Format_A8R8G8B8 to be mapping to VA_FOURCC_BGRA, because our output data is BGRA order in the hex file, and our output data don't support ARBG order in the hex file now.

LhGu:
I have another PR to resolve this issue, because Media_Format_A8R8G8B8 is mapping to VA_FOURCC_ARGB may be correct for some other workaround usages. Another PR link: https://github.com/intel/media-driver/pull/1334

LhGu:
I have verified some cases, and I found that Media_Format_A8R8G8B8 mapping to VA_FOURCC_ARGB has some workaround usages for MediaSDK.  If we change this mapping relations, some cases will be failed. So I reopened another PR to fix this issue.

Issue #1344, opened 2022-02-08
GCC v11 warnings for CM component


eero-t:
## System information
- Ubuntu 21.10 container with GCC v11.2.0

### Describe the current behavior

```
cmake -LH -Wno-dev -G Ninja -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR} .  &&  ninja
...
// Turn compiler warnings into fatal errors
MEDIA_BUILD_FATAL_WARNINGS:BOOL=ON
...
../media_driver/agnostic/common/cm/cm_mem_sse2_impl.cpp: In function 'void FastMemCopy_SSE2_movntdq_movdqa(void*, void*, size_t)':
../media_driver/agnostic/common/cm/cm_mem_sse2_impl.cpp:42:64: warning: expression does not compute the number of elements in this array; element type is 'uint32_t' {aka 'unsigned int'}, not 'DQWORD' {aka '__m128'} [-Wsizeof-array-div]
   42 |     const size_t doubleQuadWordsPerPrefetch = sizeof(PREFETCH) / sizeof(DQWORD);
      |                                               ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
../media_driver/agnostic/common/cm/cm_mem_sse2_impl.cpp:42:66: note: add parentheses around 'sizeof (DQWORD)' to silence this warning
   42 |     const size_t doubleQuadWordsPerPrefetch = sizeof(PREFETCH) / sizeof(DQWORD);
      |                                                                  ^~~~~~~~~~~~~~
      |                                                                  (             )
../media_driver/agnostic/common/cm/cm_mem_sse2_impl.cpp: In function 'void FastMemCopy_SSE2_movdqu_movdqa(void*, void*, size_t)':
../media_driver/agnostic/common/cm/cm_mem_sse2_impl.cpp:84:64: warning: expression does not compute the number of elements in this array; element type is 'uint32_t' {aka 'unsigned int'}, not 'DQWORD' {aka '__m128'} [-Wsizeof-array-div]
   84 |     const size_t doubleQuadWordsPerPrefetch = sizeof(PREFETCH) / sizeof(DQWORD);
      |                                               ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
../media_driver/agnostic/common/cm/cm_mem_sse2_impl.cpp:84:66: note: add parentheses around 'sizeof (DQWORD)' to silence this warning
[...]

PR #1350, opened 2022-02-10
build: fix build with gcc-11

dvrogozh:
Fixes: https://github.com/intel/media-driver/issues/1344

### DIFF
--- a/media_driver/agnostic/common/cm/cm_mem_sse2_impl.cpp
+++ b/media_driver/agnostic/common/cm/cm_mem_sse2_impl.cpp
@@ -31,6 +31,12 @@
 
 #include <mmintrin.h>
 
+// GCC>=11 gives a warning when dividing sizeof(array) with sizeof(another_type)
+// in case division was mistyped and meant to calculate array size.
+// As that's not the case here, use extra divisor parenthesis to suppress
+// the warning.
+#define DQWORD_PER_PREFETCH(P) ( sizeof(P)/(sizeof(DQWORD)) )
+
 void FastMemCopy_SSE2_movntdq_movdqa(
     void* dst,
     void* src,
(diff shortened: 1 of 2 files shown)


media_driver/agnostic/common/cm/cm_mem_sse2_impl.cpp
```diff
@@ -39,7 +39,7 @@ void FastMemCopy_SSE2_movntdq_movdqa(
     CM_ASSERT( IsAligned( dst, sizeof(DQWORD) ) );
     CM_ASSERT( IsAligned( src, sizeof(DQWORD) ) );
 
-    const size_t doubleQuadWordsPerPrefetch = sizeof(PREFETCH) / sizeof(DQWORD);
+    const size_t doubleQuadWordsPerPrefetch = sizeof(PREFETCH) / (sizeof(DQWORD));
```
dvrogozh, line 42: To code owners: is that possible to redefine PREFETCH to DQWORD to avoid the need to additional `()`?

XinfengZhang:
@shawnli2 , could you help to take a look and answer the question

media_driver/agnostic/common/cm/cm_mem_sse2_impl.cpp
HeFan2017: Maybe no. PREFETCH is a 32-byte type, and its size is double of DQWORD.

dvrogozh: Thank you, Fan. So, this needs to remain as is. I updated PR follow suggestions from @eero-t. Introduced a macro and added a comment why we make this change.

media_driver/agnostic/common/cm/cm_mem_sse2_impl.cpp
```diff
@@ -31,6 +31,12 @@
 
 #include <mmintrin.h>
 
+// GCC>=11 gives a warning when dividing sizeof(array) with sizeof(another)type)
+// in case division was mistyped and meant to calculate array size.
+// That's not out case here, so extra parenthesis for the divisor suppresses
+// the warning.
+#define DQWORD_PER_PREFETCH(P) ( sizeof(P)/(sizeof(DQWORD)) )
```
eero-t, line 38: Few typos: 
s/another)type/another type/
s/not out case/not the case/

And the last clause would read better with a verb: "As that's not the case here, use extra (divisor) parenthesis to suppress the warning."

LGTM otherwise.

dmitryermilov (approved):

PR #1361, opened 2022-03-18
add wayland output surport

XinfengZhang:


media_driver/linux/common/ddi/wayland-drm-client-protocol.h
```diff
@@ -0,0 +1,307 @@
+/* Generated by wayland-scanner 1.12.0 */
```
ceyusa, line 1: If this is autogenerated, wouldn't it be left out of git and generate it within the build system?

XinfengZhang: yes, agree. will change it , but now will keep it for test purpose.

sandrobonazzola:
Is this still relevant?

PR #1384, opened 2022-04-13
[CM] Skip parsing info when device is unbound

### LINKED ISSUE #1383: "readlink -errno 2" segfault when device does not have "driver"
## System information
- Impacted GPU platform: All?
- Host machine if it's descreted card: ?
- Good libva/libva-utils/gmmlib/media-driver version? media-driver 22.1.0
- Bad libva/libva-utils/gmmlib/media-driver version? media-driver 22.3.1

### Describe the current behavior
When trying to initialize iHD vaapi (e.g. running `vainfo`), it seems to try opening every card in `/dev/dri` then going to its `/sys/devices/pci...` entry to do a `readlink()` on a `driver` file, which it assumes exists. If it does not, then readlink fails and the program (e.g. `vainfo`) segfaults.

The `driver` file does not exist because it goes into my NVIDIA card, which I have unbound from the nvidia module when I am not using it. This is useful for dual GPU setups on laptops where power management is a consideration.

### Describe the expected behavior
If `readlink()` fails or if the `driver` file/link does not exist, then it should ignore this device/card and try another one.

omcaif:
When a device is not bound to a driver, the `driver` link does not
exist, so we can stop parsing the `sysfs` files, as that device is not
useful, and return.

Fixes #1383,

### DIFF
--- a/cmrtlib/linux/hardware/drm_device.h
+++ b/cmrtlib/linux/hardware/drm_device.h
@@ -585,7 +585,13 @@ static int parse_separate_sysfs_files(int maj, int min,
     snprintf(resourcename, sizeof(resourcename), "%s/resource", pci_path);
 
     if (readlink(driverpath, drivername, PATH_MAX) < 0)
-        printf("   readlink -errno %d\n", errno);
+    {
+        /* Some devices might not be bound to a driver */
+        if (errno == ENOENT)
+            return -errno;
+        else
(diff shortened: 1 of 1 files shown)


XinfengZhang (approved):
LGTM,
@[kankanzh](https://github.com/kankanzh) could you help to take a look

cqian2 (approved):
Looks good to me.

PR #1365, opened 2022-03-24
Fixes #1354

pingliwa:
Fixes #1354. It is to fix the compiling issue for gcc12.0.1.
[-Werror=address] [-Werror=dangling-pointer=][-Werror=alloc-size-larger-than=]

Signed-off-by: Wang, Pingli <pingli.wang@intel.com>

Issue #1354, opened 2022-02-27
Compile error when compiling with -Werror using gcc 12.0.1


heitbaum:
## System information
Cross compiling media-driver 22.2.1 using gcc 12-20220220

### Describe the current behavior
Fails compile with:
- `warning: dangling pointer 'address2Counter' to 'ctr' may be used [-Wdangling-pointer=]`
- `warning: argument 1 range [18446744071562067968, 18446744073709551615] exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]`

### Describe the expected behavior
Compiles correctly with gcc 11.2.0
Using the following compile options allows these errors to pass as warnings:
- `export CXXFLAGS="${CXXFLAGS} -Wno-error=alloc-size-larger-than= -Wno-error=dangling-pointer"`

## build log

```
[43/942] Building CXX object media_driver/CMakeFiles/iHD_drv_video_OBJ.dir/agnostic/common/cm/cm_kernel_ex.cpp.o
In file included from /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-11.0-devel/build/media-driver-22.2.1/media_common/agnostic/common/os/mos_utilities.h:138,
                 from /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-11.0-devel/build/media-driver-22.2.1/media_common/agnostic/common/os/mos_os.h:32,
                 from /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-11.0-devel/build/media-driver-22.2.1/media_driver/agnostic/common/os/mos_context.h:30,
                 from /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-11.0-devel/build/media-driver-22.2.1/media_driver/linux/common/ddi/media_libva_common.h:37,
                 from /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-11.0-devel/build/media-driver-22.2.1/media_driver/linux/common/cm/ddi/media_libva_cm.h:30,
                 from /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-11.0-devel/build/media-driver-22.2.1/media_driver/linux/common/cm/hal/cm_def_os.h:41,
[...]

Issue #1353, opened 2022-02-27
Fatal compile error of media_ddi_decode_vp8.cpp when compiling with gcc 12.0.1


heitbaum:
## System information
Cross compiling media-driver 22.2.1 using gcc 12-20220220

### Describe the expected behavior
Compiles correctly with gcc 11.2.0

## build log

```
[562/942] Building CXX object media_driver/CMakeFiles/iHD_drv_video_OBJ.dir/linux/common/codec/ddi/media_ddi_decode_vp8.cpp.o
FAILED: media_driver/CMakeFiles/iHD_drv_video_OBJ.dir/linux/common/codec/ddi/media_ddi_decode_vp8.cpp.o 
[...]

PR #1385, opened 2022-04-14
Modify cmakelist.txt to use ubuntu default locations,readme.md reflects expected behaviour, post-install guide, extra trouple shooting steps 

FCLC:
This is mostly a product of what happened in #1342 

Due to a problem with the installation mechanisms, when following the documentation word for word, the files would end up in the wrong locations and improperly sourced. 

For example, Ubuntu defaults to using `/usr/lib/x86_64-linux-gnu/dri` for VAAPI, but cmake was using `/usr/lib/dri` instead. 

We now instruct the user to export the wanted location *before* running cmake so that cmake can properly register the desired location if a prefix isnt specified. 

The default has now also been changed to the Ubuntu default of `/usr/lib/x86_64-linux-gnu/` for libraries within cmake. 

As a catch all for previous installations/edge cases, the suggestion is now to export `LIBVA_DRIVERS_PATH=/usr/lib/x86_64-linux-gnu/dri:/usr/lib/dri:<path-contains-iHD_drv_video.so>`

Added guide for verifying if installation worked correctly under new header titled "Post-Installation"

Finally, in reference to #1353 and partially in reference to #1354 and #1365, mark gcc-12.x as pre-release in the issues section while we wait for the official launch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,7 @@ include(os_release_info.cmake)
 get_os_release_info(os_name os_version)
 
 if(NOT DEFINED CMAKE_INSTALL_PREFIX OR CMAKE_INSTALL_PREFIX STREQUAL "")
-    set(CMAKE_INSTALL_PREFIX "/usr/local" CACHE PATH "..." FORCE)
+    set(CMAKE_INSTALL_PREFIX "/usr/lib/x86_64-linux-gnu" CACHE PATH "..." FORCE)
 endif()
 message(STATUS "CMAKE_INSTALL_PREFIX = ${CMAKE_INSTALL_PREFIX}")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,7 @@ include(os_release_info.cmake)
 get_os_release_info(os_name os_version)
 
 if(NOT DEFINED CMAKE_INSTALL_PREFIX OR CMAKE_INSTALL_PREFIX STREQUAL "")
-    set(CMAKE_INSTALL_PREFIX "/usr/lib/x86_64-linux-gnu" CACHE PATH "..." FORCE)
+    set(CMAKE_INSTALL_PREFIX "/usr/lib/x86_64-linux-gnu/" CACHE PATH "..." FORCE)
 endif()
 message(STATUS "CMAKE_INSTALL_PREFIX = ${CMAKE_INSTALL_PREFIX}")
 
@@ -80,8 +80,12 @@ endif()
 
 add_subdirectory(media_driver)
 
+message(STATUS "LIBVA_DRIVERS_PATH = ${LIBVA_DRIVERS_PATH}")
+
+set(LIBVA_DRIVERS_PATH "/usr/lib/x86_64-linux-gnu/dri/")
+
 if("${LIBVA_DRIVERS_PATH}" STREQUAL "")
-    set(LIBVA_DRIVERS_PATH "${CMAKE_INSTALL_FULL_LIBDIR}/dri/")
+    set(LIBVA_DRIVERS_PATH "/usr/lib/x86_64-linux-gnu/dri/")
 endif()
 message(STATUS "LIBVA_DRIVERS_PATH = ${LIBVA_DRIVERS_PATH}")
 
@@ -134,4 +138,3 @@ string(TOLOWER "${BUILD_TYPE}.${CPACK_RPM_PACKAGE_ARCHITECTURE}" PACKAGE_TYPE)
 set(CPACK_PACKAGE_FILE_NAME "${CPACK_PACKAGE_NAME}-${CPACK_PACKAGE_VERSION}-${DISTRO_INFO}-${PACKAGE_TYPE}")
 
 include(CPack)
-
--- a/README.md
+++ b/README.md
@@ -73,7 +73,34 @@ For iHD_drv_video.so please export related LIBVA environment variables.
 export LIBVA_DRIVERS_PATH=/usr/lib/x86_64-linux-gnu/dri:/usr/lib/dri:<path-contains-iHD_drv_video.so>
 export LIBVA_DRIVER_NAME=iHD
 ```
+## Post-Install 
 
+To confirm that the installation conpleted correctly and that the new driver is loadable, use the following command and compare the VAAPI and Driver Version strings to those from `make install`
+
+    vainfo --display drm  --device /dev/dri/render[device] 
+
+
+Example on AlderLake-S iGPU (i7-12700k) loading Driver 22.3.1, VA-API 1.15 ( libva version 2.13)
+
+    vainfo --display drm  --device /dev/dri/renderD128
+
+ ```
+ libva info: VA-API version 1.15.0
+ libva info: Trying to open /usr/lib/x86_64-linux-gnu/dri/iHD_drv_video.so
+ libva info: Found init function __vaDriverInit_1_14
+ libva info: va_openDriver() returns 0
+ vainfo: VA-API version: 1.15 (libva 2.13.0)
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -80,8 +80,12 @@ endif()
 
 add_subdirectory(media_driver)
 
+message(STATUS "LIBVA_DRIVERS_PATH = ${LIBVA_DRIVERS_PATH}")
+
+set(LIBVA_DRIVERS_PATH "/usr/lib/x86_64-linux-gnu/dri/")
```
dvrogozh, line 85: You can not hardcode Debian/Ubuntu specific path, this will break other distributions and user setups if user has drivers in a different location.

CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ include(os_release_info.cmake)
 get_os_release_info(os_name os_version)
 
 if(NOT DEFINED CMAKE_INSTALL_PREFIX OR CMAKE_INSTALL_PREFIX STREQUAL "")
-    set(CMAKE_INSTALL_PREFIX "/usr/local" CACHE PATH "..." FORCE)
+    set(CMAKE_INSTALL_PREFIX "/usr/lib/x86_64-linux-gnu/" CACHE PATH "..." FORCE)
```
dvrogozh, line 31: I do not think that any manual adjustments of CMAKE_INSTALL_PREFIX should be here at all. Cmake does have defaults for this stuff which depends on the underlying OS (one default for Debian/Ubuntu, another for Fedora, etc.). If this default is not being picked up, then most probably something is just wrong in the cmake scripts. Most likely the following include should be done before doing anything with prefix:
```
include(GNUInstallDirs)
```

CMakeLists.txt
```diff
@@ -80,8 +80,12 @@ endif()
 
 add_subdirectory(media_driver)
 
+message(STATUS "LIBVA_DRIVERS_PATH = ${LIBVA_DRIVERS_PATH}")
+
+set(LIBVA_DRIVERS_PATH "/usr/lib/x86_64-linux-gnu/dri/")
+
 if("${LIBVA_DRIVERS_PATH}" STREQUAL "")
-    set(LIBVA_DRIVERS_PATH "${CMAKE_INSTALL_FULL_LIBDIR}/dri/")
+    set(LIBVA_DRIVERS_PATH "/usr/lib/x86_64-linux-gnu/dri/")
```
dvrogozh, line 88: same - you can't hardcode this.

README.md
```diff
@@ -43,12 +43,16 @@ then the workspace looks like below
 ```
 $ cd <workspace>/build_media
 ```
-6. 
+6.  Export the current LIBVA installation directory(ies)
```
dvrogozh, line 46: Imho, this should not be needed at all. At most this might be an optional step.

README.md
```diff
@@ -43,12 +43,16 @@ then the workspace looks like below
...
 ```
-6. 
+6.  Export the current LIBVA installation directory(ies)  
 ```
-$ cmake ../media-driver
+$  export LIBVA_DRIVERS_PATH=/usr/lib/x86_64-linux-gnu/dri:/usr/lib/dri
 ```
 7. 
 ```
+$ cmake ../media-driver -D[options]
```
dvrogozh, line 52: I don't see the point of adding `-D[options]`, that's not a cmake documentation and this makes command example not copy-pastable.

README.md
```diff
@@ -66,10 +70,37 @@ This will install the following files (e.g. on Ubuntu):
 
 For iHD_drv_video.so please export related LIBVA environment variables.
 ```
-export LIBVA_DRIVERS_PATH=<path-contains-iHD_drv_video.so>
+export LIBVA_DRIVERS_PATH=/usr/lib/x86_64-linux-gnu/dri:/usr/lib/dri:<path-contains-iHD_drv_video.so>
```
dvrogozh, line 73: We can't list all possible examples here. For example, you miss Fedora flavored systems. Please, revert.

README.md
```diff
@@ -66,10 +70,37 @@ This will install the following files (e.g. on Ubuntu):
...
 export LIBVA_DRIVER_NAME=iHD
 ```
+## Post-Install 
+
+To confirm that the installation conpleted correctly and that the new driver is loadable, use the following command and compare the VAAPI and Driver Version strings to those from `make install`
```
dvrogozh, line 78: s/conpleted/completed/

README.md
```diff
@@ -192,11 +223,11 @@ build configuration option.
 ## Known Issues and Limitations
 
 1. Intel(R) Media Driver for VAAPI is recommended to be built against gcc compiler v6.1
-or later, which officially supported C++11.
+or later, which officially supports C++11.
 
-2. SKL: Green or other incorrect color will be observed in output frames when using YV12/I420 as input format for csc/scaling/blending/rotation, etc. on Ubuntu 16.04 stock (with kernel 4.10). The issue can be addressed with the kernel patch: [WaEnableYV12BugFixInHalfSliceChicken7](https://cgit.freedesktop.org/drm-tip/commit/?id=0b71cea29fc29bbd8e9dd9c641fee6bd75f68274)
+2. SKL: Green or other incorrect colours will be observed in output frames when using YV12/I420 as input format for csc/scaling/blending/rotation, etc. on Ubuntu 16.04 stock (with kernel 4.10). The issue can be addressed with the kernel patch: [WaEnableYV12BugFixInHalfSliceChicken7](https://cgit.freedesktop.org/drm-tip/commit/?id=0b71cea29fc29bbd8e9dd9c641fee6bd75f68274)
```
dvrogozh, line 228: s/colours/colors/

dvrogozh (changes requested):

CMakeLists.txt
FCLC: Agreed that defaults should be avoided, I also like the idea of using `include(GNUInstallDirs)`.

Specifically on GNUInstallDirs, It's currently used/called at +56, after we set the users preferred config.  This may be the source of some of the issues. 


As for what the last case scenario default should be, since the docs target Ubuntu for their defaults configurations, following the Ubuntu standard seems reasonable.

README.md
FCLC: As of now it's a just in case/can't hurt to deal with the wrong directories being chosen by default. 

Regardless I'm fine with putting `(Optional)` in front of it

CMakeLists.txt
FCLC: as above

CMakeLists.txt
FCLC: Sounds good, will revert

README.md
FCLC: Was attempting to bring in similarities to https://github.com/intel/gmmlib#building

README.md
FCLC: Will do. Ideally a user already knows where their expected dri path is, and the installation path from cmake will be the same

README.md
FCLC: Should the standard be US english or standard/international english? 

There were a few typos where it was mixed and I didnt find a style guide

FCLC (commented):
Most changes made/resolved + a few I'd like clarity/feedback on around the readme

FCLC:
@dvrogozh Updated to reflect your requested changes, minus 2:

The first is that, in the event that we do have a base case, and the documentation typically references Ubuntu as that base case, should we be using the Ubuntu defaults? 

The second is a localization question. Should the standard be US English(simplified), or international English (traditional, as used in Canada, the UK, Australia etc.) I didn't see a style guide

FCLC:
For now I'm going to mark the comments as resolved since the hard coded issues were removed, it's now questions of preference. 

1. Since documentation typically refers to Ubuntu, should we use Ubuntu defaults? I think so. 

2. We don't have a standards guide AFAIK, so for the english, do we use US (simplified)  English, or Traditional (International) English?

Issue #449, opened 2018-12-07
Remove custom -DLINUX pre-compile definitions


dvrogozh:
Please, remove custom -DLINUX and any similar custom pre-compile definitions:
```
# fgrep -rsn "DLINUX"
Tools/MediaDriverTools/GenKrnBin/CMakeLists.txt:34:add_definitions(-DLINUX_)
Tools/MediaDriverTools/KernelBinToSource/CMakeLists.txt:25:add_definitions(-DLINUX_)
Tools/MediaDriverTools/KrnToHex_IGA/CMakeLists.txt:29:add_definitions(-DLINUX_)
media_driver/cmake/linux/media_compile_flags_linux.cmake:76:    -DLINUX=1
media_driver/cmake/linux/media_compile_flags_linux.cmake:77:    -DLINUX
# fgrep -rsn "D_LINUX"
# fgrep -rsn "D__LINUX"
#
```

This requires cmake-level change and code-level change. Instead, please, use standard `__linux__` macro mandated by Linux kernel: http://www.faqs.org/docs/Linux-HOWTO/GCC-HOWTO.html#INDEX.25.

We consider to add installation of cmrt header files (see #445) and there is a risk that sooner or later our custom define will end up in them. We need to prevent this.

ljia5:
PR created #936

PR #1396, opened 2022-04-25
[VP]Remove -DLINUX pre-compile definitions

ljia5:
Remove -DLINUX pre-compile definitions
Fixes: #449


dvrogozh (changes requested):
That's correct for Linux, but will cause issues and more work on BSD side. See https://github.com/intel/media-driver/pull/1296#issuecomment-982084427. I think we need a buy off from @jbeich and bsd maintainers to merge this in, I mean we could merge this in if they are ok to provide a follow up patch to handle BSD part.

PR #1399, opened 2022-04-28
Align pre-Gen12 hierarchical behavior with G12

Xiaogangli-intel:
Align pre-Gen12 hierarchical behavior with Gen12 when get BRC frame level according to Hierarchical level for HEVC


Xiaogangli-intel:
Check the source code again, we cannot align the pre-Gen12 behavior simply. Because kernel required input parameter is also different between pre-Gen12 and Gen12. So close this PR.

PR #1426, opened 2022-06-06
[Encode] fix AV1 caps for obu size bytes support

feiwan1:
Signed-off-by: Fei Wang <fei.w.wang@intel.com>

### DIFF
--- a/media_driver/linux/Xe_M/ddi/media_libva_caps_dg2.cpp
+++ b/media_driver/linux/Xe_M/ddi/media_libva_caps_dg2.cpp
@@ -74,6 +74,7 @@ VAStatus MediaLibvaCapsDG2::LoadAv1EncProfileEntrypoints()
         VAConfigAttribValEncAV1Ext2 attribValAV1ToolsExt2;
         memset(&attribValAV1ToolsExt2, 0, sizeof(attribValAV1ToolsExt2));
         attribValAV1ToolsExt2.bits.tile_size_bytes_minus1 = 3;
+        attribValAV1ToolsExt2.bits.obu_size_bytes_minus1  = 3;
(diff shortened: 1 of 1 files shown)


feiwan1:
@XinfengZhang @Xuhenry1 Could you help to review this PR? Thanks.

media_driver/linux/Xe_M/ddi/media_libva_caps_dg2.cpp
```diff
@@ -74,6 +74,7 @@ VAStatus MediaLibvaCapsDG2::LoadAv1EncProfileEntrypoints()
         VAConfigAttribValEncAV1Ext2 attribValAV1ToolsExt2;
         memset(&attribValAV1ToolsExt2, 0, sizeof(attribValAV1ToolsExt2));
         attribValAV1ToolsExt2.bits.tile_size_bytes_minus1 = 3;
+        attribValAV1ToolsExt2.bits.obu_size_bytes_minus1  = 3;
```
cchunbo, line 77: could you help to explain the value?

feiwan1: > could you help to explain the value?

This value indicate the fixed byte numbers reserved for obu_size instead of leb128() in spec. In BRC mode, media-driver will update  obu_size according to byte_offset_frame_hdr_obu_size with the fixed 4 bytes. Without this caps, up layer app may use leb128() to get the shortest bytes for obu_szie, which may caused overlap when media-driver update obu_szie.

cchunbo (approved):

Xiaogangli-intel (approved):
looks good to me

PR #1429, opened 2022-06-07
Optimize SCC capability report for Gen12 HEVC

### LINKED ISSUE #1391: Encoder: HEVC: The VAConfigAttribValEncHEVCBlockSizes and VAConfigAttribValEncHEVCFeatures is not precise for VME
## System information
- CPU information(`cat /proc/cpuinfo | grep "model name" | uniq`): Tiger Lake
- GPU information(`lspci -nn | grep -E 'VGA|isplay`): Gen12

### Describe the current behavior
The query results of VAConfigAttribValEncHEVCBlockSizes and VAConfigAttribValEncHEVCFeatures for VME(VAEntrypointEncSlice) are not very precise. For example, log2_max_coding_tree_block_size_minus3 is still set to 2, which causes the CTU size still be 32.
In fact, according to MSDK's result, this can be set to 3 and set CTU size to 64. This obsolete parameters do not block our encoding result, but the compression performance may not be good enough.

### Describe the expected behavior
Please update these parameters setting, for example, sync with MSDK's setting.

Xiaogangli-intel:
1. optimize call flow
2. add a new attribute list for SCC to correct some incorrect values
3. fixes #1391

### DIFF
--- a/media_driver/linux/Xe_M/ddi/media_libva_caps_dg2.cpp
+++ b/media_driver/linux/Xe_M/ddi/media_libva_caps_dg2.cpp
@@ -647,7 +647,9 @@ VAStatus MediaLibvaCapsDG2::CreateEncAttributes(
     attrib.type = VAConfigAttribRateControl;
     attrib.value = VA_RC_CQP;
     if (entrypoint != VAEntrypointEncSliceLP ||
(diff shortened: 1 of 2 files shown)


thomasli21801 (approved):

XinfengZhang:
@xhaihao , could you help to check whether it fix the issue

XinfengZhang (approved):

PR #1437, opened 2022-06-10
Fix uClibc build

bkuhls:
uClibc does not provide execinfo.h

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,12 @@ option (BUILD_CMRTLIB "Build and Install cmrtlib together with media driver" ON)
 
 option (ENABLE_PRODUCTION_KMD "Enable Production KMD header files" OFF)
 
+include(CheckIncludeFileCXX)
+check_include_file_cxx("execinfo.h" HAVE_EXECINFO)
+if (HAVE_EXECINFO)
+    add_definitions(-DHAVE_EXECINFO)
+endif()
+
 include(GNUInstallDirs)
 
 if (BUILD_CMRTLIB AND NOT CMAKE_WDDM_LINUX)
--- a/media_softlet/linux/common/os/osservice/mos_utilities_specific.cpp
+++ b/media_softlet/linux/common/os/osservice/mos_utilities_specific.cpp
@@ -34,7 +34,9 @@
 #include <signal.h>
 #include <unistd.h>  // fork
 #include <algorithm>
+#ifdef HAVE_EXECINFO
 #include <execinfo.h> // backtrace
+#endif
 #include <sys/types.h>
 #include <sys/stat.h>  // fstat
 #include <sys/ipc.h>  // System V IPC
@@ -2473,6 +2475,7 @@ void MosUtilities::MosTraceEvent(
                 MOS_FreeMemory(pTraceBuf);
             }
         }
+#ifdef HAVE_EXECINFO
         if (m_mosTraceFilter & (1ULL << TR_KEY_CALL_STACK))
         {
             // reserve space for header and stack size field.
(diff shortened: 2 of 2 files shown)


concatime:
Same with musl libc.

lindongw (approved):
looks good

bkuhls:
sent rebased patch needed due to https://github.com/intel/media-driver/commit/5be501a3807c630535407fb2c2c935ec78231694

bkuhls:
ubuntu-20.04 / gcc-11 build error is unrelated to this PR

concatime:
Can you use `FindBacktrace` please?

CMakeLists.txt
```diff
@@ -53,6 +53,12 @@ option (BUILD_CMRTLIB "Build and Install cmrtlib together with media driver" ON)
 
 option (ENABLE_PRODUCTION_KMD "Enable Production KMD header files" OFF)
 
+include(CheckIncludeFileCXX)
+check_include_file_cxx("execinfo.h" HAVE_EXECINFO)
```
dvrogozh, line 57: With that we exclude some functionality on certain libc versions. As suggested in one of the comments to #1437, it would be reasonable to use some search mechanism to find the provider for the features. Looks like https://cmake.org/cmake/help/v3.6/module/FindBacktrace.html provides such a way.

dvrogozh (changes requested):

CMakeLists.txt
dvrogozh: Thanks for pointing out. I was not aware about #1470.

> Combined with this patch libexecinfo can stay an optional dependency.

It seems it should stay as optional dependency, but we need to add detection mechanism which potentially may result in: "no one provides this". Thus, yes, I believe #1470 needs to be combined with this change.

bkuhls: > but we need to add detection mechanism which potentially may result in: "no one provides this". 

Isn't it what this PR is all about?

dvrogozh: Yes, but as of now this PR gives strong check on "execinfo.h" which might not exist and some support in media-driver will switch off. It seems the better way would be to use FindBacktrace which (if I read correctly) will provide Backtrace_HEADER file name to use to get backtrace(3), it might be "execinfo.h" or something else.

bkuhls: What about systems which do not provide execinfo.h and backtrace() at all? On these system the code added in https://github.com/intel/media-driver/commit/b66bb0c234b49508b8dd0ad54a6ea2c89b4dd83b needs to be switched-off.

dvrogozh: I think the code should be something like:
```
$ cat media_driver/media_top_cmake.cmake
...
set(MEDIA_LINK_EXTERNAL_LIBS "${PKG_PCIACCESS_LIBRARIES} m pthread dl")
find_package(Backtrace)
if(Backtrace_FOUND)
    configure_file(compat.h.in compat.h)
    set(MEDIA_LINK_EXTERNAL_LIBS "${MEDIA_LINK_EXTERNAL_LIBS} ${Backtrace_LIBRARY}")
endif()

$ cat compat.h.in
#cmakedefine01 Backtrace_FOUND
#if Backtrace_FOUND
#include <${Backtrace_HEADER}>
#endif

$ cat media_softlet/linux/common/os/osservice/mos_utilities_specific.cpp
#include "compat.h"

#if Backtrace_FOUND
int num = backtrace(stack, ((sizeof(traceBuf)-nLen)/sizeof(void *)));
#endif
```

bkuhls: > I think the code should be something like:

Just an honest reply: My coding skills are not enough to come up with such code, sorry.

dvrogozh: Here is how I would see that done: #1536. I did not have a chance to try it under different configurations though. Appreciate if you can give it a try for your case and provide feedback.

bkuhls:
Fixed by https://github.com/intel/media-driver/pull/1536

PR #1446, opened 2022-06-17
Add build options to disable SSE intrinsics usage

XinfengZhang:
cmake ../media-driver/ -DENABLE_PRODUCTION_KMD=ON -DMEDIA_BUILD_FATAL_WARNINGS=OFF -DBUILD_CMRTLIB=OFF -DENABLE_X86_INTRINSICS=OFF

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,10 @@ cmake_dependent_option( BUILD_KERNELS
 option (BUILD_CMRTLIB "Build and Install cmrtlib together with media driver" ON)
 
 option (ENABLE_PRODUCTION_KMD "Enable Production KMD header files" OFF)
+option (ENABLE_X86_INTRINSICS "Enable x86 intrinsics" ON)
+if(ENABLE_X86_INTRINSICS)
+    add_definitions(-DIGFX_ENABLE_X86_INTRINSICS)
+endif()
 
 include(GNUInstallDirs)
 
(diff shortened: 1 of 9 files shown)


dvrogozh (changes requested):
Build fails. Why do we need this option to disable intrinsic? What's the rationale behind that?

HeFan2017 (approved):
cm-related filed LGTM

CMakeLists.txt
```diff
@@ -52,6 +52,10 @@ cmake_dependent_option( BUILD_KERNELS
 option (BUILD_CMRTLIB "Build and Install cmrtlib together with media driver" ON)
 
 option (ENABLE_PRODUCTION_KMD "Enable Production KMD header files" OFF)
+option (ENABLE_X86_INTRINSICS "Enable x86 intrinsics" ON)
```
dvrogozh, line 55: This should not be configuration step check. Instead, this should be runtime check whether some features are supported or not. If not supported - disable these feature or select a supported path (since few implementations typically possible).

dvrogozh (changes requested):

CMakeLists.txt
eero-t: Intel has `_may_i_use_cpu_feature()` intrinsics function: https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_may_i_use_cpu_feature

GCC has `__builtin_cpu_supports()` for runtime checks: https://gcc.gnu.org/onlinedocs/gcc-13.1.0/gcc/x86-Built-in-Functions.html

And other compilers have similar features.

Google provides "cpu_features" library (under Apache license).

PR #1470, opened 2022-08-12
Fix build with libexecinfo

ffontaine:
Fix the following build failure with musl or uclibc-ng and libexecinfo:

```
/home/giuliobenetti/autobuild/run/instance-2/output-1/build/intel-mediadriver-22.5.1/media_softlet/linux/common/os/osservice/mos_utilities_specific.cpp:2481: undefined reference to `backtrace'
```

Fixes:
 - http://autobuild.buildroot.org/results/3316e39c5113bb7600374eda45497a87c9ac9873

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/media_driver/media_top_cmake.cmake
+++ b/media_driver/media_top_cmake.cmake
@@ -271,10 +271,16 @@ set_target_properties(${LIB_NAME} PROPERTIES LINK_FLAGS ${MEDIA_LINK_FLAGS})
 set_target_properties(${LIB_NAME}        PROPERTIES PREFIX "")
 set_target_properties(${LIB_NAME_STATIC} PROPERTIES PREFIX "")
 
+set(MEDIA_LINK_EXTERNAL_LIBS "${PKG_PCIACCESS_LIBRARIES} m pthread dl")
+find_package(Backtrace)
+if(Backtrace_FOUND)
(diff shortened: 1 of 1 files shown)


motorto:
isn;t this duplicate of #1437 ?

ffontaine:
This is not a duplicate of #1437.
- #1437 is needed to fix build **without** libexecinfo (i.e. on uclibc or musl **without** https://github.com/mikroskeem/libexecinfo)
- #1470 is needed  to fix build **with** libexecinfo (i.e. on uclibc or musl **with** https://github.com/mikroskeem/libexecinfo)

dvrogozh (changes requested):
> This is not a duplicate of https://github.com/intel/media-driver/pull/1437

Both #1470 and #1437 are part of the common story which is: there might be different providers of backtrace(3) depending on underlying runtime. There might be different headers needed to fetch definition of backtrace(3) to compile and different libraries to link against. Better if all these requirements will be addressed at once, not separately.

dvrogozh:
> Better if all these requirements will be addressed at once, not separately.

For that I propose to consider #1536. I did not have a chance to try it under different configurations though. Appreciate if you can give it a try for your case and provide feedback.

bkuhls:
Fixed by https://github.com/intel/media-driver/pull/1536, this PR can be closed.

PR #1475, opened 2022-08-19
In VD+SFC case, Honor the region height from the setting of API inste…

FurongZhang:
…ad of overriding.

Signed-off-by: Furong Zhang <furong.zhang@intel.com>


XinfengZhang (approved):

yawenyan (approved):
LGTM

XinfengZhang (approved):

PR #1477, opened 2022-08-23
[Encode] Fix MB QP Data shadow buffer has potential alignment issue

TangZhiZhen:
If the base address of the staging buffer is not 64 bytes aligned,
it can bring in performance issue when memcpy between staging buffer
 and the graphics buffer.

Signed-off-by: Zhizhen Tang <zhizhen.tang@intel.com>


thomasli21801 (approved):

Xiaogangli-intel (approved):

cchunbo:
Hi, Zhizhen

Could you help to share some more details? seems like the system memory base address alignment will impact the performance, is there any perf data to understand the benefit? great thanks

PR #1485, opened 2022-09-08
add support for xyuv in vp 

xhaihao:
XYUV is used in 4:4:4 8bit content decoding and encoding on Linux, so we need to support this format in vp too


FurongZhang (commented):
Looks good to me. Thanks @xhaihao for the code changes.

FurongZhang (approved):

PR #1500, opened 2022-09-30
Disable object capture as default for dGfx

Xiaogangli-intel:
Disable object capture as default for recoverable context on dGfx , it's a short-term solution because a KMD change in drm-tip.


XinfengZhang (commented):
could we change the logic to if local_memory exist , the exec_capture = false?

Jexu (approved):

XinfengZhang (commented):
two concerns:
1. why we add a variable named object_capture_disabled instead of object_capture?
2. suppose MTL also need to disable object_capture , so , we will have next change to cover it?

XinfengZhang (approved):

XinfengZhang:
closed by https://github.com/intel/media-driver/commit/f8812f26a35755714d6386df969b0e99fbab56ef

Issue #1507, opened 2022-10-06
[Bug]: Use of uninitialized memory.


stolk:
### Which component impacted?

Decode, Encode, Video Processing

### Is it regression? Good in old configuration?

No, this issue exist a long time

### What happened?

The driver will access unintialized memory, leading to undefined behaviour.

This error can be caught with running valgrind.

```
$ valgrind ./vainfo 
==33231== Memcheck, a memory error detector
==33231== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==33231== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==33231== Command: ./vainfo
==33231== 
Trying display: wayland
Trying display: x11
libva info: VA-API version 1.17.0
libva info: Trying to open /usr/local/lib/dri/iHD_drv_video.so
libva info: Found init function __vaDriverInit_1_17
==33231== Conditional jump or move depends on uninitialised value(s)
==33231==    at 0x484A830: rindex (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==33231==    by 0x52B4C49: parse_separate_sysfs_files(int, int, _drmPciDeviceInfo*, bool) (drm_device.h:596)
==33231==    by 0x52B520D: drmParsePciDeviceInfo(int, int, _drmPciDeviceInfo*, unsigned int) (drm_device.h:699)
==33231==    by 0x52B5573: drmProcessPciDevice(_drmDevice**, char const*, int, int, int, bool, unsigned int) (drm_device.h:823)
==33231==    by 0x52B5FCC: process_device(_drmDevice**, char const*, int, bool, unsigned int) (drm_device.h:1079)
==33231==    by 0x52B6401: drmGetDevice2(int, unsigned int, _drmDevice**) (drm_device.h:1252)
==33231==    by 0x52B65C6: drmGetDevice(int, _drmDevice**) (drm_device.h:1298)
==33231==    by 0x52B9EC6: MosInterface::GetAdapterBDF(_MOS_OS_CONTEXT*, _ADAPTER_BDF_*) (mos_interface.cpp:613)
==33231==    by 0x5305EEB: OsContextSpecificNext::Init(void*) (mos_context_specific_next.cpp:178)
==33231==    by 0x52B6E39: MosInterface::CreateOsDeviceContext(void*, OsContextNext**) (mos_interface.cpp:127)
[...]

PR #1508, opened 2022-10-06
Fix access of unintialized memory. Issue #1507

stolk:
As confirmed by valgrind, with these added memset() operations, the driver will no longer access uninitialized memory.

Also see: `$ man 2 readlink` 

The destination buffer is not null-terminated.

To test: use the vainfo tool, running through valgrind, and compare behaviour before and after patch.


cmrtlib/linux/hardware/drm_device.h
```diff
@@ -950,6 +951,7 @@ static int drmParseSubsystemType(int maj, int min)
     snprintf(path, PATH_MAX, "/sys/dev/char/%d:%d/device/subsystem",
         maj, min);
 
+    memset(link, 0, sizeof(link));
```
XinfengZhang, line 954: link should be initialized by line 937？
https://github.com/intel/media-driver/pull/1508/files#diff-4ef24d4d6996a3f7d92a338cfee90f859972182c715f2adfeb4e3605ece1a722R937

stolk:
I need to re-do this pull request... My attempt to rebase has made a mess of the commit history. My apologies. I will make a new one.

PR #1510, opened 2022-10-11
[CI] add libx11-xcb-dev libxcb-dri3-dev to trigger X11 backend build

Xiaogangli-intel:
Fix x11 libraries missing issue which will cause build fail

Signed-off-by: Xiaogang Li <xiaogang.li@intel.com>


Xiaogangli-intel:
conflict with Carl's PR, close it.

PR #1511, opened 2022-10-12
Fix reads from uninitialized memory.

stolk:
This fixes issue #1507 

readlink() will not place a terminating 0, which means that we cannot use unintialized memory as a destination.

NOTE: it is not sufficient to write a null-string to the buffer, because that only sets the first character to 0. Every byte in the destination string should be 0, so that when readlink() has written content, it is always null-terminated.

### DIFF
--- a/cmrtlib/linux/hardware/drm_device.h
+++ b/cmrtlib/linux/hardware/drm_device.h
@@ -584,6 +584,7 @@ static int parse_separate_sysfs_files(int maj, int min,
 
     snprintf(resourcename, sizeof(resourcename), "%s/resource", pci_path);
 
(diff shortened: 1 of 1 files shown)


cmrtlib/linux/hardware/drm_device.h
```diff
@@ -950,6 +951,7 @@ static int drmParseSubsystemType(int maj, int min)
     snprintf(path, PATH_MAX, "/sys/dev/char/%d:%d/device/subsystem",
         maj, min);
 
+    memset(link, 0, sizeof(link));
```
XinfengZhang, line 954: link is initialized at line 937?

Jexu (approved):

PR #1545, opened 2022-11-13
fix build error with gcc-4.9

bkuhls:
gcc-4.9 does not like {0}:

error: array must be initialized with a brace-enclosed initializer


bkuhls:
Closed, media-driver needs gcc >= 6.1

Issue #1553, opened 2022-11-21
[Bug]:  error: format '%lu' expects argument of type 'long unsigned int', but argument 4 has type 'uint64_t' {aka 'long long unsigned int'} [-Werror=format=]


SuperSandro2000:
### Which component impacted?

Build

### Is it regression? Good in old configuration?

Yes, it's good in old version

### What happened?

After updating to 2.6.3 building on i686-linux fails with the following error:

```
/build/source/media_softlet/linux/common/ddi/media_libva_util_next.cpp: In static member function 'static void MediaLibvaUtilNext::MediaPrintFps()':
/build/source/media_softlet/linux/common/ddi/media_libva_util_next.cpp:2199:47: error: format '%lu' expects argument of type 'long unsigned int', but argument 4 has type 'uint64_t' {aka 'long long unsigned int'} [-Werror=format=]
 2199 |         sprintf(temp,"FPS:%6.4f, Interval:%11lu\n", fps,((uint64_t)tv2.tv_sec)*1000 + (tv2.tv_usec/1000));
      |                                           ~~~~^         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                               |                                     |
      |                                               long unsigned int                     uint64_t {aka long long unsigned int}
```

### What's the usage scenario when you are seeing the problem?

Others

### What impacted?

I am maintaining the intel-media-driver package for NixOS and the 32 bit variant is required for hardware acceleration in combination with Steam.

### Do you want to contribute a patch to fix the issue?

No.

### Other

This issue is similar to #1493 but since my C knowledge is very limited I don't know how to apply that to %11lu.

intel-mediadev:
Auto Created VSMGWL-59303 for further analysis.

cqian2:
"%lu" is for unsigned long, in current gcc (gcc12) 32bit compiler, unsigned long takes 4 bytes, while unsigned long long takes 8 bytes, 8 bytes -> 4 bytes has the data lossing risk, though number "11" was added in "%11lu" as the original purpose was only low11 bits data were cared (needed to be printed), but the compiler would not recognize this. 
While in current gcc (gcc12) 64bit compiler, unsigned long takes 8 bytes, 8 bytes -> 8 bytes has no risk.
Whether or not "11" is added has no impact on the compiler behavior.

cqian2:
[...]

PR #1557, opened 2022-11-23
Fix print message format '%lu' issue

cqian2:
Fix issue 1553


SuperSandro2000:
Rebased on top of 22.6.3 because it didn't apply straight away and I can confirm that it works.

cqian2:
Fixes #1553

PR #1569, opened 2022-12-04
[Media Common] vaGetImage optimization on D-card.

### LINKED ISSUE #1537: [Bug]: vaGetImage is very slow on dGPU when decode a 4K 10bit clip
### Which component impacted?

Decode

### Is it regression? Good in old configuration?

_No response_

### What happened?

1. On linux, using gstreamer va plugins to decode a 4K p010 format bitstream:<pre> gst-launch-1.0 -vf filesrc location=4k_p010.h265 ! h265parse ! vah265dec ! fakesink</pre>
2. The whole process for running the cmd line above is extremely slow (**~3sec for one frame**). Debugging into it, I found the slow step happened at GPU to CPU copy using vaGetImage. I don't see it slow on iGPU. Also, when I tried to use vaDeriveImage, the speed is normally quick.

### What's the usage scenario when you are seeing the problem?

Others

### What impacted?

Decoding a 4K ( with formats expect NV12) clip is a quite basic function. We expect faster decoding on dGPU instead of much slower than iGPU.

### Debug Information

_No response_

### Do you want to contribute a patch to fix the issue?

_No response_

MicroYY:
Fix #1537

On discrete cards which have local memory, map shadow buffer instead of surface bo. Otherwise cpu memcpy will be very slow.

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -5260,9 +5260,6 @@ static VAStatus DdiMedia_CopySurfaceToImage(
         }
     }
 
-    if (image->format.fourcc != VA_FOURCC_NV12)
-       flag = flag | MOS_LOCKFLAG_NO_SWIZZLE;
-
     void* surfData = DdiMediaUtil_LockSurface(surface, flag);
 
     if (surfData == nullptr)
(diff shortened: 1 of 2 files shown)


fenhu (approved):

media_driver/linux/common/ddi/media_libva.cpp
```diff
@@ -5260,7 +5260,9 @@ static VAStatus DdiMedia_CopySurfaceToImage(
         }
     }
 
-    if (image->format.fourcc != VA_FOURCC_NV12)
+    // HW swizzle doesn't support 422 format. Use SW swizzle instead
```
fenhu, line 5263: seems also  VA_FOURCC_444P.

fenhu (approved):

MicroYY:
> Hi Jay, Please help to make the same fix on the refactored path. MediaLibvaInterfaceNext::CopySurfaceToImage
> 
> Thank you.

Ported change to softlet.

XinfengZhang (approved):

PR #1589, opened 2023-01-09
[Media Common] SW swizzling regression fix for Gen8/9/10

MicroYY:
Signed-off-by: Jay Yang <jay.yang@intel.com>
Fix: https://github.com/intel/media-driver/issues/1590

Add sw swizzling sku for Gen8/9/10, without which there will be perf
drop because mos_gem_bo_map_unsynchronized is called.

### DIFF
--- a/media_driver/linux/gen9/ddi/media_sku_wa_g9.cpp
+++ b/media_driver/linux/gen9/ddi/media_sku_wa_g9.cpp
@@ -206,6 +206,8 @@ static bool InitSklMediaSku(struct GfxDeviceInfo *devInfo,
 
     MEDIA_WR_SKU(skuTable, FtrTileY, 1);
 
+    MEDIA_WR_SKU(skuTable, FtrUseSwSwizzling, 1);
+
     return true;
 }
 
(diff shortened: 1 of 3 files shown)


media_driver/linux/gen9/ddi/media_sku_wa_g9.cpp
```diff
@@ -540,6 +546,8 @@ static bool InitGlkMediaSku(struct GfxDeviceInfo *devInfo,
 
     MEDIA_WR_SKU(skuTable, FtrHDR, 1);
 
+    MEDIA_WR_SKU(skuTable, FtrUseSwSwizzling, 1);
```
XinfengZhang, line 549: GLK is small core , is using map_gtt, when this WA take effect?

MicroYY: Used before mapping
https://github.com/intel/media-driver/blob/3f754f70437309a95682368d43e2c6708631f1b3/media_driver/linux/common/ddi/media_libva_util.cpp#L1209

XinfengZhang (approved):

Issue #1540, opened 2022-11-03
[Bug]: Building media-driver with {AVC,HEVC}Encode_{VME,VDEnc}_Supported="no" results in linking failure


frantisekz:
### Which component impacted?

Build

### What happened?

Build of media-driver 22.6.1 with:

  -DENABLE_NONFREE_KERNELS:BOOL=OFF \
  -DENABLE_KERNELS:BOOL=OFF \
  -DBUILD_KERNELS:BOOL=OFF \
  -DMEDIA_BUILD_FATAL_WARNINGS=OFF \
  -DBUILD_SHARED_LIBS:BOOL=OFF \
  -DGEN10:BOOL=OFF \
  -DAVC_Encode_VME_Supported="no" \
  -DAVC_Encode_VDEnc_Supported="no" \
  -DHEVC_Encode_VME_Supported="no" \
  -DHEVC_Encode_VDEnc_Supported="no" \
  -G Ninja

results with linker failure:

```
/builddir/build/BUILD/media-driver-intel-media-22.6.1/media_driver/agnostic/common/vp/hal/vphal_render_vebox_base.h:568: warning: type 'struct _VPHAL_VEBOX_SURFACE_STATE_CMD_PARAMS' violates the C++ One Definition Rule [-Wodr]
  568 | typedef struct _VPHAL_VEBOX_SURFACE_STATE_CMD_PARAMS
      | 
/builddir/build/BUILD/media-driver-intel-media-22.6.1/media_softlet/agnostic/common/vp/hal/packet/vp_vebox_common.h:31: note: a different type is defined in another translation unit
   31 | typedef struct _VPHAL_VEBOX_SURFACE_STATE_CMD_PARAMS
      | 
/builddir/build/BUILD/media-driver-intel-media-22.6.1/media_driver/agnostic/common/vp/hal/vphal_render_vebox_base.h:570: note: the first difference of corresponding definitions is field 'pSurfInput'
  570 |     PVPHAL_SURFACE                  pSurfInput;
      | 
/builddir/build/BUILD/media-driver-intel-media-22.6.1/media_softlet/agnostic/common/vp/hal/packet/vp_vebox_common.h:33: note: a field of same name but different type is defined in another translation unit
   33 |     PVP_SURFACE                     pSurfInput;
      | 
[...]

PR #1594, opened 2023-01-13
[Encode] Fix build issue when setting HEVC_Encode_Supported="no"

bai-isaac:
This commit fix the compilation error when setting HEVC_Encode_Supported="no"
Fix for https://github.com/intel/media-driver/issues/1540

### DIFF
--- a/media_driver/agnostic/gen11/codec/hal/codechal_huc_cmd_initializer_g11.h
+++ b/media_driver/agnostic/gen11/codec/hal/codechal_huc_cmd_initializer_g11.h
@@ -177,8 +177,6 @@ public:
     virtual MOS_STATUS CmdInitializerAllocateResources(
         CodechalHwInterface*    m_hwInterface) override;
 
-    virtual MOS_STATUS CmdInitializerSetDmem(bool brcEnabled) override;
-
     //!
     //! \brief    Free Resources
     //!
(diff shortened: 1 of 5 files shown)


media_driver/agnostic/gen11/codec/hal/codechal_huc_cmd_initializer_g11.h
```diff
@@ -199,6 +197,9 @@ class CodechalCmdInitializerG11 : public CodechalCmdInitializer
         PMOS_COMMAND_BUFFER cmdBuffer,
         HucCopyParams* params);
 
+#if defined (_HEVC_ENCODE_VME_SUPPORTED) || defined (_HEVC_ENCODE_VDENC_SUPPORTED)
+    virtual MOS_STATUS CmdInitializerSetDmem(bool brcEnabled) override;
```
linlinz1, line 201: should the functions call this also check the hevc macro?

bai-isaac: These functions only used by hevc in codechal_vdenc_hevc_g11.cpp. So we don't need more hevc macro.
Thanks!

linlinz1 (approved):

media_driver/agnostic/gen11/codec/hal/codechal_huc_cmd_initializer_g11.h
bai-isaac: codechal_huc_cmd_initializer_g11.cpp and codechal_huc_cmd_initializer.cpp have use hevc macro for these functions.

PR #1604, opened 2023-02-02
Disable Gen10 driver code when -DGEN10=OFF

mattst88:
Allowing the driver code to be disabled for Gen10 reduces the binary size by an additional 6.6 MiB over the 4 MiB reduction when removing the media kernels alone.

````
45485460  2374412 38736  47898608  2dadff0  iHD_drv_video.so # -DGEN10=ON
41422146  2237132 38768  43698046  29ac77e  iHD_drv_video.so # -DGEN10=OFF before this patch
34483374  2219940 38704  36742018  230a382  iHD_drv_video.so # -DGEN10=OFF after this patch
````

This change was enabled by the compile fix in commit e47460dc0363 ("Encode] Fix build issue when setting -DGEN10=OFF").

Closes: https://github.com/intel/media-driver/issues/1065

### DIFF
--- a/media_driver/agnostic/media_srcs.cmake
+++ b/media_driver/agnostic/media_srcs.cmake
@@ -326,7 +326,7 @@ if(ENABLE_REQUIRED_GEN_CODE OR GEN9_KBL)
     media_include_subdirectory(gen9_kbl)
 endif()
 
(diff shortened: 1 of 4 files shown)


mattst88:
Cc: @XinfengZhang

XinfengZhang (approved):

PR #1605, opened 2023-02-03
 Disable Gen9 driver code when -DGEN9=OFF

mattst88:
Disable Gen9 driver code when -DGEN9=OFF

Allowing the driver code to be disabled for Gen9 reduces the
binary size by an additional 2.1 MiB over the 5.0 MiB reduction when
removing the media kernels alone.

````
41573236  2237132 38768  43849136  29d15b0  iHD_drv_video.so # -DGEN9=ON
36324898  2237132 38768  38600798  24d005e  iHD_drv_video.so # -DGEN9=OFF before this patch
34124420  2196548 38384  36359352  22accb8  iHD_drv_video.so # -DGEN9=OFF after this patch
````

### DIFF
--- a/media_driver/agnostic/media_srcs.cmake
+++ b/media_driver/agnostic/media_srcs.cmake
@@ -298,31 +298,31 @@ if(GEN8_BDW)
     media_include_subdirectory(gen8_bdw)
 endif()
 
-if(ENABLE_REQUIRED_GEN_CODE OR GEN9)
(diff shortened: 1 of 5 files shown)


mattst88:
Cc: @XinfengZhang

XinfengZhang (approved):

mattst88:
Trivial rebase conflict resolved.

PR #1606, opened 2023-02-03
Disable Gen11 driver code when -DGEN11=OFF

mattst88:
Disable Gen11 driver code when -DGEN11=OFF

Allowing the driver code to be disabled for Gen11 reduces the
binary size by an additional 1.0 MiB over the 3.8 MiB reduction when
removing the media kernels alone.

````
41573236  2237132 38768  43849136  29d15b0  iHD_drv_video.so # -DGEN11=ON
38439404  1417324 38832  39895560  260c208  iHD_drv_video.so # -DGEN11=OFF before this patch
37402536  1387428 38672  38828636  2507a5c  iHD_drv_video.so # -DGEN11=OFF after this patch
````

### DIFF
--- a/media_driver/agnostic/media_srcs.cmake
+++ b/media_driver/agnostic/media_srcs.cmake
@@ -330,15 +330,15 @@ if(GEN10)
     media_include_subdirectory(gen10)
 endif()
 
-if(ENABLE_REQUIRED_GEN_CODE OR GEN11)
+if(GEN11)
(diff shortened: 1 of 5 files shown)


mattst88:
Cc: @XinfengZhang

XinfengZhang (approved):

mattst88:
Trivial rebase conflict resolved.

PR #1607, opened 2023-02-03
Remove stale cmake subdirectory

mattst88:
I believe this was missed in commit df7b0bd796ae ("[Encode] move codechal setting create from hal to ddi")


mattst88:
Cc: @linlinz1

mattst88:
Cc: @XinfengZhang

It should not be this hard to get a trivial patch in...

XinfengZhang (approved):

linlinz1 (approved):

PR #1613, opened 2023-02-16
use possible upstream definition of CCS modifier

Dylan-debug:
description: use possible upstream definition of CCS modifier

Signed-off-by: Jianxing Cai [jianxing.cai@intel.com](mailto:jianxing.cai@intel.com)

### DIFF
--- a/media_softlet/linux/Xe_M_plus/ddi/media_libva_caps_mtl_base.cpp
+++ b/media_softlet/linux/Xe_M_plus/ddi/media_libva_caps_mtl_base.cpp
@@ -99,11 +99,11 @@ const VAImageFormat m_supportedImageformatsXe_Lpm_Plus_Base[] =
     (fourcc_mod_code(INTEL, (val)) | INTEL_PRELIM_ID_FLAG)
 
 /* this definition is to avoid duplicate drm_fourcc.h this file is updated seldom */
-#ifndef PRELIM_I915_FORMAT_MOD_4_TILED_MTL_MC_CCS
-#define PRELIM_I915_FORMAT_MOD_4_TILED_MTL_MC_CCS    intel_prelim_fourcc_mod_code(17)
+#ifndef I915_FORMAT_MOD_4_TILED_MTL_MC_CCS
+#define I915_FORMAT_MOD_4_TILED_MTL_MC_CCS    fourcc_mod_code(INTEL, 14)
 #endif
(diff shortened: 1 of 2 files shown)


XinfengZhang (approved):

dvrogozh (changes requested):
media-driver has very specific way to work with i915 uAPI. Depending on the build flag it uses either upstream or prelim uAPI. How respective headers should be updated is described in readme documents:
* https://github.com/intel/media-driver/tree/master/media_softlet/linux/common/os/i915_production/include
* https://github.com/intel/media-driver/tree/master/media_softlet/linux/common/os/i915/include/uapi

Self-defining i915 uAPI outside of the above procedure should never happen since this is a potential risk to break driver compatibility with the kernel. If uAPI is missing on  the official i915 headers then such a change can not be merged into media driver at the first place. I.e. it is wrong from the very beginning that such a code appeared in the media driver at the first place.

To proceed with this PR:
1. Update prelim and/or upstream i915 header files as described in above readme documents
2. If any definitions you need will still be missing after update - STOP and raise a red flag
3. Assuming definitions are there, we need to understand compatibility story behind the change. Basically you are tryin to update from prelim API to upstream API which risks to break compatibility with Intel custom kernel for MTL. This needs to be carefully evaluated. and maybe #ifdef protection might need to be used to differentiate the build accordingly.

media_softlet/linux/Xe_M_plus/ddi/media_libva_caps_mtl_base.cpp
```diff
@@ -99,11 +99,11 @@ const VAImageFormat m_supportedImageformatsXe_Lpm_Plus_Base[] =
     (fourcc_mod_code(INTEL, (val)) | INTEL_PRELIM_ID_FLAG)
 
 /* this definition is to avoid duplicate drm_fourcc.h this file is updated seldom */
-#ifndef PRELIM_I915_FORMAT_MOD_4_TILED_MTL_MC_CCS
-#define PRELIM_I915_FORMAT_MOD_4_TILED_MTL_MC_CCS    intel_prelim_fourcc_mod_code(17)
+#ifndef I915_FORMAT_MOD_4_TILED_MTL_MC_CCS
```
dvrogozh, line 102: These self-definitions should be completely pruned ASAP. Definitions from i915 header files driver carries should be used instead.

PR #1630, opened 2023-03-21
[Encode] Fix the HEVC ICQ capability on DG2

nyanmisaka:
Tested with FFmpeg `hevc_{qsv,vaapi} -global_quality 25` on DG2/A380.

Also the AV1 ICQ need extra handling since it is not implemented yet.


OttoYang (approved):
LGTM

nyanmisaka:
@OttoYang It seems your internal test case needs an update for this commit. HEVC ICQ is now `MFX_ERR_NONE`.

Issue #1628, opened 2023-03-13
[Bug][DG2][VPP]: tonemap_vaapi extremely slow on certain input size 1280*(722 vs 720)


nyanmisaka:
### Which component impacted?

Video Processing

### Is it regression? Good in old configuration?

No, this issue exist a long time

### What happened?

1. Linux 6.2.5, FFmpeg 5/6/cartwheel, Arc A380, ReBAR On
2. libva 2.17, gmmlib 22.3.4, media-driver 23.1.3
3. Run FFmpeg cli and watch the fps:

Two bad resolutions: 1280x**722** and 1280x**692** - Results in very slow FPS - < 5 FPS.

```
frame=   22 fps=3.8 q=-0.0 Lsize=N/A time=00:00:00.35 bitrate=N/A speed=0.0606x
```

But 1280x**720** and 1280x**694** works just fine - hundreds FPS.

Other resolutions may also affected but I haven't found the pattern yet.

```
ffmpeg -hwaccel vaapi -hwaccel_output_format vaapi -i ANY_10Bit_HDR_CLIP.mp4 -vf scale_vaapi=1280:722,tonemap_vaapi=format=nv12:t=bt709:m=bt709:p=bt709 -f null -
```

```
ffmpeg -hwaccel vaapi -hwaccel_output_format vaapi -i ANY_10Bit_HDR_CLIP.mp4 -vf scale_vaapi=1280:692,tonemap_vaapi=format=nv12:t=bt709:m=bt709:p=bt709 -f null -
```

### What's the usage scenario when you are seeing the problem?

Transcode for media delivery

### What impacted?

HDR10 to SDR tone-mapping in media delivery.

### Debug Information

1. libva 2.17, gmmlib 22.3.4, media-driver 23.1.3
2. yes renderD128
3. [vainfo.log](https://github.com/intel/media-driver/files/10958206/vainfo.log)
4. [libva_trace.log.141251.zip](https://github.com/intel/media-driver/files/10958240/libva_trace.log.141251.zip)
5. no i915 hang

Guess it's due to the incorrect image size alignment on DG2 platform?
Older platforms are not affected such as KBLx and TGLx.

Other common VPP filters are not affected too.
```
 ... deinterlace_vaapi V->V       Deinterlacing of VAAPI surfaces
 ... denoise_vaapi     V->V       VAAPI VPP for de-noise
 ... overlay_vaapi     VV->V      Overlay one video on top of another
[...]

PR #1646, opened 2023-04-20
[VP] Fix tonemap vaapi extremely slow on certain input size issue

VickyZengg:
Fixes: intel/media-driver#1628

Sku table did not init on Linux, it need do 8x alignment.


MicroYY (approved):

FurongZhang (approved):
It looks good to me.

PR #1675, opened 2023-06-01
GetModifier linear format

MicroYY:


XinfengZhang (approved):
actually , I dont understand the difference between NONE and Linear, but I am ok with the change

Issue #1672, opened 2023-05-26
 [Bug][ADL-S 13th Gen][VPP]: tonemap_vaapi fails on certain input size 1280*(722 vs 720)


almightiest:
### Which component impacted?

Video Processing

### Is it regression? Good in old configuration?

Yes, it's good in old version

### What happened?

Use a file with HDR10, like a sample file from [here](https://www.demolandia.net/downloads.html?id=654828652) - click on "<star> Download the UHD HDR10 - Sony Bravia OLED File <star>" link on that page to download/save.

Then run the following 2 commands on i7-13700k or similar iGPU on ubuntu lunar with kernel 6.2 and the latest master media-driver:
1. 1920x802 scale_vaapi - **fails input/output error on 23.2.3 (master) - slow fps but no failures on 23.1.6**
`ffmpeg -analyzeduration 200M -init_hw_device vaapi=va:,driver=iHD,kernel_driver=i915 -init_hw_device qsv=qs@va -filter_hw_device qs -hwaccel vaapi -hwaccel_output_format vaapi -autorotate 0 -i file:"/path/to/sample.mkv" -autoscale 0 -vf "setparams=color_primaries=bt2020:color_trc=smpte2084:colorspace=bt2020nc,scale_vaapi=w=1920:h=802:extra_hw_frames=24,tonemap_vaapi=format=nv12:p=bt709:t=bt709:m=bt709" -f null -`

2. 1920x800 scale_vaapi - **succeeds >200fps on 23.2.3 (master) and 23.1.6**
`ffmpeg -analyzeduration 200M -init_hw_device vaapi=va:,driver=iHD,kernel_driver=i915 -init_hw_device qsv=qs@va -filter_hw_device qs -hwaccel vaapi -hwaccel_output_format vaapi -autorotate 0 -i file:"/path/to/sample.mkv" -autoscale 0 -vf "setparams=color_primaries=bt2020:color_trc=smpte2084:colorspace=bt2020nc,scale_vaapi=w=1920:h=800:extra_hw_frames=24,tonemap_vaapi=format=nv12:p=bt709:t=bt709:m=bt709" -f null -`

### What's the usage scenario when you are seeing the problem?

Transcode for media delivery

### What impacted?

Any video with HDR* which requires tone mapping to SDR will fail with various croppings, when using VAAPI.

### Debug Information

[...]

PR #1679, opened 2023-06-06
Fix tonemap vaapi on certain input size issue

VickyZengg:
Fixes: https://github.com/intel/media-driver/issues/1672

Sku table did not init on Linux ADL-S and RPL-S, and it didn't update 3dlutenable params, it need do 8x alignment.


FurongZhang (approved):
Looks good to me.

Issue #1654, opened 2023-05-03
[Bug]: every app  using the intel-media-driver tries to create /etc/igfx_user_feature{,_next}.txt


darix:
### Is it regression? Good in old configuration?

None

### What happened?

Found via running the apps in apparmor. it even tries to open the file in write mode. which makes me wonder ... should distributions install those 2 files? if not ... can the code that tries to write those files properly guarded that it is not built during normal building?

```
type=AVC msg=audit(1682983399.735:1932): apparmor="DENIED" operation="mknod" class="file" profile="vlc" name="/etc/igfx_user_feature.txt" pid=18661 comm="vlc" requested_mask="c" denied_mask="c" fsuid=1000 ouid=1000
type=AVC msg=audit(1682983399.735:1933): apparmor="DENIED" operation="mknod" class="file" profile="vlc" name="/etc/igfx_user_feature_next.txt" pid=18661 comm="vlc" requested_mask="c" denied_mask="c" fsuid=1000 ouid=1000
```

If those files should be user writeable ... /etc/ seems to be a weird path for them.

### What's the usage scenario when you are seeing the problem?

Others

### What impacted?

-

### Debug Information

intel-media-23.1.3

### Do you want to contribute a patch to fix the issue?

None

intel-mediadev:
Auto Created VSMGWL-64146 for further analysis.

VickyZengg:
Hi @darix， May I know what version of driver you are using? Release, debug or release internal?

VickyZengg:
Hi @darix ，I checked the issue in code, we will create "/etc/igfx_user_feature.txt" and "/etc/igfx_user_feature_next.txt" in release/debug/release internal version. So you find you use release version but it still open the two files. Because those two files is for regkey write and read. We have regkey used in release version not only in debug and release internal. So, the two files need create in release version. May I know if you still have concern about it?

VickyZengg:
Thank you for raising the issue, we will consider it and reply to you as soon as possible.

VickyZengg:
Hello, we already add code to guard igfx_user_feature.txt and igfx_user_feature_next.txt auto create in release version, if you want to use the files, you can create it manually. But it still under review and test. After merge will let you know. :) Thank you!

PR #1682, opened 2023-06-13
[Media Common] [VP] Guard Regkey file auto create in release vesion

VickyZengg:
Fixes: https://github.com/intel/media-driver/issues/1654

### DIFF
--- a/media_softlet/linux/common/os/osservice/mos_utilities_specific.cpp
+++ b/media_softlet/linux/common/os/osservice/mos_utilities_specific.cpp
@@ -1003,7 +1003,25 @@ MOS_STATUS MosUtilitiesSpecificNext::UserFeatureDumpDataToFile(const char *szFil
     MOS_PUF_KEYLIST   pKeyTmp;
     int32_t           j;
 
+   // For release version: writes to the file if it exists,
+// skips if it does not exist
+// For other version: always writes to the file
+
+#if(_RELEASE)
+    File = fopen(szFileName, "r");
+    if (File == NULL)
+    {
+        return MOS_STATUS_SUCCESS;
+    }
+    else
+    {
+        fclose(File);
+        File = fopen(szFileName, "w+");
+    }
+#else
     File = fopen(szFileName, "w+");
+#endif
+
     if ( !File )
     {
         return MOS_STATUS_USER_FEATURE_KEY_WRITE_FAILED;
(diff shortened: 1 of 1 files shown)


MicroYY (approved):

media_softlet/linux/common/os/osservice/mos_utilities_specific.cpp
```diff
@@ -1003,7 +1003,25 @@ MOS_STATUS MosUtilitiesSpecificNext::UserFeatureDumpDataToFile(const char *szFil
     MOS_PUF_KEYLIST   pKeyTmp;
     int32_t           j;
 
+   // For release version: writes to the file if it exists,
+// skips if it does not exist
+// For other version: always writes to the file
+
+#if(_RELEASE)
+    File = fopen(szFileName, "r");
```
darix, line 1011: isnt that a kinda weird way to say `access()` or `stat()`?

media_softlet/linux/common/os/osservice/mos_utilities_specific.cpp
```diff
@@ -1003,7 +1003,25 @@ MOS_STATUS MosUtilitiesSpecificNext::UserFeatureDumpDataToFile(const char *szFil
     MOS_PUF_KEYLIST   pKeyTmp;
     int32_t           j;
 
+   // For release version: writes to the file if it exists,
+// skips if it does not exist
+// For other version: always writes to the file
+
+#if(_RELEASE)
```
XinfengZhang, line 1010: I suppose we dont need this macro, should not create it even with debug version. could you help to check why debug version need create it?

media_softlet/linux/common/os/osservice/mos_utilities_specific.cpp
VickyZengg: Hi thank you! I will use access() replace this way. May I know if the code can solve this issue which you reporter?

media_softlet/linux/common/os/osservice/mos_utilities_specific.cpp
VickyZengg: Hi Xinfeng, debug/releaseinternal version need this regkey files to read/write regkey, it will be more convenient. Otherwise, we have to create it manually, it will be troublesome

media_softlet/linux/common/os/osservice/mos_utilities_specific.cpp
darix: all i want is not trying to open the file anymore for writing :)

so just checking with `access()` and `stat()` would be fine I guess.

VickyZengg: yeah, access() will be fine, I will change it.

PR #1702, opened 2023-07-25
Fix address of ‘VPHAL_RENDER_PARAMS::pSrc’ will never be NULL

cwilling:
Compile with gcc 13.1.0 produces error due to line 1107 check whether params->pSrc == nullptr. Removing the check enables compilation to complete.

Although gcc 13.1.0 could be blamed, if the error message is correct (it will never be NULL) then the check can safely be removed for all compilers.


cwilling:
I see this is now already fixed by commit 4bc121084644caaed6b2183d8a4cc7a2dc4a3885

PR #1705, opened 2023-07-27
fix pps override, this will cause a thread safety issue

leyu-yao:
Please do not merge it now. We may need more discussion.


1proprogrammerchant:
when can we merge it

PR #1708, opened 2023-08-05
[Media Common] Add DRM format mappings for JPEG decoder output

philipl:
The output formats from the JPEG decoder are currently not included
in the DRM format mapping functions. This means that any attempt to
call vaExportSurface on JPEG images will fail.

However, there is no functional gap beyond the missing mappings, so if
we add them in, exporting will then work correctly.

The four formats in question are IMC3, 422H, 422V, and 444P.

* 422H and 444P have trivial 1:1 mappings.
* IMC3 can be treated the same as I420, because the only difference is
  the chroma pitch, and that's explicitly set in the descriptor
* 422V (aka 440P) has the same memory layout as 422H - it's just a
  difference in how the consumer handle the chroma

I also added the missing XYUV composite object mapping - it was already
present for separate planes.

Tested with ffmpeg+mpv.

### DIFF
--- a/media_driver/linux/common/ddi/media_libva.cpp
+++ b/media_driver/linux/common/ddi/media_libva.cpp
@@ -6989,8 +6989,12 @@ static uint32_t DdiMedia_GetDrmFormatOfSeparatePlane(uint32_t fourcc, int plane)
         {
         case VA_FOURCC_NV12:
         case VA_FOURCC_I420:
+        case VA_FOURCC_IMC3:
         case VA_FOURCC_YV12:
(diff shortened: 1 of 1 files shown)


Jexu (approved):

Jexu:
Hi @philipl , thanks for your fix that helps it work in media lagecy code path. But intel xe plaforms will switch to the APO code path under media_softlet, so could you also help port this fix there in https://github.com/intel/media-driver/blob/master/media_softlet/linux/common/ddi/media_libva_interface_next.cpp?

philipl:
See #1718

PR #1732, opened 2023-10-16
[Media Common] [VP] Add 444p to linux vp xe_lpm_plus caps

vcheah:
This is to fix negotiation issue on MTL when constructing pipeline involving vajpegdec (444p) with vapostproc. 422H and 422V is already there on the media-driver vp caps.


vcheah:
> LGTM, but could you add signoff and add more words into the title? it is not clear

I've updated it accordingly.

vcheah:
Closing the MR since media VP does not support 422H/422V/444P.

vcheah:
Closing the MR since media VP does not support 422H/422V/444P.

PR #1733, opened 2023-10-18
[Encode] XRGB force to do swizzle for AVC/HEVC

vcheah:
This is to fix msdkh264enc/msdkh265enc incorrect encoded output when having BGRx input.


XinfengZhang (approved):

OttoYang (approved):
LGTM

thomasli21801 (approved):

PR #1735, opened 2023-11-03
[Media Common] Add non-free build info inside the vendor string

rgonzalezfluendo:
The behavior of the driver is very different if is built with the closed source pre-built binary shaders. Adding this info in the vendor string is very useful for debugging. Also, it take advantage of the empty parentheses that release build contains.


rgonzalezfluendo:
Many applications add libva driver vendor string information to their logs. Adding free/non-free information in vendor string is very useful when debugging.

PR #1739, opened 2023-11-20
Add FORCE_DISABLE_MTL_MEDIA_CCS env variable

SiewHoon:
Default media driver is enable Media CCS, would
like to add new FORCE_DISABLE_MTL_MEDIA_CCS in environment variable to disable Media CCS.
This is allow user choose to disable MTL Media CCS if needed.


SiewHoon:
New changed submit in https://github.com/intel/media-driver/pull/1760 
Closed this PR.

PR #1749, opened 2023-12-11
[Encode] Add QVBR rate control mode for AV1

BlakeB415:
Relies on https://github.com/oneapi-src/oneVPL-intel-gpu/pull/312

### DIFF
--- a/media_common/agnostic/common/codec/shared/codec_def_common_encode.h
+++ b/media_common/agnostic/common/codec/shared/codec_def_common_encode.h
@@ -548,6 +548,7 @@ struct CodechalTileInfo
     (rateControl == RATECONTROL_VBR)                || \
     (rateControl == RATECONTROL_AVBR)               || \
     (rateControl == RATECONTROL_CQL)                || \
+    (rateControl == RATECONTROL_QVBR)               || \
     ((( rateControl == RATECONTROL_VCM)       || \
       ( rateControl == RATECONTROL_ICQ)       || \
       ( rateControl == RATECONTROL_QVBR)      || \
--- a/media_driver/linux/Xe_M/ddi/media_libva_caps_dg2.cpp
+++ b/media_driver/linux/Xe_M/ddi/media_libva_caps_dg2.cpp
@@ -90,6 +90,7 @@ VAStatus MediaLibvaCapsDG2::LoadAv1EncProfileEntrypoints()
         AddEncConfig(VA_RC_VBR);
         AddEncConfig(VA_RC_ICQ);
         AddEncConfig(VA_RC_TCBRC);
+        AddEncConfig(VA_RC_QVBR);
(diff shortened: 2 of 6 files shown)


leyu-yao:
Hi @BlakeB415.
Encoding in BRC mode requires firmware to work together with engines. Our huc firmware does not support QVBR. As a result, this patch will not get the expected quality.
May I know in which scenario you would like to use QVBR? And is current VBR and ICQ not meeting your requirement?

BlakeB415:
> Hi @BlakeB415. Encoding in BRC mode requires firmware to work together with engines. Our huc firmware does not support QVBR. As a result, this patch will not get the expected quality. May I know in which scenario you would like to use QVBR? And is current VBR and ICQ not meeting your requirement?

The specific behavior we want is more like capped ICQ (similar to capped CRF) to where the ICQ BRC doesn’t exceed the set max bitrate. QVBR might not be the correct RC mode for this though considering this.

We want to ensure the bitrate stays within a certain range while also having it use the ICQ target quality.

leyu-yao:
> > Hi @BlakeB415. Encoding in BRC mode requires firmware to work together with engines. Our huc firmware does not support QVBR. As a result, this patch will not get the expected quality. May I know in which scenario you would like to use QVBR? And is current VBR and ICQ not meeting your requirement?
> 
> The specific behavior we want is more like capped ICQ (similar to capped CRF) where the ICQ BRC does not exceed the set max bitrate. QVBR might not be the correct RC mode for this use case though.
> 
> We want to ensure the bitrate stays within a certain range while also having it use the ICQ target quality scale.
> 
> This patch does have the result I needed, from my testing at least.

Did you try if you can get expected quality if using VBR mode? 
As our firmware does not support QVBR currently. So due to our code check-in policy, we cannot merge this PR to our mainline. If this PR can meet your requirement, I'd prefer to keep it as part of private build.

BlakeB415:
> > > Hi @BlakeB415. Encoding in BRC mode requires firmware to work together with engines. Our huc firmware does not support QVBR. As a result, this patch will not get the expected quality. May I know in which scenario you would like to use QVBR? And is current VBR and ICQ not meeting your requirement?
> > 
> > 
> > The specific behavior we want is more like capped ICQ (similar to capped CRF) where the ICQ BRC does not exceed the set max bitrate. QVBR might not be the correct RC mode for this use case though.
> > We want to ensure the bitrate stays within a certain range while also having it use the ICQ target quality scale.
> > This patch does have the result I needed, from my testing at least.
> 
> Did you try if you can get expected quality if using VBR mode? As our firmware does not support QVBR currently. So due to our code check-in policy, we cannot merge this PR to our mainline. If this PR can meet your requirement, I'd prefer to keep it as part of private build.

We tried VBR but the output files are larger than what we can achieve with ICQ given the perceived quality.

I'll close this and the linked issue. Thank you.

PR #1811, opened 2024-06-04
Expose va-attrib to configure scanout flag

vcheah:
This can be used to tell the driver set scanout surface. XE DMA DRM render case required this.
The MR has dependencies to libVA header: https://github.com/intel/libva/pull/820


vcheah:
Closing the MR.  Track this:  https://github.com/intel/media-driver/pull/1822

PR #1846, opened 2024-08-23
[Decode] Reenable VC1 decode for DG2

### LINKED ISSUE #1658: [Feature]: VC-1 decoding support for DG2
### What Feature?

Currently, VC-1 decoding support is not available on Linux on DG2 devices even though they support it and decoding of vc1 works with Windows. Intel reference document also shows VC-1 support for DG2 cards. => https://www.intel.com/content/www/us/en/docs/onevpl/developer-reference-media-intel-hardware/1-1/overview.html#DECODE-OVERVIEW-DISCRETE

### What's the usage scenario would be benifited?

Transcode for media delivery, Playback

### What impacted?

Having VC-1 decoding support is necessary for playback and also transcoding scenarios. 

Accelerating playback is nice, but more importantly for transcoding purposes. The sw decoder for vc-1 is limited to a single thread and therefore limiting transcodes quite a bit - having a GPU that could accelerate this is a must. As it already works on Windows I don´t really understand why it's not supported on Linux (yet).

Would be awesome if it could be added in a future release. Thanks.

### Do you want to contribute a patch to develop this feature?

None

nyanmisaka:
The [VC1 OLP kernel](https://github.com/intel/media-driver/blob/master/media_driver/agnostic/Xe_M/Xe_HPM/codec/kernel/XE_HPM_VC1_OLP.c) used by DG2 can share the same legacy CM interface with Gen12/TGLLP.
(BTW the existing kernel seems to target `GENX_TGLLP` instead of `GENX_DG2`)

Tested on Arc A380 with a few VC1 video samples.

Fixes: #1658

Signed-off-by: nyanmisaka <nst799610810@gmail.com>


nyanmisaka:
cc @Jexu

Zepmann:
Tested this as a patch for an Arch Linux package building recipe based on package [intel-media-driver](https://archlinux.org/packages/extra/x86_64/intel-media-driver/). It works flawlessly on Arc A310.

@XinfengZhang, @guowillyy
I am not trying to interfere in your planning, but can you please give some indication on whether:

1. This PR has a chance of being merged?
2. An indication of when you have time to thoroughly review this and maybe merge it (if approved, of course)?

I see no disadvantages. My 2 cents:

* The hardware supports this.
* The code is fully open source, so I would say there is no conflict of interest.
* Windows already supports VC-1 decoding on DG2. Why not Linux? :smiley:

nyanmisaka:
@Zepmann Unfortunately, Intel's own open-source collaboration methods differ significantly from similar projects like Mesa. If they have a reason not to support a deprecated feature and don't want to invest resources in testing, the code may never be merged, and you'll need to carry the patches yourself.

media_driver/media_interface/media_interfaces_dg2/media_interfaces_dg2.cpp
```diff
@@ -938,12 +938,18 @@ MOS_STATUS CMHalInterfacesXe_Hpm::Initialize(CM_HAL_STATE *pCmState)
         MHW_ASSERTMESSAGE("Create CM Hal interfaces failed.")
         return MOS_STATUS_NO_SPACE;
     }
+#ifdef _MEDIA_RESERVED
```
XinfengZhang, line 939: it means that these two kernel ISA will removed , why?

media_driver/media_interface/media_interfaces_dg2/media_interfaces_dg2.cpp
```diff
@@ -938,12 +938,18 @@ MOS_STATUS CMHalInterfacesXe_Hpm::Initialize(CM_HAL_STATE *pCmState)
...
     device->SetCopyKernelIsa((void*)pGPUCopy_kernel_isa_dg2, iGPUCopy_kernel_isa_size_dg2);
     device->SetInitKernelIsa((void*)pGPUInit_kernel_isa_dg2, iGPUInit_kernel_isa_size_dg2);
+#endif
 
     m_cmhalDevice = device;
     m_cmhalDevice->SetGenPlatformInfo(PLATFORM_INTEL_DG2, PLATFORM_INTEL_GT2, "DG2");
-    uint32_t cisaIDs[] = { GENX_DG2 , GENX_XEHP };
+#ifdef _MEDIA_RESVERED
```
XinfengZhang: looks you are using TGL isa on DG2?

XinfengZhang (commented):
I m not sure whether there is some risk that you replace current DG2 ISA with TGL one.

media_driver/media_interface/media_interfaces_dg2/media_interfaces_dg2.cpp
XinfengZhang: typo

nyanmisaka: Yes, this was done intentionally. The last platform CM supported was TGLx. DG2's VC1 kernel contained GENX_TGLLP instead of GENX_DG2, and not doing so would cause CM to fail when loading and verifying the kernel.

media_driver/media_interface/media_interfaces_dg2/media_interfaces_dg2.cpp
nyanmisaka: These two DG2 kernels have never been open sourced or upstreamed to this repo, so they are disabled by `_MEDIA_RESERVED`. Do you have any closed-source drivers using them internally?

media_driver/media_interface/media_interfaces_dg2/media_interfaces_dg2.cpp
nyanmisaka: Fixed

XinfengZhang:
@nyanmisaka , thanks for contribution.  about the PR, I still have 1 concern about changing  CMHal from  CmHalXe_Xpm;
to CM_HAL_G12_X, I actually not sure whether this cm HAL only used for OLP kernel or all kernels?

nyanmisaka:
> @nyanmisaka , thanks for contribution. about the PR, I still have 1 concern about changing CMHal from CmHalXe_Xpm; to CM_HAL_G12_X, I actually not sure whether this cm HAL only used for OLP kernel or all kernels?

Other functions on the DG2 do not depend on the CM device - they are either already embedded in the hardware via Low-Power mode or are film-grain kernels in AV1 decoding that do not depend on the CM. VC1 OLP kernel is the only function that requires CM initialization.

Regarding reliability, we have a large number of users using Alchemist in their transcoding servers. We've included this patch in our own iHD driver build since 2024, and we haven't received any reports of a regression.

We cannot guarantee that we have tested each type of VC1 sample, so if you later find a regression, feel free to revert it.

XinfengZhang (approved):
just like you mentioned , from my perspective, accepting new feature always means we should enlarge our test/release scope. but we will accept the feature enabling PR if there is real user, or someone's usage was blocked. maybe will not claim support on frontpage (is it ok for you? ) .
for this PR itself,  from my personal perspective,  VC1 should be deprecated, but just like your comments , someone is still using it. (actually, maybe I misunderstand something, looks the original request if for B580? it is battlemage), the PR is LGTM now.

Zepmann:
@XinfengZhang 

I think whether a codec should be deprecated or not solely based on its age/usage is not a good discussion. DG2 also supports decoding of MPEG-2, which is even older and also rarely used these days. Furthermore, CPU's have no problem decoding MPEG-2 in software. 

Instead, I would look at what the hardware can support. DG2 supports decoding of VC-1. It already does this in Windows. So why not support it in Linux? The hardware is there, the use cases are there (I also use this). Why not include it in the software?

I agree that it must be tested well and not cause regressions, of course. However, looking at @nyanmisaka and my experience with this patch, I would not expect trouble coming from this. And if in an obscure, untested scenario trouble occurs, the patch can always be reverted.

nyanmisaka:
> just like you mentioned , from my perspective, accepting new feature always means we should enlarge our test/release scope. but we will accept the feature enabling PR if there is real user, or someone's usage was blocked. maybe will not claim support on frontpage (is it ok for you? ) . for this PR itself, from my personal perspective, VC1 should be deprecated, but just like your comments , someone is still using it. (actually, maybe I misunderstand something, looks the original request if for B580? it is battlemage), the PR is LGTM now.

I understand that not stating it in the README is a form of disclaimer. This is perfectly acceptable as community level support.

VC1 has been removed from Battlemage and similar Xe2+ platforms, making it impossible to enable using this method. Therefore, this PR is entirely Alchemist oriented.

PR #1892, opened 2024-12-16
Remove debugging symbols by default

hmaarrfk:
Typically these are added by package integrator.

I don't think these should be enabled by default.


hmaarrfk:
They increase the "size" of the binary from 50MB -> 450MB.

PR #1897, opened 2025-01-09
[Encode] Fix ACM multi-channel AVC encode performance degradation

hoesheng:
disable DG2 Height alignment WA

### DIFF
--- a/media_driver/linux/Xe_M/ddi/media_sku_wa_xe.cpp
+++ b/media_driver/linux/Xe_M/ddi/media_sku_wa_xe.cpp
@@ -918,6 +918,8 @@ static bool InitDg2MediaWa(struct GfxDeviceInfo *devInfo,
 
     MEDIA_WR_WA(waTable, Wa_15013355402, 1);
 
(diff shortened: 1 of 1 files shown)


media_driver/linux/Xe_M/ddi/media_sku_wa_xe.cpp
```diff
@@ -443,7 +443,6 @@ static bool InitTglMediaWaExt(struct GfxDeviceInfo *devInfo,
     }
 
     MEDIA_WR_WA(waTable, WaEnableVPPCopy, 1);
-    MEDIA_WR_WA(waTable, Wa_AvcUnalignedHeight, 1);
```
bai-isaac, line 446: InitTglMediaWaExt is a common function for gen12, include TGL, DG2 ...  please only remove for DG2, thanks.

bai-isaac (changes requested):

media_driver/linux/Xe_M/ddi/media_sku_wa_xe.cpp
hoesheng: updated

bai-isaac (approved):

PR #1936, opened 2025-06-04
add loongarch64 build support

amazingfate:
Initial work is done at https://github.com/FanFansfan/media-driver/tree/loongarch by @FanFansfan.

I use libsimde-dev on system for SIMDE headers.

This pr depends on https://github.com/intel/gmmlib/pull/133 for build dep.

Tested with Intel Arc B580 on Loongson 3A6000 desktop platform. Vaapi decoding/encoding is working fine.

### DIFF
--- a/media_driver/agnostic/common/cm/cm_mem.h
+++ b/media_driver/agnostic/common/cm/cm_mem.h
@@ -25,9 +25,15 @@
 //!
 #pragma once
 
+#if defined(__loongarch64)
+#define SIMDE_X86_SSE2_ENABLE_NATIVE_ALIASES
+#include <simde/x86/sse2.h>
+#else
 #include <mmintrin.h>
 #include <xmmintrin.h>
 #include <emmintrin.h>
+#endif
+
 #include "cm_debug.h"
 #include "mos_utilities.h"
 
@@ -43,7 +49,11 @@ enum CPU_INSTRUCTION_LEVEL
     NUM_CPU_INSTRUCTION_LEVELS
 };
 
+#if defined(__loongarch64)
+typedef __m128i             DQWORD;         // 128-bits,   16-bytes
+#else
 typedef __m128              DQWORD;         // 128-bits,   16-bytes
+#endif
 typedef uint32_t            PREFETCH[8];    //             32-bytes
 typedef uint32_t            CACHELINE[8];   //             32-bytes
 typedef uint16_t            DHWORD[32];     // 512-bits,   64-bytes
@@ -230,9 +240,12 @@ inline CPU_INSTRUCTION_LEVEL GetCpuInstructionLevel( void )
     int cpuInfo[4];
     memset( cpuInfo, 0, 4*sizeof(int) );
 
+#if !defined(__loongarch64)
     GetCPUID(cpuInfo, 1);
+#endif
 
     CPU_INSTRUCTION_LEVEL cpuInstructionLevel = CPU_INSTRUCTION_LEVEL_UNKNOWN;
+#if !defined(__loongarch64)
     if( (cpuInfo[2] & BIT(19)) && TestSSE4_1() )
     {
         cpuInstructionLevel = CPU_INSTRUCTION_LEVEL_SSE4_1;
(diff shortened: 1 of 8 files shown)


media_driver/agnostic/common/cm/cm_mem.h
```diff
@@ -25,9 +25,15 @@
 //!
 #pragma once
 
+#if defined(__loongarch64)
+#define SIMDE_X86_SSE2_ENABLE_NATIVE_ALIASESAdd commentMore actions
```
yxd-ym, line 29: ?

media_driver/linux/common/cm/hal/osservice/cm_mem_os.h
```diff
@@ -26,8 +26,13 @@
 #pragma once
 
 #include <iostream>
+#if defined(__loongarch64)
+#define SIMDE_X86_SSE2_ENABLE_NATIVE_ALIASESAdd commentMore actions
```
yxd-ym, line 30: same here

media_driver/agnostic/common/cm/cm_mem.h
amazingfate: Sorry for this mistake, fixed.

media_driver/cmake/linux/media_compile_flags_linux.cmake
```diff
@@ -18,6 +18,54 @@
...
+
+    # Global defines
+    -DLINUX=1
+    -DLINUX
+    -DNO_RTTI
+    -DNO_EXCEPTION_HANDLING
+    -DINTEL_NOT_PUBLIC
+    -g
+    -Wno-overloaded-virtual
+)
```
nyanmisaka, line 67:
```
# Disable x86-only legacy iGPUs
# DG1, DG2 and BMG (possibly) dGPUs rely on GEN12
-DGEN8=OFF
-DGEN9=OFF
-DGEN11=OFF
```

This will [trim away](https://github.com/intel/media-driver/blob/master/media_driver/cmake/linux/media_gen_flags_linux.cmake) code that will never be reached on non-x86 platforms to save space. Theoretically it is also possible to disable the GEN12+ iGPUs. However, this would increase the maintenance burden.

The same if true for `media_softlet/cmake/linux/media_compile_flags_linux.cmake`.

amazingfate: I'm thinking of reusing the common flags for all architectures and then add architecture specific flags in conditions, so we don't have the long same code.

If GEN8, GEN9 and GEN11 are x86 only, it should be fine to disable them on other platforms.

nyanmisaka: Yes. Apply common build options first, then conditionally add platform-specific options.

Intel has never released iGPU for non-x86 platforms. Now iHD*.so has swelled to 40 MB+ on x86, disabling irrelevant platforms (especially media kernels) on non-x86 should shrink it a lot.

XinfengZhang: it is loongarch CPU and Intel Arc , so suppose you could disable all integrate card support. 
about the patch itself:
no need to check whether it is loongarch, because if host is arm or other arch it also need some changes. , just check whether SSE instincts exists?   there are another similar PR
#1446 to enable ARC on non-x86 platforms 

from another perspective , I am wondering whether ARC still use cmfastcopy function .

amazingfate: @nyanmisaka I find that add flags like `-DGEN8=OFF` won't work, because MEDIA_COMPILER_FLAGS_COMMON defines flags used by compiler, not cmake. We should add it when running cmake command.

nyanmisaka: > @nyanmisaka I find that add flags like `-DGEN8=OFF` won't work, because MEDIA_COMPILER_FLAGS_COMMON defines flags used by compiler, not cmake. We should add it when running cmake command.

You seem to be right. But it is unlikely that downstream package maintainers will be aware of these undocumented flags and make small builds.

If possible, it might be better to let `CMAKE_SYSTEM_PROCESSOR` take effect in `media_gen_flags_linux.cmake`.

PR #1990, opened 2026-02-28
[MOS] Fix SIGBUS crash on xe small BAR systems during vaInitialize

KevinKickass:
## Summary

- Fix SIGBUS crash in `iHD_drv_video.so` during `vaInitialize` on xe small BAR systems
- Detect small BAR by comparing `cpu_visible_size` against `total_size` for VRAM regions
- Set `DRM_XE_GEM_CREATE_FLAG_NEEDS_VISIBLE_VRAM` on VRAM allocations when small BAR is detected, with system memory as fallback placement

## Problem

On discrete GPUs (DG2 / Arc A750/A770) running with the xe kernel driver on platforms without Resizable BAR support, only a fraction of VRAM is CPU accessible (e.g. 256MB of 8GB). GEM objects allocated in VRAM via `mos_bo_alloc_xe` may be placed in non-CPU-accessible VRAM regions. When the driver later mmaps and accesses these buffers, the CPU faults with SIGBUS.

Crash stack:
```
#0  libc.so (SIGBUS)
#1  iHD_drv_video.so  (mos_bufmgr_xe.c - buffer mapping)
...
#13 libva.so           (vaInitialize)
```

This makes VA-API completely unusable on xe small BAR systems, affecting all applications using hardware video acceleration (Chromium/Electron apps crash on startup).

## Fix

The xe KMD already provides `DRM_XE_GEM_CREATE_FLAG_NEEDS_VISIBLE_VRAM` for exactly this scenario. The flag ensures VRAM allocations land in the CPU visible portion, and is a noop on full BAR systems. System memory is added as a fallback placement so the kernel can spill when CPU visible VRAM is full, matching the i915 `I915_GEM_CREATE_EXT_FLAG_NEEDS_CPU_ACCESS` behavior.

## Test plan

- Tested on Intel Arc A750 (DG2, device ID 56a1) with xe driver on X299 platform without Resizable BAR (256MB BAR / 8GB VRAM)
- Verified `vaInitialize` no longer crashes with SIGBUS
- No impact on full BAR systems (flag is a noop per xe KMD documentation)

### DIFF
--- a/media_softlet/linux/common/os/xe/mos_bufmgr_xe.c
+++ b/media_softlet/linux/common/os/xe/mos_bufmgr_xe.c
@@ -233,6 +233,7 @@ typedef struct mos_xe_bufmgr_gem {
 
     //Note: DON't put these fields in xe_device
     bool     has_vram;
+    bool     has_small_bar;
     uint8_t  va_bits;
     /** bitmask of all memory regions */
     uint64_t mem_regions_mask;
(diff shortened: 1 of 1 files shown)


simplexforce:
I tested it on my B580, and it works perfectly.

philipborg:
I also tested it with my B580 on a system without resizable BAR support and now VAINFO works. It would be amazing to get this merged so I don't need to compile custom builds just to use my GPU.

XinfengZhang (approved):

jmuecke:
@Jexu Can we get this merged? This is really annoying from a user's persepective.

agowa:
Agree, @jmuecke in the mean time you can try to see if https://github.com/xCuri0/ReBarUEFI works on your Mainboard. You may need an in-circuit chip programming clip though.

PR #2003, opened 2026-05-04
Enable FtrSWMediaReset for EHL

SiewHoon:


XinfengZhang (dismissed):
media reset is not a official feature on EHL, so, I could not guarantee it works on EHL,  have you test it , and ensure it works withouth regression？

XinfengZhang (approved):

SiewHoon:
> media reset is not a official feature on EHL, so, I could not guarantee it works on EHL, have you test it , and ensure it works withouth regression？

No regression issues have been found so far that are caused by enabling FtrSWMediaReset on the EHL platform with sanity test run using gst-va and gst-msdk side.

Issue #1989, opened 2026-02-26
[Feature]: Hardware Encoding (32 bit) driver for Steam Play remote streaming


michaelr0:
### What Feature?

Hello there,

It appears that there isn't a 32 bit version of this driver, which in 2026 I'd normally understand, however from the perspective of using Steam/Proton, it seems rather important that 32 bit support is maintained, because without it, the Arc B580 is unable to use hardware encoding for Steam's Remote Play Streaming feature and falls back to CPU/Software video encoding which is pretty lacklustre in performance.

Would you consider adding a 32 bit version or an alternative solution that will allow this functionality to work via the Intel Driver/Hardware.

Thanks

### What's the usage scenario would be benifited?

Cloud Gaming

### Do you want to contribute a patch to develop this feature?

None

PR #1493, opened 2022-09-19
Fix the 32 bit build regression

cqian2:
Fixes: #1468

PR #2005, opened 2026-05-07
[Mos] Fix Xe media engine reset on 32-bit media-driver

matte-schwartz:
batch_addrs[0] (uint64_t) was cast through uintptr_t before assignment to exec.address (__u64). On 32-bit builds this truncates GPU VAs above 4GB, causing the kernel to reject or fault the dispatch and reset the media engine.

Pass batch_addrs[0] directly. For the array pointer branch, use the standard (uint64_t)(uintptr_t) idiom.

### DIFF
--- a/media_softlet/linux/common/os/xe/mos_bufmgr_xe.c
+++ b/media_softlet/linux/common/os/xe/mos_bufmgr_xe.c
@@ -2561,7 +2561,7 @@ mos_bo_context_exec_with_sync_xe(struct mos_linux_bo **bo, int num_bo, struct mo
      * exec.address only accepts batch->offset64 when num bo == 1;
      * and it only accepts batch array when num bo > 1
     */
-    exec.address = (num_bo == 1 ? (uintptr_t)batch_addrs[0] : (uintptr_t)batch_addrs);
+    exec.address = (num_bo == 1 ? batch_addrs[0] : (uint64_t)(uintptr_t)batch_addrs);
     exec.num_batch_buffer = num_bo;
     ret = drmIoctl(bufmgr_gem->fd, DRM_IOCTL_XE_EXEC, &exec);
     if (ret)


matte-schwartz:
Sorry, I was under the impression given some older merged PRs in this repo (https://github.com/intel/media-driver/pull/1493) that 32-bit was somewhat supported, but I see now I was mistaken and that you've stated 32-bit builds are not supported: https://github.com/intel/media-driver/issues/1989#issuecomment-4108818599.

This was actually for a media engine reset I ran into while looking into the feature requested in that issue, so it probably makes more sense to continue discussions there.

fviolence:
I tested this patch on an Intel Arc Pro B70 system using the `xe` kernel driver, and it fixes the 32-bit H.264 VAAPI encode failure I was seeing.

Test environment (32-bit VAAPI stack):
```text
OS: Arch Linux x86_64
GPU: Intel Arc Pro B70
Kernel driver: xe
lib32-libva 2.23.0
lib32-intel-gmmlib 22.10.0
lib32-intel-media-driver 26.1.5
```

Before applying this PR, 32-bit VAAPI initialized successfully, but actual encode failed. A minimal 32-bit FFmpeg H.264 VAAPI test failed with:
```text
Failed to map output buffers: 24 (internal encoding error)
Output failed: -5
Error encoding a frame: Input/output error
```

Steam Remote Play showed the same effective failure through its 32-bit VAAPI path:
```text
CGameStreamVideoStageVAAPI: Error encoding: Input/output error
```

After applying this PR, the same 32-bit FFmpeg H.264 VAAPI test succeeds. Steam Remote Play H.264 hardware encoding also works after applying the patch. The stream now uses HW encode, video is visible, and I no longer see the previous slow-encode / black-screen behavior in a short test.

So from my testing, this PR fixes a real 32-bit VAAPI H.264 encode failure on Battlemage / `xe`, and it also fixes Steam Remote Play H.264 hardware encoding on this system.

Any plans on merging the PR?

matte-schwartz:
@fviolence fwiw, in case it doesn't get merged, using the SteamRT3 beta seems to work with just the 64-bit `intel-media-driver` needed.

fviolence:
Yes, SteamRT3 indeed uses the 64-bit `intel-media-driver`. However, the feature itself is experimental, and the stream is barely usable because of occasional, but frequent image freezes.

